PR #11, opened 2026-02-11
[feat] Proto API for Gateway service

sbalabanov:
API object model for the controller implementing getting a SubmitQueue land request


gateway/proto/gateway.proto
```diff
@@ -25,8 +29,72 @@ message PingResponse {
...
+    // Same as REBASE but squash commits into a single commit before rebase
+    SQUASH_REBASE = 2;
+    // Merge commits into the target branch by creating a separate merge commit, preserving the commit history along with hashes
+    MERGE = 3;
+}
+
+// Change represents a set of related code changes identified by one or more IDs from a particular code change supplier, like Github Pull Requests.
+message Change {
+    // The code change supplier (e.g., "github", "gerrit", "phabricator").
+    string source = 1;
```
behinddwalls, line 47: should say `provider` maybe?

gateway/proto/gateway.proto
```diff
@@ -25,8 +29,72 @@ message PingResponse {
...
+    // List of change IDs, in a format specific to the code change supplier, that should be landed together. SubmitQueue guarantees that the changes are landed in the order of the list,
+    // and no other changes are landed in between. SubmitQueue does not guarantee that each change is individually valid, but produces a special validity
+    // marker on such changes. The user is responsible to include all changes in a change stack in the order of the list, starting from the earlist change.
+    repeated string ids = 2;
```
behinddwalls, line 51: maybe explicitly `source_ids` or `provider_ids`

gateway/proto/gateway.proto
```diff
@@ -25,8 +29,72 @@ message PingResponse {
...
+    // Change (such as a pull request) to land into the target branch. Target branch is defined by the queue configuration.
+    Change change = 2;
+    // Source control integration strategy to use for this land operation. If not specified, the default queue strategy is used.
+    Strategy strategy = 4;
+}
+
+// LandResponse defines the response to a land request.
+message LandResponse {
+    // Globally unique identifier for the land request. Used to track the land request lifecycle.
+    string sqid = 1;
```
behinddwalls, line 67: maybe we just call it `submission_id` or `request_id`

gateway/proto/gateway.proto
```diff
@@ -25,8 +29,72 @@ message PingResponse {
...
+    // The code change supplier (e.g., "github", "gerrit", "phabricator").
+    string source = 1;
+    // List of change IDs, in a format specific to the code change supplier, that should be landed together. SubmitQueue guarantees that the changes are landed in the order of the list,
+    // and no other changes are landed in between. SubmitQueue does not guarantee that each change is individually valid, but produces a special validity
+    // marker on such changes. The user is responsible to include all changes in a change stack in the order of the list, starting from the earlist change.
+    repeated string ids = 2;
+}
+
+// LandRequest defines a request to land (merge into target git branch) a set of code changes represented by a list of PRs.
+message LandRequest {
```
behinddwalls, line 55: Do we want to call it Land vs Merge? or just `Enqueue` or `Submit`

gateway/proto/gateway.proto
sbalabanov: sure

gateway/proto/gateway.proto
sbalabanov: I prefer short names and long comments :)

gateway/proto/gateway.proto
sbalabanov: I was debating between Land and Merge and chose the former because Merge is often associated with a very distinct git behavior... and SubmitQueue is designed around rebase-driven development first.
Thoughts?

gateway/proto/gateway.proto
sbalabanov: was thinking request_id but picked sqid (submit queue id) as this is the only identifier the customer operates with. Decided to make it first class. Not super attached to it.

sbalabanov:
only changed comments supplier-<provider so far, let me know if you have strong preference on naming

gateway/controller/land.go
```diff
@@ -0,0 +1,54 @@
...
+	"fmt"
+	"time"
+
+	"github.com/uber-go/tally/v4"
+	pb "github.com/uber/submitqueue/gateway/protopb"
+	"go.uber.org/zap"
+)
+
+// LandController handles land business logic for the gateway
+type LandController struct {
```
manjari25, line 14: Can this be called `Controller` simply?

gateway/controller/land.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+// NewLandController creates a new instance of the gateway land controller
+func NewLandController(logger *zap.Logger, scope tally.Scope) *LandController {
+	if logger == nil {
+		logger = zap.NewNop()
+	}
+	if scope == nil {
+		scope = tally.NoopScope
+	}
```
manjari25, line 26: Can we default initialize? For tests we can initialize with `noop` or as needed (if validating things in the logger or scope).

gateway/controller/land_test.go
```diff
@@ -0,0 +1,34 @@
...
+	}
+	resp, err := controller.Land(ctx, req)
+
+	if err != nil {
+		t.Fatalf("Land() returned error: %v", err)
+	}
+
+	if resp.Sqid == "" {
+		t.Fatal("Expected sqid to be set, got empty string")
+	}
```
manjari25, line 33: Any reason we are not performing assertions here instead? Why do we need `t.Fatal`?

gateway/proto/gateway.proto
```diff
@@ -25,8 +29,72 @@ message PingResponse {
...
+// Strategy defines the possible source control integration methods
+enum Strategy {
+    // Default strategy (let server decide based on configuration)
+    STRATEGY_DEFAULT = 0;
+    // Rebase commits onto target branch before landing
+    REBASE = 1;
+    // Same as REBASE but squash commits into a single commit before rebase
+    SQUASH_REBASE = 2;
+    // Merge commits into the target branch by creating a separate merge commit, preserving the commit history along with hashes
+    MERGE = 3;
```
manjari25: We should prefix all these with `STRATEGY_` to avoid any potential collisions.

gateway/proto/gateway.proto
manjari25: Can it be an enum?

gateway/proto/gateway.proto
```diff
@@ -25,8 +29,72 @@ message PingResponse {
...
+message Change {
+    // The code change provider (e.g., "github", "gerrit", "phabricator").
+    string source = 1;
+    // List of change IDs, in a format specific to the code change provider, that should be landed together. SubmitQueue guarantees that the changes are landed in the order of the list,
+    // and no other changes are landed in between. SubmitQueue does not guarantee that each change is individually valid, but produces a special validity
+    // marker on such changes. The user is responsible to include all changes in a change stack in the order of the list, starting from the earlist change.
+    repeated string ids = 2;
+}
+
+// LandRequest defines a request to land (merge into target git branch) a set of code changes represented by a list of PRs.
```
manjari25:
```suggestion
// LandRequest defines a request to land (merge into target git branch) a set of code changes represented by a list of changeIDs.
```

gateway/proto/gateway.proto
```diff
@@ -25,8 +29,72 @@ message PingResponse {
...
+message LandRequest {
+    // Name of the queue processing the land request. The queue should be defined in the configuration, otherwise an error is returned.
+    string queue = 1;
+    // Change (such as a pull request) to land into the target branch. Target branch is defined by the queue configuration.
```
manjari25, line 58: Can we clarify if this is a url or a change ID with an e.g?

gateway/controller/land.go
behinddwalls: we could but then we might need create another folder say `controller/land/controller.go`

gateway/proto/gateway.proto
behinddwalls: i think it things easier to just put anything depending on use-case updating models

gateway/proto/gateway.proto
behinddwalls: Land to fine...

gateway/proto/gateway.proto
behinddwalls: Yes, i think we can just call request_id.

gateway/controller/land.go
sbalabanov: I leave it as is for now, following the "Ping" example

gateway/controller/land.go
sbalabanov: we probably should, again just following the example for now

gateway/controller/land_test.go
sbalabanov: following the example, we should switch to asserts at some point, will require adding a dependency

gateway/proto/gateway.proto
sbalabanov: enum will mean every new provider need to be announced here - but we want people to be able to hook in an arbitrary implementation without interacting with the publicly facing code

behinddwalls (approved):

manjari25 (approved):

PR #15, opened 2026-02-13
feat(speculator): add API spec for speculate and signal requests

behinddwalls:
This change add the API spec for Speculator Service. We have 2 scenarios where speculation/respeculation needs to happen,

1. A new request is for speculation
2. Build or Merge/Finalize are done

in both cases, we want to speculate, reassert on the speculated paths based on signal and determine whether we can finalize or not.


behinddwalls:
This change is part of the following stack:

- #15 ◀

<sub>Change managed by [git-spice](https://abhinav.github.io/git-spice/).</sub>

PR #17, opened 2026-02-15
[feat] Draft showcase implementation of Request entity

sbalabanov:
Request entity and corresponding Request store interface.
Slightly redesigned folder structure.

### DIFF
--- a/entities/request.go
+++ b/entities/request.go
@@ -1,0 +1,97 @@
+package entities
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+)
+
+// RequestLandStrategy defines the possible source control integration methods.
+type RequestLandStrategy int
+
+const (
+	// RequestLandStrategyDefault lets the server decide based on configuration.
+	RequestLandStrategyDefault RequestLandStrategy = 0
+	// RequestLandStrategyRebase rebases commits onto the target branch before landing.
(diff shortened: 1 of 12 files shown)


entities/request.go
```diff
@@ -0,0 +1,97 @@
...
+type Change struct {
+	// Source is the code change provider (e.g., "github", "gerrit", "phabricator").
+	Source string
+	// IDs is a list of change IDs, in a format specific to the code change provider, that should be landed together.
+	IDs []string
+}
+
+// Request defines a request to land (merge into target branch of the source control repository) a set of code changes.
+// The object is immutable after creation.
+type Request struct {
```
behinddwalls, line 50: Thought?

I have contemplating whether to enforce immutability when building entities? enforce Builder pattern across the board do achieve that...so even in code, things are immutable....and are not changed

behinddwalls: Maybe it's too much and complicated to add everytime new things...

sbalabanov: My stance on this is here: unless there is a language support to enforce immutability (Go does not), do it, otherwise keep it simple and rely on proper documentation / code review.
I do not think the runtime tax and "line-of-code" tax is worth it.

behinddwalls (approved):

PR #18, opened 2026-02-16
feat(queue): add vendor-agnostic messaging queue interfaces

behinddwalls:
## Summary
###  Why?
                                                                                                                                                                                                                                                                                          
  To enable vendor-agnostic queue integration across multiple backends (Kafka, SQS, RabbitMQ) without coupling to specific implementations.
                                                                                                                                                                                                                                                                                          
###  What?                                                                    

  Adds queue extension interfaces and message entity:

  extensions/queue/
  - Delivery interface - provides Message(), Ack(), Nack(), ExtendVisibilityTimeout()
  - Subscriber interface - returns (<-chan Delivery, error)
  - Publisher interface - publishes messages to topics
  - Factory interface - creates publishers/subscribers

  entities/queue/
  - Message value type with ID, Payload, Metadata, PartitionKey, PublishedAt
  - Includes Copy() for deep copying

## Test Plan
- Message entity tests pass
  - Extension interfaces build successfully
  - Maps.Clone() handles nil metadata

## Stack
1. @ #18
1. #25
1. #20
1. #21
1. #22
1. #23
1. #24

### DIFF
--- a/entities/queue/message.go
+++ b/entities/queue/message.go
@@ -1,0 +1,55 @@
+package queue
+
+import (
+	"maps"
+	"time"
+)
+
+// Message represents a queue message entity.
+// Immutable - use Copy() for modifications.
+type Message struct {
+	// ID uniquely identifies the message for deduplication and tracing.
+	ID string
+
+	// Payload is the message body as raw bytes.
+	Payload []byte
+
+	// Metadata contains key-value pairs for headers and attributes.
+	// Use for trace IDs, request IDs, and cross-service metadata.
+	Metadata map[string]string
+
+	// PartitionKey determines which partition/shard this message goes to.
+	// Messages with the same PartitionKey are guaranteed ordered delivery.
+	// Optional - if empty, backend may use round-robin distribution.
+	PartitionKey string
+
+	// PublishedAt is when the message was published (Unix milliseconds).
+	PublishedAt int64
+}
+
+// NewMessage creates a new message with the given ID and payload.
+// Metadata is initialized as an empty map.
+// PublishedAt is set to the current time.
+func NewMessage(id string, payload []byte) Message {
+	return Message{
+		ID:          id,
+		Payload:     payload,
+		Metadata:    make(map[string]string),
+		PublishedAt: time.Now().UnixMilli(),
+	}
+}
+
+// Copy creates a deep copy of the message.
+// Safe to call concurrently.
+func (m Message) Copy() Message {
+	payloadCopy := make([]byte, len(m.Payload))
+	copy(payloadCopy, m.Payload)
+
+	return Message{
+		ID:           m.ID,
+		Payload:      payloadCopy,
+		Metadata:     maps.Clone(m.Metadata),
+		PartitionKey: m.PartitionKey,
+		PublishedAt:  m.PublishedAt,
+	}
+}
--- a/extensions/queue/publisher.go
+++ b/extensions/queue/publisher.go
@@ -1,0 +1,17 @@
+package queue
+
+import (
+	"context"
+
+	"github.com/uber/submitqueue/entities/queue"
+)
+
+// Publisher publishes messages to topics.
+// Implementations must be thread-safe.
+type Publisher interface {
+	// Publish sends a message to the specified topic.
+	Publish(ctx context.Context, topic string, message queue.Message) error
+
+	// Close gracefully shuts down the publisher, flushing pending messages.
+	Close() error
+}
--- a/extensions/queue/queue.go
+++ b/extensions/queue/queue.go
@@ -1,0 +1,17 @@
+package queue
+
+// Queue creates and manages queue publishers and subscribers.
+// Implementations handle connection pooling, consumer group configuration,
+// and resource lifecycle.
+type Queue interface {
+	// Publisher returns a Publisher instance.
+	// May return a singleton or new instance depending on implementation.
+	Publisher() Publisher
+
+	// Subscriber returns a Subscriber instance.
+	// May return a singleton or new instance depending on implementation.
+	Subscriber() Subscriber
+
+	// Close shuts down the queue and all associated resources.
+	Close() error
+}
--- a/extensions/queue/subscriber.go
+++ b/extensions/queue/subscriber.go
@@ -1,0 +1,24 @@
+package queue
+
+import (
+	"context"
+)
+
+// Subscriber consumes messages from topics.
+// Implementations must be thread-safe.
+type Subscriber interface {
+	// Subscribe starts consuming messages from the specified topic.
+	// Returns a channel of Delivery instances and an error if subscription fails.
+	//
+	// The channel is closed when the subscriber is closed or context is cancelled.
+	// Implementations should handle infrastructure errors internally (e.g., reconnect).
+	//
+	// Each Delivery provides the message and methods to acknowledge or reject it.
+	// Consumers should call delivery.Ack() or delivery.Nack() for each delivery.
+	Subscribe(ctx context.Context, topic string) (<-chan Delivery, error)
+
+	// Close gracefully shuts down the subscriber.
+	// All delivery channels will be closed.
+	// Idempotent - safe to call multiple times.
+	Close() error
+}
--- a/entities/queue/BUILD.bazel
+++ b/entities/queue/BUILD.bazel
@@ -1,0 +1,15 @@
+load("@rules_go//go:def.bzl", "go_library", "go_test")
+
+go_library(
+    name = "queue",
+    srcs = ["message.go"],
+    importpath = "github.com/uber/submitqueue/entities/queue",
+    visibility = ["//visibility:public"],
+)
+
+go_test(
+    name = "queue_test",
+    srcs = ["message_test.go"],
+    embed = [":queue"],
+    deps = ["@com_github_stretchr_testify//assert"],
+)
--- a/entities/queue/message_test.go
+++ b/entities/queue/message_test.go
@@ -1,0 +1,80 @@
+package queue
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestNewMessage(t *testing.T) {
+	id := "test-id"
+	payload := []byte("test payload")
+
+	msg := NewMessage(id, payload)
+
+	assert.Equal(t, id, msg.ID)
+	assert.Equal(t, payload, msg.Payload)
+	assert.NotNil(t, msg.Metadata)
+	assert.Empty(t, msg.Metadata)
+	assert.NotZero(t, msg.PublishedAt)
+}
+
+func TestMessage_Copy(t *testing.T) {
+	original := NewMessage("id-123", []byte("payload"))
+	original.Metadata["key"] = "value"
+	original.PartitionKey = "partition-1"
+
+	copied := original.Copy()
+
+	// Verify immutable fields are equal
+	assert.Equal(t, original.ID, copied.ID)
+	assert.Equal(t, original.PublishedAt, copied.PublishedAt)
+	assert.Equal(t, original.PartitionKey, copied.PartitionKey)
+
+	// Verify deep copy of payload
+	assert.Equal(t, original.Payload, copied.Payload)
+	original.Payload[0] = 'X'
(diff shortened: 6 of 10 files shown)


entities/queue/delivery.go
```diff
@@ -0,0 +1,94 @@
...
+import (
+	"context"
+	"sync"
+	"time"
+)
+
+// Delivery wraps a Message with delivery semantics and backend-specific metadata.
+type Delivery struct {
+	// msg is the underlying message (unexported to avoid field/method collision)
+	msg *Message
```
sbalabanov: I suggest to architect entities around value types for as much as possible

entities/queue/delivery.go
```diff
@@ -0,0 +1,94 @@
...
+
+	// Internal delivery control functions
+	ackFunc                     func(ctx context.Context) error
+	nackFunc                    func(ctx context.Context, requeueAfter time.Duration) error
+	extendVisibilityTimeoutFunc func(ctx context.Context, d time.Duration) error
```
sbalabanov: why need to be members?
more idiomatic approach is free standing functions or a separate interface, while entities are kept DTO-style

entities/queue/delivery.go
```diff
@@ -0,0 +1,94 @@
...
+
+// Acked returns a channel that is closed when the message is acknowledged.
+// Useful for testing and observability.
+func (d *Delivery) Acked() <-chan struct{} {
+	return d.ackedChan
+}
+
+// Nacked returns a channel that is closed when the message is rejected.
+// Useful for testing and observability.
+func (d *Delivery) Nacked() <-chan struct{} {
```
sbalabanov: may be single channel returning enum of events is cleaner?

entities/queue/delivery_test.go
```diff
@@ -0,0 +1,302 @@
...
+				return nil
+			},
+			nil,
+			nil,
+		)
+
+		// Verify acked channel is open
+		select {
+		case <-delivery.Acked():
+			t.Fatal("acked channel should be open")
```
sbalabanov: prefer asserts: `require.Fail()`

sbalabanov: to reliably verify the channel is open, use `_, ok := <- ch` then check for `ok`

entities/queue/delivery_test.go
```diff
@@ -0,0 +1,302 @@
...
+
+		err := delivery.Ack(ctx)
+		assert.NoError(t, err)
+		assert.True(t, ackCalled)
+
+		// Verify acked channel is closed
+		select {
+		case <-delivery.Acked():
+			// Expected
+		case <-time.After(10 * time.Millisecond):
```
sbalabanov: should not rely on hardcoded timeouts in tests

entities/queue/message.go
```diff
@@ -0,0 +1,72 @@
...
+// The returned message has copied Payload and Metadata, and shares the same PublishedAt.
+// Safe to call concurrently.
+func (m *Message) Copy() *Message {
+	// Deep copy payload
+	payloadCopy := make([]byte, len(m.Payload))
+	copy(payloadCopy, m.Payload)
+
+	// Deep copy metadata
+	metadataCopy := make(map[string]string, len(m.Metadata))
+	for k, v := range m.Metadata {
```
sbalabanov: can use `maps.Clone()`

extensions/queue/subscriber.go
```diff
@@ -0,0 +1,21 @@
...
+
+// Subscriber consumes messages from topics/queues.
+type Subscriber interface {
+	// Subscribe starts consuming messages from the specified topic.
+	// Returns a read-only channel that delivers Delivery instances.
+	// The channel is closed when the subscriber is closed or context is cancelled.
+	//
+	// Deliveries have Ack() and Nack() methods. Consumers should call
+	// delivery.Ack(ctx) or delivery.Nack(ctx, requeueAfter) for each delivery.
+	Subscribe(ctx context.Context, topic string) (<-chan *queue.Delivery, error)
```
sbalabanov: related to earlier comment, this seems to violate Go-idiomatic patterns of separating data and bahavior.
Can Subscribe return a channel over Message, and a standalone interface to be used to
`ack (ctx context.Context, messageID string)` or `ack(ctx context.Context, message Message)` ?

Makefile
```diff
@@ -43,7 +43,7 @@ build:
 # Run unit tests using Bazel (excludes integration tests which require running servers)
 test:
 	@echo "Running unit tests..."
-	@$(BAZEL) test //... --test_tag_filters=-manual,-integration || echo "No unit tests found (only integration tests exist)"
+	@$(BAZEL) test //... --test_tag_filters=-manual,-integration --cache_test_results=no || echo "No unit tests found (only integration tests exist)"
```
sbalabanov: why not cache results?

sbalabanov (changes requested):

entities/queue/delivery.go
behinddwalls: done

entities/queue/message.go
```diff
@@ -0,0 +1,75 @@
...
+	// - SQS FIFO: MessageGroupId
+	// Messages with the same PartitionKey are guaranteed to be ordered.
+	// Optional - if empty, backend may use round-robin or random distribution.
+	PartitionKey string
+
+	// PublishedAt is when the message was published to the queue.
+	// For messages created locally, this is set to time.Now().
+	// For messages received from queue backends, this should be set from the
+	// backend's timestamp (e.g., SQS SentTimestamp, Kinesis ApproximateArrivalTimestamp).
+	PublishedAt time.Time
```
sbalabanov: For DTO objects, timestamps (unix millies as int64) might be a better choice than `time.Time`
- they are platform-agnostic
- they are better copyable (Time stores a reference to a timezone, stressing GC)
- they do not require an implicit convention of whether the backend should also store the original timezone or not; neither what timezone should be set by default
A helper method can be made to translate to Time.

entities/queue/message.go
```diff
@@ -0,0 +1,75 @@
...
+// The returned message has copied Payload and Metadata, and shares the same PublishedAt.
+// Safe to call concurrently.
+func (m Message) Copy() Message {
+	// Deep copy payload
+	payloadCopy := make([]byte, len(m.Payload))
+	copy(payloadCopy, m.Payload)
+
+	// Deep copy metadata (maps.Clone returns nil if input is nil, so handle that)
+	var metadataCopy map[string]string
+	if m.Metadata != nil {
```
sbalabanov: I believe it should never be null if a) the object is immutable and b) NewMessage is always used?
runtime nil checks, in general, are waste

entities/queue/message.go
```diff
@@ -0,0 +1,75 @@
...
+	return Message{
+		ID:          id,
+		Payload:     payload,
+		Metadata:    make(map[string]string),
+		PublishedAt: time.Now().UTC(),
+	}
+}
+
+// Copy creates a deep copy of the message.
+// The returned message has copied Payload and Metadata, and shares the same PublishedAt.
```
sbalabanov: > shares the same PublishedAt

technically, it is is still a copy, the reference to the timezone will be pointing the same object though

extensions/queue/publisher.go
```diff
@@ -0,0 +1,18 @@
...
+
+	"github.com/uber/submitqueue/entities/queue"
+)
+
+// Publisher publishes messages to topics/queues.
+// Implementations must be thread-safe.
+type Publisher interface {
+	// Publish sends one or more messages to the specified topic.
+	// Returns error if any message fails to publish.
+	Publish(ctx context.Context, topic string, messages ...queue.Message) error
```
sbalabanov: If we allow the interface to publish multiple messages at once, we will likely have to deal with the case of partial success/error.
Might be easier to accept a single message and let the client process errors as they see fit. Unless it is beneficial for some reason to batch them.

sbalabanov: Unless we put a constraint on implementation that all messages should be published, or none.

extensions/queue/queue.go
```diff
@@ -0,0 +1,17 @@
+package queue
+
+// Factory creates and manages queue publishers and subscribers.
+// Implementations handle connection pooling, consumer group configuration,
+// and resource lifecycle.
+type Factory interface {
```
sbalabanov: Should we just name it `Queue` ? :)

extensions/queue/subscriber.go
```diff
@@ -0,0 +1,34 @@
...
+)
+
+// Subscriber consumes messages from topics/queues.
+type Subscriber interface {
+	// Subscribe starts consuming messages from the specified topic.
+	// Returns:
+	//   - A read-only channel that delivers Delivery instances
+	//   - An error if subscription fails
+	//
+	// The channel is closed when the subscriber is closed or context is cancelled.
```
sbalabanov, line 13: what happens on infra error of the queue? We likely need to signal it back.
Alternatively, put a constraint on the implementation to handle all the errors (i.e. try reconnect).

extensions/queue/subscriber.go
```diff
@@ -0,0 +1,34 @@
...
+	//       msg := delivery.Message()
+	//       if err := process(msg); err != nil {
+	//           delivery.Nack(ctx, 0)
+	//           continue
+	//       }
+	//       delivery.Ack(ctx)
+	//   }
+	Subscribe(ctx context.Context, topic string) (<-chan Delivery, error)
+
+	// Close gracefully shuts down the subscriber, stopping message consumption.
```
sbalabanov: plz indicate what the implementation should do (i.e. close all Delivery channels), also if it should be synchronized with Subscribe or not.
Whether it should be idempotent.

extensions/queue/subscriber.go
```diff
@@ -0,0 +1,34 @@
...
+	//   if err != nil { ... }
+	//   for delivery := range deliveryCh {
+	//       msg := delivery.Message()
+	//       if err := process(msg); err != nil {
+	//           delivery.Nack(ctx, 0)
+	//           continue
+	//       }
+	//       delivery.Ack(ctx)
+	//   }
+	Subscribe(ctx context.Context, topic string) (<-chan Delivery, error)
```
sbalabanov, line 18: Indicate if thread safe or not

sbalabanov (changes requested):

entities/queue/message.go
behinddwalls: Sounds good, changed to inte64

entities/queue/message.go
behinddwalls: made it int

extensions/queue/queue.go
behinddwalls: done

extensions/queue/subscriber.go
behinddwalls: Added docs

extensions/queue/subscriber.go
behinddwalls: Add a line, but implementations should ensure, it's thread safe.

extensions/queue/subscriber.go
behinddwalls: I think implementation should handle those errors

extensions/queue/publisher.go
behinddwalls: removed batches, i think one by one is fine here

entities/queue/message.go
behinddwalls: removed nil check

extensions/queue/subscriber.go
```diff
@@ -0,0 +1,24 @@
...
+// Implementations must be thread-safe.
+type Subscriber interface {
+	// Subscribe starts consuming messages from the specified topic.
+	// Returns a channel of Delivery instances and an error if subscription fails.
+	//
+	// The channel is closed when the subscriber is closed or context is cancelled.
+	// Implementations should handle infrastructure errors internally (e.g., reconnect).
+	//
+	// Each Delivery provides the message and methods to acknowledge or reject it.
+	// Consumers should call delivery.Ack() or delivery.Nack() for each delivery.
```
sbalabanov, line 17: nit: also define when the implementation can return an error.

sbalabanov (approved):

PR #20, opened 2026-02-17
feat(queue/sql): add MySQL schema and configuration

behinddwalls:
## Why?
Need database schema and configuration for SQL-based queue implementation to support distributed message processing.

## What?
- MySQL schema with 4 tables: messages, offsets, partition leases, and DLQ
- Configuration struct with validation for consumer groups, timeouts, and retry policies
- Bazel build integration

## Test Plan
- Config validation tests pass
- Default config values are correct
- Invalid configs are rejected appropriately

## Stack
1. @ #20
1. #22

### DIFF
--- a/extensions/queue/sql/config_test.go
+++ b/extensions/queue/sql/config_test.go
@@ -1,0 +1,113 @@
+package sql
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestDefaultConfig(t *testing.T) {
+	cfg := DefaultConfig("test-consumer", "test-worker")
+
+	assert.Equal(t, "test-consumer", cfg.ConsumerGroup)
+	assert.Equal(t, "test-worker", cfg.WorkerID)
+	assert.Equal(t, 100*time.Millisecond, cfg.PollInterval)
+	assert.Equal(t, 10, cfg.BatchSize)
+	assert.Equal(t, 60*time.Second, cfg.VisibilityTimeout)
+	assert.Equal(t, 10*time.Second, cfg.LeaseRenewalInterval)
+	assert.Equal(t, 30*time.Second, cfg.LeaseDuration)
+	assert.True(t, cfg.DLQ.Enabled)
+	assert.Equal(t, 3, cfg.Retry.MaxAttempts)
+	assert.Equal(t, 1*time.Second, cfg.Retry.InitialBackoff)
+	assert.Equal(t, 30*time.Second, cfg.Retry.MaxBackoff)
+	assert.Equal(t, 2.0, cfg.Retry.BackoffMultiplier)
+}
+
+func TestConfigValidation(t *testing.T) {
+	tests := []struct {
+		name        string
+		config      Config
+		expectError bool
+		errorMsg    string
+	}{
+		{
+			name:        "valid config",
+			config:      DefaultConfig("test-consumer", "test-worker"),
+			expectError: false,
+		},
+		{
+			name: "empty consumer group",
+			config: Config{
+				ConsumerGroup:        "",
+				WorkerID:             "test-worker",
+				PollInterval:         100 * time.Millisecond,
+				BatchSize:            10,
+				VisibilityTimeout:    60 * time.Second,
+				LeaseRenewalInterval: 10 * time.Second,
+				LeaseDuration:        30 * time.Second,
+				Retry:                DefaultConfig("dummy", "dummy").Retry,
+			},
+			expectError: true,
+			errorMsg:    "ConsumerGroup is required",
+		},
+		{
+			name: "empty worker ID",
+			config: Config{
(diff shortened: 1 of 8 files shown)


sbalabanov:
Have we tried an existing oss solution (queue over mysql) first?

extensions/queue/sql/schema/queue_dlq.sql
```diff
@@ -0,0 +1,30 @@
...
+    payload BLOB NOT NULL,
+    metadata JSON,
+
+    -- Original timestamps (epoch milliseconds)
+    created_at BIGINT UNSIGNED NOT NULL,
+    published_at BIGINT UNSIGNED NOT NULL,
+
+    -- DLQ-specific fields
+    failed_at BIGINT UNSIGNED NOT NULL,
+    failure_count INT UNSIGNED NOT NULL DEFAULT 0,
```
sbalabanov: prefer avoiding defaults at db level, they tend to hide code flow errors

extensions/queue/sql/schema/queue_messages.sql
```diff
@@ -0,0 +1,37 @@
...
+    -- Messages invisible until this timestamp expires
+    invisible_until BIGINT UNSIGNED NOT NULL DEFAULT 0,
+
+    -- Timestamps (epoch milliseconds)
+    created_at BIGINT UNSIGNED NOT NULL,
+    published_at BIGINT UNSIGNED NOT NULL,
+
+    -- Indexes optimized for partition-based queries with visibility
+    INDEX idx_topic_partition_visible_offset (topic, partition_key, invisible_until, offset),
+    UNIQUE KEY idx_topic_partition_id (topic, partition_key, id)
```
sbalabanov, line 40: Should this be primary key instead?

extensions/queue/sql/schema/queue_messages.sql
```diff
@@ -0,0 +1,37 @@
...
+    -- Visibility timeout (epoch milliseconds)
+    -- Messages invisible until this timestamp expires
+    invisible_until BIGINT UNSIGNED NOT NULL DEFAULT 0,
+
+    -- Timestamps (epoch milliseconds)
+    created_at BIGINT UNSIGNED NOT NULL,
+    published_at BIGINT UNSIGNED NOT NULL,
+
+    -- Indexes optimized for partition-based queries with visibility
+    INDEX idx_topic_partition_visible_offset (topic, partition_key, invisible_until, offset),
```
sbalabanov, line 36: I suggest for each index to add a comment what functionality it is to serve. Indexes scale very poorly. Each needs to be justified.

extensions/queue/sql/config_test.go
```diff
@@ -0,0 +1,119 @@
...
+			errorMsg:    "BatchSize must be positive",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := tt.config.Validate()
+			if tt.expectError {
+				require.Error(t, err)
+				assert.Contains(t, err.Error(), tt.errorMsg)
```
sbalabanov, line 107: AI likes to generate string comparisons for error instead of error types (because typically there are no error types); I think this is a bad practice that makes tests fragile.
Just asserting the error is usually ok; when required by code flow, can dedicate an error type and use Is/As semantics.

extensions/queue/sql/config_test.go
```diff
@@ -0,0 +1,119 @@
...
+			} else {
+				require.NoError(t, err)
+			}
+		})
+	}
+}
+
+func TestTableConstants(t *testing.T) {
```
sbalabanov: this is a change detector test

sbalabanov (changes requested):

extensions/queue/sql/config_test.go
behinddwalls: i think it sort of validate which error is being returned and not short-circuiting due to any other errors in the call chain

extensions/queue/sql/schema/queue_messages.sql
behinddwalls: offset is the primary which is easier to reason it seems as per AI, primarily how it auto-increments and avoids page fragmentation,  so having this secondary index is not too bad for this table.

extensions/queue/sql/schema/queue_messages.sql
behinddwalls: Also, we can revisit these indexes if we need to change or delete some of these

extensions/queue/sql/config_test.go
sbalabanov: If error message is not part of the protocol, this makes test fragile to changes for not reason dictated by a change.
If it is part of the protocol, the general convention is for it to be anywhere in the chain and discoverable with errors.Is/errors.As, and I believe the test should follow the convention.

sbalabanov (approved):
Accepting, but very opinionated that assert.Contains should not be used - can this be removed?

PR #21, opened 2026-02-17
feat(queue/sql): add data access layer stores

behinddwalls:
## Why?
Need data access layer to abstract database operations for message storage, offset tracking, and partition leasing in the SQL queue implementation.

## What?
- MessageStore: insert, fetch, delete, DLQ, and visibility timeout operations
- OffsetStore: consumer offset tracking with atomic ack
- PartitionLeaseStore: distributed partition leasing with automatic discovery

### DIFF
--- a/extensions/queue/sql/offset_store.go
+++ b/extensions/queue/sql/offset_store.go
@@ -1,0 +1,142 @@
+package sql
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+	"time"
+
+	"github.com/uber-go/tally/v4"
+	"go.uber.org/zap"
+)
+
+
+// sqloffsetStore is the SQL implementation of offsetStore
+type sqloffsetStore struct {
+	db      *sql.DB
+	config  Config
+	logger  *zap.SugaredLogger
+	metrics tally.Scope
+}
+
+// Metric names for offset store
+const (
+	metricAckMessageErrors = "ack_message.errors"
+)
+
+// newOffsetStore creates a new SQL offset store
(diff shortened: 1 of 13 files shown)


extensions/queue/sql/offset_store.go
```diff
@@ -0,0 +1,134 @@
...
+		return fmt.Errorf("failed to update offset: %w", err)
+	}
+
+	if err := tx.Commit(); err != nil {
+		s.metrics.Tagged(map[string]string{tagErrorType: "commit"}).Counter(metricAckMessageErrors).Inc(1)
+		return fmt.Errorf("failed to commit transaction: %w", err)
+	}
+
+	// Log and emit metrics after transaction completes
+	s.metrics.Timer("ack_message.latency").Record(time.Since(start))
```
sbalabanov: we effectively only record successful latencies. Is it what we want?
I prefer to record everything with the additional tag indicating result of the operation.

extensions/queue/sql/partition_lease_store.go
```diff
@@ -0,0 +1,313 @@
...
+}
+
+// TryAcquireLease attempts to acquire or renew a lease for a partition
+func (s *sqlPartitionLeaseStore) TryAcquireLease(ctx context.Context, topic string, partitionKey string) (bool, error) {
+	start := time.Now()
+	defer func() {
+		s.metrics.Timer("try_acquire_lease.latency").Record(time.Since(start))
+	}()
+
+	now := time.Now().UnixMilli()
```
sbalabanov: could use `start.UnixMilli()`

sbalabanov (approved):

PR #22, opened 2026-02-17
feat(queue/sql): add publisher implementation

behinddwalls:
### Why?
Need Publisher implementation to enable message publishing to SQL queue topics.

### What?

- Publisher validates topic names and publishes single messages via MessageStore
- Thread-safe with RWMutex for concurrent publish calls
- Idempotent Close() operation
- Comprehensive test coverage for publish, validation, metrics, and concurrency

- Single and multiple message publishing tested
- Invalid topic names rejected (uppercase, special chars, empty)
- Publisher closed state prevents further publishes
- Concurrent publish operations verified thread-safe
- Context cancellation handled correctly

## Test Plan
make test

## Stack
1. @ #22
1. #23
1. #21
1. #24
1. #34


extensions/queue/sql/publisher.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	// Validate topic name (SQL-safe)
+	if err := validateTopicName(topic); err != nil {
+		p.logger.Errorw("publish failure: invalid topic name", "topic", topic, "error", err)
+		p.metrics.Tagged(map[string]string{"topic": topic}).Counter("publish_errors").Inc(1)
+		return err
+	}
+
+	if err := p.messageStore.Insert(ctx, topic, []queue.Message{message}); err != nil {
```
sbalabanov, line 52: why is this under mutex?

extensions/queue/sql/publisher.go
```diff
@@ -0,0 +1,77 @@
...
+	return nil
+}
+
+// Close gracefully shuts down the publisher
+func (p *publisher) Close() error {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	p.closed = true
+	p.logger.Info("publisher closed")
```
sbalabanov: better outside mutex

extensions/queue/sql/publisher.go
```diff
@@ -0,0 +1,77 @@
...
+	return nil
+}
+
+// validateTopicName ensures topic name is safe for use as SQL table name
+func validateTopicName(topic string) error {
```
sbalabanov: unite with subscriber check?

extensions/queue/sql/publisher_test.go
```diff
@@ -0,0 +1,359 @@
...
+		topic     string
+		messages  []queue.Message
+		wantErr   bool
+		setupMock func(*mocks.MockMessageStore)
+	}{
+		{
+			name:  "publish single message",
+			topic: "test_topic",
+			messages: []queue.Message{
+				{ID: "msg1", Payload: []byte("payload1"), PartitionKey: "part1", PublishedAt: time.Now().UnixMilli()},
```
sbalabanov: prefer fixed timestamps in a test for repeatability

extensions/queue/sql/publisher_test.go
```diff
@@ -0,0 +1,359 @@
...
+	// Close the publisher
+	err := pub.Close()
+	require.NoError(t, err)
+
+	// Try to publish after close
+	msg := queue.NewMessage("msg1", []byte("payload"))
+	msg.PartitionKey = "part1"
+	err = pub.Publish(ctx, "test_topic", msg)
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "publisher is closed")
```
sbalabanov: ditto

extensions/queue/sql/publisher_test.go
```diff
@@ -0,0 +1,359 @@
...
+				}
+				errCh <- pub.Publish(ctx, topic, msg)
+			}
+		}(i)
+	}
+
+	// Wait for all goroutines
+	for i := 0; i < numGoroutines*messagesPerGoroutine; i++ {
+		err := <-errCh
+		require.NoError(t, err)
```
sbalabanov, line 334: ok to keep this test as is, however concurrency tests are best when not just the error is validated, but the integrity of the data itself. Which means, both concurrent writes and reads should lead to the desired result without data races affecting integrity.

sbalabanov (changes requested):

extensions/queue/sql/publisher.go
behinddwalls: done

extensions/queue/sql/publisher.go
behinddwalls: fixed

extensions/queue/sql/publisher_test.go
behinddwalls: fixed

extensions/queue/sql/publisher_test.go
behinddwalls: removed message

extensions/queue/sql/publisher.go
behinddwalls: fixed, moved to its own file and will use it in the subscriber as well

extensions/queue/sql/publisher.go
```diff
@@ -0,0 +1,72 @@
...
+	if err := validateTopicName(topic); err != nil {
+		p.logger.Errorw("publish failure: invalid topic name", "topic", topic, "error", err)
+		p.metrics.Tagged(map[string]string{"topic": topic}).Counter("publish_errors").Inc(1)
+		return fmt.Errorf("publish failure: invalid topic name. err: %w", err)
+	}
+
+	if err := p.messageStore.Insert(ctx, topic, []queue.Message{message}); err != nil {
+		p.metrics.Tagged(map[string]string{"topic": topic}).Counter("publish_errors").Inc(1)
+		p.logger.Errorw("publish failure: message store insert error", "topic", topic, "error", err)
+		return fmt.Errorf("publish failure: message store insert error. err: %w", err)
```
sbalabanov: go idiomatic is this format (colon is error chain delimiter)
return fmt.Errorf("publish message store insert error: %w", err)

sbalabanov (approved):

PR #23, opened 2026-02-17
feat(queue/sql): add subscriber with partition leasing

behinddwalls:
## Why?
Need Subscriber implementation with partition leasing to enable distributed message processing from SQL queue.

## What?
- Subscriber discovers partitions and acquires leases for distributed work
- Partition leasing ensures each partition processed by exactly one worker
- Automatic failover via stale lease stealing
- Visibility timeout with automatic retry on worker crash
- DLQ integration for messages exceeding retry limit
- Thread-safe delivery acknowledgment
- Moved Store in separate package to avoid circular dependency with mocks

## Test Plan
bazel test


extensions/queue/sql/offset_store.go
```diff
@@ -0,0 +1,172 @@
...
+// AckMessage atomically deletes a message and updates the acked offset
+func (s *sqlOffsetStore) AckMessage(ctx context.Context, topic string, partitionKey string, messageID string, offset int64, messageStore MessageStore) error {
+	start := time.Now()
+	defer func() {
+		s.metrics.Timer("ack_message.latency").Record(time.Since(start))
+	}()
+
+	tx, err := s.db.BeginTx(ctx, nil)
+	if err != nil {
+		s.logger.Errorw("failed to begin transaction for ack",
```
sbalabanov: do we need to log error if it would be returned and probably logged again later?

extensions/queue/sql/offset_store.go
```diff
@@ -0,0 +1,172 @@
...
+		return fmt.Errorf("failed to begin transaction: %w", err)
+	}
+	defer tx.Rollback()
+
+	// Delete message
+	_, err = tx.ExecContext(ctx, fmt.Sprintf(`
+		DELETE FROM %s WHERE topic = ? AND partition_key = ? AND id = ?
+	`, MessagesTableName), topic, partitionKey, messageID)
+	if err != nil {
+		s.logger.Errorw("failed to delete message",
```
sbalabanov: prefer logging and metrics outside the transaction.
collect data, then do once transaction is done.

extensions/queue/sql/offset_store_test.go
```diff
@@ -0,0 +1,361 @@
...
+		validate     func(t *testing.T, topic, partition string)
+	}{
+		{
+			name:         "initialize new partition",
+			topic:        "test-topic",
+			partitionKey: "partition1",
+			wantErr:      false,
+			validate: func(t *testing.T, topic, partition string) {
+				var offset int64
+				err := db.QueryRow("SELECT offset_acked FROM queue_offsets WHERE consumer_group = ? AND topic = ? AND partition_key = ?",
```
sbalabanov: tests are change detectors. May be we only need integration/e2e tests  for the queue.

extensions/queue/sql/partition_lease_store.go
```diff
@@ -0,0 +1,326 @@
...
+	"database/sql"
+	"fmt"
+	"time"
+
+	"github.com/uber-go/tally/v4"
+	"go.uber.org/zap"
+)
+
+// PartitionLeaseStore handles partition lease operations
+type PartitionLeaseStore interface {
```
sbalabanov: do we need these interfaces to be public?

extensions/queue/sql/partition_lease_store.go
```diff
@@ -0,0 +1,326 @@
...
+		s.logger.Errorw("failed to release lease",
+			logTopic, topic,
+			logPartitionKey, partitionKey,
+			logError, err,
+		)
+		s.metrics.Tagged(map[string]string{tagErrorType: "exec_release"}).Counter("release_lease.errors").Inc(1)
+		return fmt.Errorf("failed to release lease: %w", err)
+	}
+
+	s.metrics.Counter("release_lease.success").Inc(1)
```
sbalabanov: may be only if `RowsAffected > 0` if you want this function to be idempotent?

extensions/queue/sql/partition_lease_store.go
```diff
@@ -0,0 +1,326 @@
...
+// Returns the number of new leases acquired
+func (s *sqlPartitionLeaseStore) DiscoverAndAcquirePartitions(ctx context.Context, topic string) (int, error) {
+	start := time.Now()
+	defer func() {
+		s.metrics.Timer("discover_and_acquire.latency").Record(time.Since(start))
+	}()
+
+	// Query distinct partition_keys from messages table
+	rows, err := s.db.QueryContext(ctx, fmt.Sprintf(`
+		SELECT DISTINCT partition_key FROM %s WHERE topic = ? LIMIT 100
```
sbalabanov: why 100 is the limit?
this at least needs somehow documented

extensions/queue/sql/publisher.go
```diff
@@ -23,23 +23,22 @@ type publisher struct {
...
-		p.logger.Errorw("publish failure: invalid topic name", "topic", topic, "error", err)
+		p.logger.Errorw("publish failed: invalid topic name", "topic", topic, "error", err)
 		p.metrics.Tagged(map[string]string{"topic": topic}).Counter("publish_errors").Inc(1)
 		return err
 	}
 
 	if err := p.messageStore.Insert(ctx, topic, []queue.Message{message}); err != nil {
 		p.metrics.Tagged(map[string]string{"topic": topic}).Counter("publish_errors").Inc(1)
-		p.logger.Errorw("publish failure: message store insert error", "topic", topic, "error", err)
 		return err
```
sbalabanov: wrap with context?

extensions/queue/sql/publisher.go
sbalabanov: "debug" seems more appropriate

extensions/queue/sql/test_helpers.go
```diff
@@ -0,0 +1,153 @@
...
+	contextFactory := func(ctx context.Context, opts ...sqldb.ContextOption) *sqldb.Context {
+		return sqldb.NewContext(ctx, opts...)
+	}
+
+	s, err := server.NewServer(config, engine, contextFactory, sessionBuilder, nil)
+	require.NoError(t, err)
+
+	// Start server in background
+	go func() {
+		_ = s.Start()
```
sbalabanov: goroutine needs to be collected at some point

extensions/queue/sql/test_helpers.go
```diff
@@ -0,0 +1,153 @@
...
+	sqlDB, err := sql.Open("mysql", dsn)
+	require.NoError(t, err)
+
+	// Verify connection
+	ctx := context.Background()
+	for i := 0; i < 50; i++ {
+		if err := sqlDB.PingContext(ctx); err == nil {
+			break
+		}
+		time.Sleep(100 * time.Millisecond)
```
sbalabanov: can use `require.Eventually` instead, but overall relying on hard coded timeouts is not a good design. I'd prefer to rely on test timeout, at least this gives more prominent signal (timeout, not failure) on busy CI.

sbalabanov: Also, it still proceeds even if no ping is successful

extensions/queue/sql/test_helpers.go
```diff
@@ -0,0 +1,153 @@
...
+func createSchema(t *testing.T, db *sql.DB) {
+	t.Helper()
+
+	// Schema files in order
+	schemaFiles := []string{
+		"schema/queue_messages.sql",
+		"schema/queue_partition_leases.sql",
+		"schema/queue_offsets.sql",
+		"schema/queue_dlq.sql",
+	}
```
sbalabanov: https://github.com/uber/submitqueue/pull/29 is an example on how to avoid hardcoding these

extensions/queue/sql/test_helpers.go
```diff
@@ -0,0 +1,153 @@
...
+					line = line[:idx]
+				}
+				cleanLines = append(cleanLines, line)
+			}
+		}
+
+		cleanSQL := strings.Join(cleanLines, "\n")
+
+		// Split by semicolon and execute each statement
+		statements := strings.Split(cleanSQL, ";")
```
sbalabanov: can't it just execute them all with the semicolon?

MODULE.bazel
```diff
@@ -27,16 +27,27 @@ go_sdk.host()
 go_deps = use_extension("@gazelle//:extensions.bzl", "go_deps")
 go_deps.from_file(go_mod = "//:go.mod")
 
+# Override flatbuffers importpath: dolthub fork uses module path github.com/dolthub/flatbuffers/v23,
```
sbalabanov: what is the reason to use dolthub at all?
it seems to bring in a gazillion of dependencies

sbalabanov (changes requested):

MODULE.bazel
behinddwalls: Yes..this one for mysql inmemory for unit tests..though i think i can just mock the call to execute rather than testing in-memory

extensions/queue/sql/partition_lease_store.go
behinddwalls: they don't need to, let me fix it in a follow up PR

extensions/queue/sql/partition_lease_store.go
behinddwalls: fixed here: https://github.com/uber/submitqueue/pull/21/changes#diff-de3cc7ded3b119f6b765a7b907130f5f43701f8ef1cf148f7c281f8f5b0c51adR182

extensions/queue/sql/publisher.go
behinddwalls: fixed here https://github.com/uber/submitqueue/pull/22/changes#diff-a8bed93317fa34a7f530b8f2160c0c2062a597d0b12d90b4aab1e5345cf27452R55

extensions/queue/sql/test_helpers.go
behinddwalls: removed

extensions/queue/sql/test_helpers.go
behinddwalls: removed

extensions/queue/sql/test_helpers.go
behinddwalls: removed from unit tests

extensions/queue/sql/stores.go
```diff
@@ -8,8 +8,73 @@ import (
 	"github.com/uber/submitqueue/entities/queue"
 )
 
+// StoreConfig contains configuration needed by store implementations
+
+// MessageRow represents a row from the messages table
+type MessageRow struct {
+	Offset       int64
```
sbalabanov: document fields

extensions/queue/sql/stores.go
```diff
@@ -8,8 +8,73 @@ import (
 	"github.com/uber/submitqueue/entities/queue"
 )
 
+// StoreConfig contains configuration needed by store implementations
+
+// MessageRow represents a row from the messages table
+type MessageRow struct {
```
sbalabanov: why do we need a separate entity at the implementation level? Does it need to be public?

extensions/queue/sql/subscriber.go
```diff
@@ -0,0 +1,542 @@
...
+	d.acknowledged = true
+	return nil
+}
+
+// Nack implements extqueue.Delivery.Nack
+func (d *sqlDelivery) Nack(ctx context.Context, requeueAfterMillis int64) error {
+	d.mu.Lock()
+	defer d.mu.Unlock()
+
+	if d.acknowledged {
```
sbalabanov, line 139: probably deserves a dedicated error type for the client to possibly process

extensions/queue/sql/subscriber.go
```diff
@@ -0,0 +1,542 @@
...
+		leaseStore:    leaseStore,
+		subscriptions: make(map[string]*subscription),
+	}
+}
+
+// Subscribe starts consuming messages from the specified topic
+func (s *subscriber) Subscribe(ctx context.Context, topic string) (<-chan extqueue.Delivery, error) {
+	s.mu.RLock()
+	if s.closed {
+		s.mu.RUnlock()
```
sbalabanov: can avoid doubling call to `s.mu.RUnlock` by saving the result of `s.closed` into a buffer, then unlocking, then checking for the buffer

extensions/queue/sql/subscriber.go
```diff
@@ -0,0 +1,542 @@
...
+			// Fetch and deliver messages from leased partitions
+			s.pollLeasedPartitions(ctx, sub)
+		}
+	}
+}
+
+// renewLeases renews leases for all partitions owned by this worker
+func (s *subscriber) renewLeases(ctx context.Context, topic string) {
+	leasedPartitions, err := s.leaseStore.GetLeasedPartitions(ctx, topic)
+	if err != nil {
```
sbalabanov, line 297: In general we should not suppress errors. If we do, move the suppression as high the stack as possible and add explanations how this is handled.

extensions/queue/sql/subscriber.go
```diff
@@ -0,0 +1,542 @@
...
+		s.logger.Errorw("failed to get leased partitions for renewal",
+			"topic", topic,
+			"error", err,
+		)
+		return
+	}
+
+	for _, partitionKey := range leasedPartitions {
+		if err := s.leaseStore.RenewLease(ctx, topic, partitionKey); err != nil {
```
sbalabanov, line 310: ditto

sbalabanov (approved):

PR #27, opened 2026-02-18
feat(storage) Implement sample MySQL storage engine for Requests

sbalabanov:
## Why?
First attempt to implement real storage while keeping required characteristics like optimistic locking and autogenerated IDs

## What?
extensions/storage/mysql implementation for Request entity and MYSQL schema

## Test Plan
No tests; will implement integration test with real MySQL as a next step


extensions/storage/mysql/schema/request.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE IF NOT EXISTS request (
```
behinddwalls, line 1: do we want to store schema here vs maybe at some common place like `/schema/storage/mysql/request_store.sql` something like this, i am also doing similar for db-queue, `/schema/queue/mysql/schema.sql`

extensions/storage/storage.go
```diff
@@ -15,4 +26,7 @@ var ErrVersionMismatch = errors.New("version mismatch")
 type StoreFactory interface {
```
behinddwalls, line 26: i we renamed the Queue one to be just queue, so maybe we can keep it consistent and name it `Store` and it can return different types of stores?

extensions/storage/mysql/factory.go
```diff
@@ -1,23 +1,71 @@
...
+// MySQLParameters defines the configuration for the MySQL storage factory.
+// TODO: integrate with configuration system.
 type MySQLParameters struct {
-}
+	// DSN is the MySQL Data Source Name.
+	// Format: [username[:password]@][protocol[(address)]]/dbname[?param1=value1&...&paramN=valueN]
+	DSN string
 
-// NewFactory creates a new MySQL storage factory
-func NewFactory(p MySQLParameters) (storage.StoreFactory, error) {
```
behinddwalls, line 10: maybe we rename this as well to `MySqlStore`

extensions/storage/mysql/factory.go
```diff
@@ -1,23 +1,71 @@
...
+
+	if p.MaxOpenConns > 0 {
+		db.SetMaxOpenConns(p.MaxOpenConns)
+	}
+	if p.MaxIdleConns > 0 {
+		db.SetMaxIdleConns(p.MaxIdleConns)
+	}
+	if p.ConnMaxLifetime > 0 {
+		db.SetConnMaxLifetime(p.ConnMaxLifetime)
+	}
```
behinddwalls, line 50: i was thinking this to leave out from here and be just part of `sql.Db` and be configured outside of this project?

extensions/storage/mysql/request_store.go
```diff
@@ -2,35 +2,142 @@ package mysql
...
 
 // NewRequestStore creates a new MySQL-backed RequestStore.
-func NewRequestStore() storage.RequestStore {
-	return &requestStore{}
+func NewRequestStore(db *sql.DB) storage.RequestStore {
+	return &requestStore{db: db}
 }
 
 // Get retrieves a land request by ID. Returns ErrNotFound if the request is not found.
 func (r *requestStore) Get(ctx context.Context, id string) (entities.Request, error) {
```
behinddwalls, line 28:
```suggestion
func (r *requestStore) Get(ctx context.Context, id string) (entity.Request, error) {
```

maybe we rename this to entity (keeping it singular), same for exetentions i guess as well

extensions/storage/mysql/schema/request.sql
sbalabanov: My preference is to store them close to the entity/implementation. If we need to remove the implementation altogether it is as easy as removing the mysql folder.

So for the queue, I'd put into extensions/queue/mysql/schemas.

Besides, queue implementation may not even have schemas.

extensions/storage/storage.go
sbalabanov: I will rename it in one of the next diffs to likely Storage

extensions/storage/mysql/factory.go
sbalabanov: Yup will rename. Ditch factories!

extensions/storage/mysql/factory.go
sbalabanov: There is a comment somewhere - the plan is to move to configuration, once we have it.

extensions/storage/mysql/request_store.go
sbalabanov: Accepted! I'd prefer a follow up diff though.

behinddwalls (approved):

sbalabanov:
ah, now a stacked commit is merged into this branch 👎 Github workflow is so awesomely flawed.

PR #28, opened 2026-02-19
feat(mysql): Wire MySQL to controller and add integration test

sbalabanov:
Complete MySQL implementation with integration test base on TestContainers


sbalabanov:
need to rework e2e test

PR #29, opened 2026-02-19
feat(mysql): Wire MySQL to controller and add integration test

sbalabanov:
Wire MySQL implementation to Controller and example services.
Revamped all of e2e test infra to make things work by using Testcontainers-Go and docker network.
It is quite slow now (~30 sec) but it is kinda hermetic. Will improve runtime in next iterations.


integration_tests/README.md
```diff
@@ -1,169 +1,83 @@
-# End-to-End Tests
+# Integration Tests
```
sbalabanov, line 1: I will rename `integration_tests` folder to `e2e_test` and related wording in the other change

PR #35, opened 2026-02-20
test(queue/sql): add comprehensive e2e integration tests

behinddwalls:
## Summary
## Why?                                                                                                                                                                                    
                                                        
  The SQL queue implementation needs comprehensive integration testing against real MySQL to verify production-like behavior including message workflows, partition management, retries, DLQ 
  handling, and consumer group coordination.                                                                                                                                                 

  ## What?

  - Add 15 e2e integration tests using testcontainers-go covering: basic ops, partitions, visibility timeout, retries, DLQ, concurrency, crash recovery, and consumer groups
  - Refactor test infrastructure: extract MySQL utilities to `testutil/mysql.go` for reusability across test suites
  - Add `integration_tests/queue/README.md` documenting test coverage and running instructions

## Test Plan
`bazel test //integration_tests/queue:queue_test`

## Issues


integration_tests/testutil/mysql.go
```diff
@@ -0,0 +1,115 @@
...
+
+	_ "github.com/go-sql-driver/mysql"
+	"github.com/stretchr/testify/require"
+	"github.com/testcontainers/testcontainers-go"
+	"github.com/testcontainers/testcontainers-go/modules/mysql"
+	"github.com/testcontainers/testcontainers-go/network"
+)
+
+// TestLogger is a simple test-aware logger that records elapsed time between logs.
+type TestLogger struct {
```
sbalabanov, line 20: mysql.go is probably not the best filename for this now

sbalabanov (approved):
We need to discuss how many e2e test suites we want to maintain. All these containers are notoriously slow to setup and start

integration_tests/testutil/mysql.go
behinddwalls: Yes..we could probably just do one container for mysql for the whole suite and use it for all cases...

PR #39, opened 2026-02-22
feat(queue): complete Gateway → Orchestrator queue integration

behinddwalls:
## Summary
Wire up the first stage of the queue pipeline: Gateway publishes land requests
to the queue, Orchestrator consumes and processes them.

**Consumer Infrastructure:**
- Add Consumer interface (Register/Start/Stop) for orchestrating multiple controllers
- Add consumer.Delivery interface to enforce separation of concerns (type-safe)
- Controllers receive consumer.Delivery (no Ack/Nack), Consumer handles ack/nack
- Implement subscription lifecycle, automatic ack/nack, metrics, graceful shutdown

**Gateway:**
- Land controller publishes requests to land_request queue after storage
- Queue infrastructure optional (controlled by QUEUE_MYSQL_DSN env var)

**Orchestrator:**
- Request controller consumes from land_request queue
- Wire up consumer with graceful shutdown in main.go

**CLAUDE.md:**
- Document RPC vs Queue Message controller patterns
- Add code style guidelines: use SugaredLogger, use interfaces for contracts

All unit and integration tests pass. Backward compatible with existing tests.

## Stack
1. @ #39
1. #41


e2e_test/suite_test.go
```diff
@@ -113,17 +113,3 @@ func (s *IntegrationSuite) TestPingSpeculator() {
 	s.log.Logf("Speculator ping: %s", resp.Message)
 }
 
-func (s *IntegrationSuite) TestLandRequest() {
```
behinddwalls, line 116: I am refactoring whole container/testing part, i will add it back in next revision, removing it for as it fails

behinddwalls: Added back here: https://github.com/uber/submitqueue/pull/41

core/consumer/consumer.go
```diff
@@ -0,0 +1,341 @@
+package consumer
```
sbalabanov, line 1: why is this in the root?

sbalabanov: Also, README would be nice

core/consumer/consumer.go
```diff
@@ -0,0 +1,341 @@
...
+}
+
+// Register adds a controller to the consumer. Must be called before Start().
+// Returns error if a controller for the same topic is already registered.
+func (m *consumer) Register(controller Controller) error {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	// Check for duplicate topic registration
+	for _, c := range m.controllers {
```
sbalabanov, line 73: I guess the number is small so O(n) is fine? Deserves a comment at least.

core/consumer/consumer.go
```diff
@@ -0,0 +1,341 @@
...
+}
+
+// subscribe subscribes a controller to its topic and spawns a consumption goroutine.
+func (m *consumer) subscribe(ctx context.Context, controller Controller) error {
+	// Get controller's subscription config
+	config := controller.SubscriptionConfig(m.subscriberName)
+
+	// Subscribe to topic
+	subscriber := m.queue.Subscriber()
+	deliveryChan, err := subscriber.Subscribe(ctx, controller.Topic(), config)
```
sbalabanov, line 135: is this ok to execute this and below under a lock?

consumer/consumer.go
```diff
@@ -0,0 +1,341 @@
...
+		"partition_key", msg.PartitionKey,
+		"attempt", delivery.Attempt(),
+		"elapsed_ms", elapsed.Milliseconds(),
+	)
+}
+
+// Stop gracefully shuts down all handlers with the specified timeout.
+// Cancels all subscription contexts and waits for consumption goroutines to finish.
+// timeoutMs is the maximum time in milliseconds to wait for graceful shutdown.
+func (m *consumer) Stop(timeoutMs int64) {
```
sbalabanov: probably want a `stop` flag to disallow new subscriptions, or document that the user code is responsible to call them in a specific orders

consumer/consumer.go
```diff
@@ -0,0 +1,341 @@
...
+		m.logger.Debugw("stopping controller",
+			"controller", sub.controller.Name(),
+			"topic", topic,
+		)
+		sub.cancelFunc()
+	}
+
+	// Wait for all consumption goroutines to finish (with timeout)
+	done := make(chan struct{})
+	go func() {
```
sbalabanov: can use a signal channel instead of waitgroup to avoid this goroutine

core/consumer/controller.go
```diff
@@ -0,0 +1,92 @@
...
+	Attempt() int
+
+	// ReceivedAt returns when this delivery was received (Unix milliseconds).
+	ReceivedAt() int64
+
+	// Metadata returns backend-specific delivery metadata.
+	Metadata() map[string]string
+}
+
+// deliveryWrapper wraps extension/queue.Delivery and exposes only the safe subset of methods.
```
sbalabanov, line 42: why need for the interface? Will it be reimplemented?

entity/request.go
```diff
@@ -48,21 +50,33 @@ type Request struct {
...
+	State RequestState `json:"state"`
 	// Version is the version of the object. It is used for optimistic locking.
 	// Versioning starts at 1 and is incremented for each change to the object.
-	Version int32
+	Version int32 `json:"version"`
+}
+
+// ToBytes serializes the Request to JSON bytes for queue message payload.
+func (r Request) ToBytes() ([]byte, error) {
+	return json.Marshal(r)
```
sbalabanov, line 74: why not yaml?

example/server/gateway/main.go
```diff
@@ -109,16 +113,52 @@ func run() error {
...
+		gatewayServer = &GatewayServer{
+			pingController: pingController,
+			landController: landController,
+		}
+	} else {
+		logger.Warn("queue infrastructure disabled (QUEUE_MYSQL_DSN not set) - only Ping API available")
+		// Only Ping controller available without queue
+		gatewayServer = &GatewayServer{
+			pingController: pingController,
+			landController: nil, // Land API will return errors if called
```
sbalabanov: I'd prefer to just error out or even panic rather than processing it

example/server/orchestrator/main.go
```diff
@@ -114,12 +170,18 @@ func run() error {
 	select {
 	case <-sigCh:
 		fmt.Println("\nShutting down orchestrator server...")
+		if c != nil {
+			c.Stop(30000) // Stop consumers with 30s timeout
```
sbalabanov: it would be probably better to still return the error indicating whether timeout was hit or not
Move error processing to the highest level possible

gateway/controller/land.go
```diff
@@ -73,16 +77,59 @@ func (c *LandController) Land(ctx context.Context, req *pb.LandRequest) (*pb.Lan
...
 	)
 
+	// Publish to queue for async processing
+	if err := c.publishToQueue(ctx, request); err != nil {
+		c.logger.Errorw("failed to publish request to queue",
+			"queue", req.Queue,
+			"sqid", request.ID,
+			"error", err,
+		)
+		c.metricsScope.Counter("publish_errors").Inc(1)
```
sbalabanov: should this be wrapped in the publisher?

orchestrator/controller/request/request.go
```diff
@@ -0,0 +1,110 @@
...
+	if err != nil {
+		c.logger.Errorw("failed to deserialize request",
+			"message_id", msg.ID,
+			"partition_key", msg.PartitionKey,
+			"attempt", delivery.Attempt(),
+			"error", err,
+		)
+		c.metricsScope.Counter("deserialize_errors").Inc(1)
+		// Ack malformed messages to prevent infinite retry loop
+		return nil
```
sbalabanov: that does not sound right. May be we need a specific error type indicating that the error is not retryable (i.e. message is broken), or wrap deserialization in the consumer, too?

sbalabanov (changes requested):

core/consumer/consumer.go
behinddwalls: I thought of it putting somewhere under some folder but this is generic consumer framework, open to relocating it somewhere else? extensions doesn't really fir given it's agnostic of queue impl

consumer/consumer.go
behinddwalls: Yes..stop make sense here

core/consumer/controller.go
behinddwalls: no..we can do without interface..but we wanted to provide ExtendVisibilityTimeout to the business logic to decide if they need need more time... I can thing about it and make entity but then i need to inject something into controller to allow extending vstimeout

example/server/gateway/main.go
behinddwalls: This was for making that integ test pass..but i removed it so let me update this back

gateway/controller/land.go
behinddwalls: yes..we can remove them..they are already in publisher impl

orchestrator/controller/request/request.go
behinddwalls: Yes..this was dummy logic so i didn't really look much as we will change it, but agreed, if we discover poison pill in controller, we should return a specific error

entity/request.go
behinddwalls: gnerally when i see yaml, mental model is config so json seemed a bit more reasonable for this.

core/consumer/consumer.go
behinddwalls: moved to core/consumer

consumer/consumer.go
behinddwalls: done

entity/request.go
behinddwalls: going with json for serde

sbalabanov:
stamping to unblock.
blockers to resolve discussed offline

sbalabanov (approved):

PR #42, opened 2026-02-23
feat(gateway): add input validation to LandController

djuloori2794:
## What?
Validate required fields before processing land requests:
- Queue name must not be empty
- Change source must not be empty
- At least one change ID must be provided

## Why?
 The LandController.Land() method currently accepts requests without validating required fields.  
 Failures occur deep in the processing pipeline instead of at the entry point with clear messages

## Test Plan                                                                                                                                                                                                       
  - [x] Added `TestLand_ReturnsErrorOnEmptyQueue`                                                                                                                                                                     
  - [x] Added `TestLand_ReturnsErrorOnEmptyChangeSource`                                                                                                                                                              
  - [x] Added `TestLand_ReturnsErrorOnNilChange`
  - [x] Added `TestLand_ReturnsErrorOnEmptyChangeIDs`                                                                                                                                                                 
  - [x] All existing tests pass

## Issue
N/A

### DIFF
--- a/gateway/controller/land.go
+++ b/gateway/controller/land.go
@@ -2,6 +2,7 @@ package controller
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"time"
 
@@ -12,6 +13,15 @@ import (
 	pb "github.com/uber/submitqueue/gateway/protopb"
 	"go.uber.org/zap"
 )
+
+// ErrInvalidRequest is returned when the request fails validation.
+// This error should be mapped to codes.InvalidArgument at the gRPC layer.
+var ErrInvalidRequest = errors.New("invalid request")
+
+// IsInvalidRequest returns true if any error in the error chain is ErrInvalidRequest.
+func IsInvalidRequest(err error) bool {
+	return errors.Is(err, ErrInvalidRequest)
+}
 
 // LandController handles land business logic for the gateway
 type LandController struct {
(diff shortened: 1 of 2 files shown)


gateway/controller/land.go
```diff
@@ -40,6 +40,17 @@ func (c *LandController) Land(ctx context.Context, req *pb.LandRequest) (*pb.Lan
 
 	c.metricsScope.Counter("land_request_count").Inc(1)
 
+	// Validate required fields.
+	if req.Queue == "" {
+		return nil, fmt.Errorf("queue name is required")
```
sbalabanov: may I suggest some specific wording to help better identify where the error came from
i.e. LandController requires the request to have a queue name specified

gateway/controller/land_test.go
```diff
@@ -184,3 +184,91 @@ func TestLand_CounterDomainIncludesQueue(t *testing.T) {
...
+	ctx := context.Background()
+
+	req := &pb.LandRequest{
+		Queue:  "test-queue",
+		Change: &pb.Change{Source: "github", Ids: []string{}},
+	}
+	_, err := controller.Land(ctx, req)
+
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "at least one change ID is required")
```
sbalabanov: Plz do not assert on eroor message
CLAUDE.md has a specific instruction on this one

gateway/controller/land.go
sbalabanov: You also likely want a specific error type to signal back to the server that the error is user error, so it could emit an appropriate GRPC code and metadata

sbalabanov (changes requested):

sbalabanov (approved):

PR #44, opened 2026-02-23
feat(entities) Adds change provider entity and mysql backed store for it

manjari25:
## Summary
feat(entities) Adds change provider entity and mysql backed store for it

## Why?

This sets up change provider entity and its corresponding sql store. This will be used by various workflows in the orchestrator to get information about changes going through the system.

## What?
* Add a `ChangeProvider` entity
* Add a `ChangeProviderStore` - the mysql store that backs this entity

## Stack
1. @ #44
1. #46
1. #47
1. #48
1. #49


entity/change_provider.go
```diff
@@ -0,0 +1,19 @@
+package entity
+
+type ChangeProvider struct {
```
sbalabanov, line 5: plz comment the struct itself

entity/change_provider.go
```diff
@@ -0,0 +1,19 @@
...
+	// ID is the globally unique identifier for the land request. Format: "<queue>/<counter_value>".
+	ID string
+	// Queue is the name of the queue processing the land request.
+	// This is defined in the configuration and should be unique within the system.
+	Queue string
```
sbalabanov: probably not needed here

entity/change_provider.go
```diff
@@ -0,0 +1,19 @@
...
+	// ChangeProviderSrc defines the source of the change. For e.g. - Github, Gitlab etc.
+	ChangeProviderSrc string
+	// ChangeProviderID is the identifier specified by the change provider source. For e.g. - Github PR ID etc.
+	ChangeProviderID string
+	// Metadata is the interesting data from the change provider that we want to store.
+	// This is a freeform JSON object.
+	Metadata map[string]any
```
sbalabanov: may be [string]string to avoid ambiguity at the storage level and conversion errors?

entity/change_provider.go
```diff
@@ -0,0 +1,19 @@
...
+	// Version is the version of the object. It is used for optimistic locking.
+	// Versioning starts at 1 and is incremented for each change to the object.
+	Version int32
```
sbalabanov: ok to keep for now, not sure if we expect changes to this? I thought it should be fully immutable, unless we want to update data from the PRs periodically...

extension/storage/mysql/schema/change_provider.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE IF NOT EXISTS change_provider (
+    id VARCHAR(255) NOT NULL,
+    queue VARCHAR(255) NOT NULL,
```
sbalabanov: queue is likely not needed, we have request id incorporating it

extension/storage/mysql/change_provider_store.go
```diff
@@ -0,0 +1,76 @@
...
+
+	if err := json.Unmarshal(metadataJSON, &cp.Metadata); err != nil {
+		return entity.ChangeProvider{}, fmt.Errorf("failed to unmarshal metadata for change provider entity id=%s from the database: %w", id, err)
+	}
+
+	return cp, nil
+}
+
+// Create creates a new change provider from a request. Returns ErrAlreadyExists if the entry already exists.
+func (s *changeProviderStore) Create(ctx context.Context, request entity.Request) error {
```
sbalabanov: I suggest to make ChangeProvider entity created by the app layer, and db layer is only to store or retrieve

extension/storage/change_provider_store.go
```diff
@@ -0,0 +1,20 @@
...
+)
+
+// ChangeProviderStore is an interface that defines methods for managing change provider information in the database.
+type ChangeProviderStore interface {
+	// Get retrieves information about a change by ID. Returns ErrNotFound if the change provider is
+	// not found.
+	Get(ctx context.Context, id string) (entity.ChangeProvider, error)
+
+	// Create creates a new change provider.
+	Create(ctx context.Context, request entity.Request) error
```
sbalabanov: ditto on above, storage layer store should in general not be aware of any entity other than the one it materializes.

sbalabanov (changes requested):

entity/change_provider.go
manjari25: Done

entity/change_provider.go
manjari25: Removed

entity/change_provider.go
manjari25: Updated

entity/change_provider.go
manjari25: I removed this - version is not needed for this entity.

extension/storage/mysql/schema/change_provider.sql
manjari25: Removed

extension/storage/mysql/change_provider_store.go
manjari25: I have changed this signature already.

extension/storage/change_provider_store.go
manjari25: I have changed this signature already.

extension/storage/mysql/change_provider_store.go
```diff
@@ -0,0 +1,69 @@
...
+type changeProviderStore struct {
+	db *sql.DB
+}
+
+// NewChangeProviderStore creates a new MySQL-backed ChangeProviderStore.
+func NewChangeProviderStore(db *sql.DB) storage.ChangeProviderStore {
+	return &changeProviderStore{db: db}
+}
+
+// Get retrieves a change provider by ID. Returns ErrNotFound if the change provider is not found.
```
sbalabanov: I guess it needs to retrieve all of the changes associated with the request? Also in the order specified in the request.

sbalabanov: So it has to be an array or a separate entity

sbalabanov (changes requested):

extension/storage/mysql/schema/change_provider.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE IF NOT EXISTS change_provider (
+    id VARCHAR(255) NOT NULL,
```
behinddwalls: maybe we call it request_id

manjari25: Done

extension/storage/mysql/change_provider_store.go
manjari25: Done

extension/storage/mysql/change_provider_store.go
```diff
@@ -0,0 +1,82 @@
...
+type changeProviderStore struct {
+	db *sql.DB
+}
+
+// NewChangeProviderStore creates a new MySQL-backed ChangeProviderStore.
+func NewChangeProviderStore(db *sql.DB) storage.ChangeProviderStore {
+	return &changeProviderStore{db: db}
+}
+
+// Get retrieves change provider(s) by request ID. Returns ErrNotFound if the change provider is not found.
```
behinddwalls, line 25: can we add a comment saying the returned order may not be same as submitted so application must handle if ordering is needed

manjari25: Updated

behinddwalls (approved):

extension/storage/mysql/change_provider_store.go
```diff
@@ -0,0 +1,88 @@
...
+//
+// Note: The order of ChangeProvider entities returned here is not guaranteed
+// to be the same as the request to which it belongs. The caller is repsonsible
+// for inspecting and mapping the result of this function to the
+// order of changes within the original request.
+//
```
behinddwalls, line 31: nit: empty line

extension/storage/mysql/schema/change_provider.sql
sbalabanov: sqid :)
also in the `Requests` table :)

sbalabanov:
It seems that the comment about retrieveing an array of changes (or at least change ids) was not addressed?
Each request may specify more than one changeid, like in a case of stacked diffs.

PR #46, opened 2026-02-23
feat(entities) Adds batch entity and batch sql schema

manjari25:
## Summary
feat(entities) Adds batch entity and batch sql schema

## Why?

This sets up batch entity and its corresponding sql schema. This establishes the core concept of a batch that will be used through the system.

## What?
* Add a `Batch` entity
* Add `batch` mysql schema
* Note: `BatchStore` in upcoming PR

## Stack
1. #44
1. @ #46
1. #47
1. #48
1. #49

### DIFF
--- a/entity/batch.go
+++ b/entity/batch.go
@@ -1,5 +1,14 @@
 package entity
 
+// BatchState defines the possible states of a batch.
+type BatchState string
+
+const (
+	// BatchStateUnknown is the unreachable state. It is set by default when the structure is initialized. It should never be seen in the system.
+	BatchStateUnknown BatchState = ""
+	// TODO: Add comprehensive list of known batch states.
+)
+
 // Batch represents a group of requests to land (merge into target branch of the source control repository).
 type Batch struct {
 	// ID is the globally unique identifier for the batch. Format: "<queue>/batch/<counter_value>".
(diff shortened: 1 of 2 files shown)


entity/batch.go
```diff
@@ -0,0 +1,21 @@
...
+type Batch struct {
+	// ID is the globally unique identifier for the batch. Format: "<queue>/batch/<counter_value>".
+	ID string
+	// Queue is the name of the queue processing the land request. Queue name is defined in the configuration and should be unique within the system.
+	Queue string
+	// Contains is a list of land request IDs that are part of this batch.
+	// Request IDs will always be part of the same queue.
+	// For e.g. - [queueA/1, queueA/2, queueA/3].
+	Contains []string
+	// Dependencies is a list of batch IDs (and associated metdata) for this batch.
```
sbalabanov: typo

entity/batch.go
```diff
@@ -0,0 +1,21 @@
...
+	// Dependencies will always be part of the same queue.
+	Dependencies []map[string]interface{}
```
sbalabanov, line 35: may be need a more thorough explanation what deps are

entity/batch.go
```diff
@@ -0,0 +1,21 @@
...
+	// The state of the batch lifecycle this batch is in.
+	State string
```
sbalabanov: prefer to type it explicitly off string type

extension/storage/mysql/schema/batch.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE IF NOT EXISTS batch (
+    id VARCHAR(255) NOT NULL,
+    queue VARCHAR(255) NOT NULL,
+    contains JSON NOT NULL,
+    dependencies JSON NOT NULL,
+    state VARCHAR(255) NOT NUll,
+    version INT NOT NULL,
+    PRIMARY KEY (id,queue)
```
sbalabanov: id is enough

sbalabanov (changes requested):

entity/batch.go
manjari25: Updated

extension/storage/mysql/schema/batch.sql
manjari25: Updated

entity/batch.go
manjari25: Updated

entity/batch.go
manjari25: Updated

behinddwalls (approved):

PR #48, opened 2026-02-24
feat(entities) Adds build entity and sql schema

manjari25:
## Summary
feat(entities) Adds build entity and sql schema

## Why?

This sets up build entity and its corresponding sql schema. This establishes a concrete entity for builds that will be used across the system.

## What?
* Add a `Build` entity
* Add `build` mysql schema
* Note: `BuildStore` in upcoming PR


sbalabanov (changes requested):
Seems like a lot of irrelevant commits got merged here accidently?
I'll let you fix this (or let know which specific commit(s) to review) before reviewing.

behinddwalls (approved):

entity/build.go
```diff
@@ -0,0 +1,36 @@
...
+	// BuildStateUnknown is the unreachable state. It is set by default when the structure is initialized. It should never be seen in the system.
+	BuildStateUnknown BuildStatus = ""
+	// TODO: Add comprehensive list of known build states.
+)
+
+// SpeculationPathInfo represents the base and head commits of a speculation path used in a build.
+type SpeculationPathInfo struct {
+	// Base is a list of batchIDs(in order) that form the base of this speculation path.
+	Base []string
+	// Head represents the head commit of this speculation path.
```
sbalabanov: Why do we need a head commit?
Build likely needs to
1. Apply all changes from all batches in order of speculation
2. Build/test targets changed targets in **all** of the commits from **all** of the requests from the **last** batch

entity/build.go
```diff
@@ -0,0 +1,36 @@
...
+	Head string
+}
+
+// Build represents a build scheduled for a batch along a specific speculation path.
+// All fields except the Status are immutable after creation.
+type Build struct {
+	// ID represents the build ID. It is the responsibility of the caller to ensure
```
sbalabanov: Isn't BuildID generated by the build system (i.e. Buildkite)?

sbalabanov (changes requested):

entity/build.go
manjari25: Yes - it is. What is the concern here?

entity/build.go
manjari25: Yes - this is correct. I added it so this is modeled explicitly. I don't have a strong opinion. Can remove this if needed.

sbalabanov: Let's only do what is absolutely needed. Probably a list of BatchIDs is enough?

entity/build.go
sbalabanov: It says "the responsibility of the caller", in which case the caller is SubmitQueue.
So it is a responsibility of a build management system.

sbalabanov (approved):
Accepting with comments

entity/build.go
manjari25: Updated

entity/build.go
manjari25: Removed the head field.

PR #49, opened 2026-02-24
feat(entities) Adds speculation tree entity and sql schema

manjari25:
## Summary
feat(entities) Adds speculation tree entity and sql schema

## Why?

This sets up a speculation tree entity and its corresponding sql schema. This establishes a concrete entity for speculation tree that will be used across the system.

## What?
* Add a `SpeculationTree` entity
* Add `speculation_tree` mysql schema
* Note: `SpeculationTreeStore` in upcoming PR

### DIFF
--- a/entity/speculation_tree.go
+++ b/entity/speculation_tree.go
@@ -8,5 +8,15 @@ type SpeculationTree struct {
 	Queue string
 	// Speculations is a list of speculation paths for this batch based on a graph of its
 	// dependents.
+	//
+	// For e.g - Consider batches - queueA/batch/1, queueA/batch/2, queueA/batch/3
+	// such that - queueA/batch/2 and queueA/batch/3 depend on queueA/batch/1
+	//
+	// Speculations for queueA/batch/1 - [{"path": "queueA/batch/1", "state": "scheduled", "score": 0.1}]
+	// Speculations for queueA/batch/2 - [{"path": "queueA/batch/2", "state": "scheduled", "score": 0.9}, {"path": "queueA/batch/1//queueA/batch/2", "state": "scheduled", "score": 0.3}]
+	// Speculations for queueA/batch/3 - [{"path": "queueA/batch/3", "state": "scheduled", "score": 0.9}, {"path": "queueA/batch/1//queueA/batch/3", "state": "scheduled", "score": 0.3}]
+	//
+	// Note that the key value pairs within the map could have additional information about the speculation path if needed.
+	//
 	Speculations []map[string]string
 }
--- a/entity/speculation_tree.go
+++ b/entity/speculation_tree.go
@@ -1,9 +1,21 @@
 package entity
 
+// SpeculationPathAction defines the possible actions for a speculation path.
+type SpeculationPathAction string
+
+const (
+	// SpeculationPathActionUnknown is the default zero value for SpeculationPathAction.
+	SpeculationPathActionUnknown SpeculationPathAction = ""
+	// TODO: Add comprehensive list of actions
+)
+
 // SpeculationInfo represents metadata about a single speculation path, including the path through the dependency graph, its current state, and the predicted build score.
 type SpeculationInfo struct {
+	// Path represents the speculation path.
 	Path string
-	State string
+	// Action is a state that this path is in.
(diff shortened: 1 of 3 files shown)


entity/speculation_tree.go
```diff
@@ -0,0 +1,12 @@
+package entity
+
+// SpeculationTree represents the set of speculation paths constructed for a batch based on its dependency graph.
+type SpeculationTree struct {
+	// BatchID is the batch for which this speculation tree is constructed.
+	BatchID string
+	// Queue is the name of the queue processing the land request. Queue name is defined in the configuration and should be unique within the system.
+	Queue string
```
sbalabanov: is it needed here? Batch should have this info

entity/speculation_tree.go
```diff
@@ -0,0 +1,12 @@
...
+	// Speculations is a list of speculation paths for this batch based on a graph of its
+	// dependents.
```
sbalabanov, line 27: please elaborate
i.e. what key is, what value is, give an example of speculation path

sbalabanov (changes requested):

entity/speculation_tree.go
```diff
@@ -0,0 +1,12 @@
...
+	Speculations []map[string]string
```
behinddwalls: do we have anything to store for each speculations? like action ? https://github.com/uber/submitqueue/pull/15/changes#diff-8fdcb81f89023611cfaa67a09cf8abaeeff7aec830049829d5b8c22dce086c82

behinddwalls: Also do we need any status as such should we always call batch table?

entity/speculation_tree.go
manjari25: If we want to query by queue name; this maybe useful.

entity/speculation_tree.go
manjari25: Added

entity/speculation_tree.go
manjari25: I added an example - we can have the status in the map as a key value pair. Let me know if this looks ok.

entity/speculation_tree.go
sbalabanov: Do we? I suggest to add it when we have a use case.It has to have a secondary index, or probably better a separate table anyways.

entity/speculation_tree.go
sbalabanov: Let's make it an array of explicit types

sbalabanov (changes requested):

entity/speculation_tree.go
manjari25: Added

entity/speculation_tree.go
manjari25: Ok - I can remove this.

behinddwalls (approved):

entity/speculation_tree.go
```diff
@@ -0,0 +1,37 @@
...
+const (
+	// SpeculationPathActionUnknown is the default zero value for SpeculationPathAction.
+	SpeculationPathActionUnknown SpeculationPathAction = ""
+	// TODO: Add comprehensive list of actions
+)
+
+// SpeculationInfo represents metadata about a single speculation path, including the path through the dependency graph, its current state, and the predicted build score.
+type SpeculationInfo struct {
+	// Path represents the speculation path.
+	Path string
```
sbalabanov: should be an array

entity/speculation_tree.go
```diff
@@ -0,0 +1,37 @@
...
+	// Action is a state that this path is in.
+	Action SpeculationPathAction
+	// Score is score for this speculation path.
```
sbalabanov, line 18: plz enhance

sbalabanov (approved):

entity/speculation_tree.go
manjari25: Updated as discussed offline.

entity/speculation_tree.go
manjari25: Will update the description in a different PR. Need to resolve this to land the change.

PR #50, opened 2026-02-24
feat(queueconfig): adding entity and basic interface for queue config

kevinlnew:
## Summary
feat(queueconfig): adding entity and basic interface for queue config

why not create an implementation of the extension?
- This is open to users to implement or have in-house solutions; generally, uberfx services can plug-and-play `configfx` with minimal extra configs to provide the struct, so the same idea holds for others. 
  - I'm open to creating a basic yaml reader if necessary.

Why?
Services require stored configuration for queue config data, readable by queue name (which is passed by request). 

What?
- Added a basic config entity for queue config
- Added a basic interface for getting a queueconfig by name.

## Test Plan
Basic unit testing for constructor for queue config entity.


entity/queueconfig/queueconfig.go
```diff
@@ -0,0 +1,68 @@
...
+
+// Destination identifies the landing target in a version control system.
+// The interpretation of Ref is VCS-specific and resolved by the change provider:
+//   - Git: branch name (e.g., "main", "release/v2")
+//   - Perforce: stream or depot path (e.g., "//depot/main/...")
+//   - SVN: repository path (e.g., "trunk/")
+//   - Mercurial: bookmark name (e.g., "main")
+//
+// Immutable after creation.
+type Destination struct {
```
sbalabanov: Here and up: why need to wrap a single field into a structure?

entity/queueconfig/queueconfig.go
```diff
@@ -0,0 +1,68 @@
...
+	// Ref is the version-control-specific reference for the landing target.
+	// Opaque to the system; meaningful only to the change provider.
+	Ref string
+}
+
+// QueueConfig holds the configuration for a single submit queue.
+// Each queue maps a repository + destination to a processing pipeline.
+// A repository can have multiple queues, but each queue has exactly one destination.
+// Immutable after creation.
+type QueueConfig struct {
```
sbalabanov: If it is a single entity, put it higher into "entity" package

entity/queueconfig/queueconfig_test.go
```diff
@@ -0,0 +1,72 @@
...
+
+	cfg := NewQueueConfig("uber/submitqueue/main", repo, dest, "github")
+
+	assert.Equal(t, "uber/submitqueue/main", cfg.Name)
+	assert.Equal(t, "uber/submitqueue", cfg.Repository.ID)
+	assert.Equal(t, "main", cfg.Destination.Ref)
+	assert.Equal(t, "github", cfg.ChangeProviderName)
+}
+
+func TestNewQueueConfig_DifferentPlatforms(t *testing.T) {
```
sbalabanov: not sure if I see value in this test on top what `TestNewQueueConfig` already provides

entity/queueconfig/queueconfig_test.go
```diff
@@ -0,0 +1,72 @@
...
+			cfg := NewQueueConfig(tt.name, tt.repo, tt.destination, tt.provider)
+
+			assert.Equal(t, tt.repo, cfg.Repository)
+			assert.Equal(t, tt.destination, cfg.Destination)
+			assert.Equal(t, tt.provider, cfg.ChangeProviderName)
+		})
+	}
+}
+
+func TestQueueConfig_ZeroValue(t *testing.T) {
```
sbalabanov: there is no functionality that depends on the object to be empty, so test is not very helpful

sbalabanov (changes requested):

entity/queueconfig/queueconfig.go
kevinlnew: My intention was to leave these sub-structs to be more extensible in case there were other config metadata that we want to capture later. Not opposed to just keeping it as simple as possible for now though.

entity/queueconfig/queueconfig.go
```diff
@@ -0,0 +1,67 @@
...
+//   - GitHub/GitLab: "owner/repo" (e.g., "uber/submitqueue")
+//   - Perforce: depot path (e.g., "//depot/project")
+//   - SVN: repository URL (e.g., "https://svn.example.com/repos/project")
+//
+// Immutable after creation.
+type Repository struct {
+	// ID is the platform-specific identifier for the repository.
+	// Opaque to the system; meaningful only to the change provider.
+	ID string
+}
```
behinddwalls: should this be full remote URI? `git@github.com:uber/submitqueue`

entity/queueconfig/queueconfig.go
```diff
@@ -0,0 +1,67 @@
...
+//   - Perforce: stream or depot path (e.g., "//depot/main/...")
+//   - SVN: repository path (e.g., "trunk/")
+//   - Mercurial: bookmark name (e.g., "main")
+//
+// Immutable after creation.
+type Destination struct {
+	// Ref is the version-control-specific reference for the landing target.
+	// Opaque to the system; meaningful only to the change provider.
+	Ref string
+}
```
behinddwalls: can this also be a URI, `git@github.com:uber/submitqueue@main`

entity/queueconfig/queueconfig.go
```diff
@@ -0,0 +1,67 @@
...
+
+// Config holds the configuration for a single submit queue.
+// Each queue maps a repository + destination to a processing pipeline.
+// A repository can have multiple queues, but each queue has exactly one destination.
+// Immutable after creation.
+type Config struct {
```
behinddwalls: should we call it QueueConfig? though it overloaded with MessagingQueue?

entity/queueconfig/queueconfig.go
```diff
@@ -0,0 +1,67 @@
...
+	// Destination is the landing target where changes are merged.
+	Destination Destination
+
+	// ChangeProviderName identifies which change provider to use (e.g., "github", "gerrit").
+	// Resolved to a ChangeProvider instance at wiring time.
+	ChangeProviderName string
+
+	// ChangeProvider is the change provider to use for this queue.
+	// TODO:
+	// ChangeProvider ChangeProvider to be defined in the changeprovider extension package
```
behinddwalls: how is change proider going to be different? i am assuming we want to know which impl of change-provider to use for this queue i guess? can a queue support multiple change providers like github and phab as we have today for go-code etc

extension/queueconfig/queueconfig.go
```diff
@@ -0,0 +1,22 @@
...
+
+	"github.com/uber/submitqueue/entity/queueconfig"
+)
+
+// ErrNotFound is returned when the requested queue configuration does not exist.
+var ErrNotFound = errors.New("queue config not found")
+
+// Store loads and provides queue configurations.
+// Implementations may read from YAML files, databases, remote services, etc.
+type Store interface {
```
behinddwalls, line 15: should we add Create as well?

behinddwalls (changes requested):

extension/queueconfig/queueconfig.go
kevinlnew: Do we need one? My understanding is that users should have their queue configs explicitly and not be modifiable in runtime

entity/queueconfig/queueconfig_test.go
kevinlnew: removed

entity/queueconfig/queueconfig_test.go
kevinlnew: removed

entity/queueconfig/queueconfig.go
kevinlnew: moved into top-level entity package

entity/queueconfig/queueconfig.go
kevinlnew: I'm not sure about this, this seems like the responsibility of the change provider to know. Or should it be in the top level queue config?

entity/queueconfig/queueconfig.go
kevinlnew: I have the same question as above, does it make most sense to have queue config responsible for this, or change provider?

entity/queueconfig/queueconfig.go
kevinlnew: Yep, moved to entity top-level and changed to `QueueConfig`

entity/queueconfig/queueconfig.go
kevinlnew: I changed this to just use: 
```
ChangeProviders []string 
```
so we can have a list of something like
```
changeProviders:
    - github
    - phabricator
```

Might need new sub-configs to hook it up later but it should be able to support multiple if necessary

entity/queueconfig/queueconfig.go
kevinlnew: Changed `Destination` and `Repository` to just strings fields for now, can make them more complicated if necessary

entity/queue_config.go
```diff
@@ -0,0 +1,44 @@
...
+	// VCSType identifies the version control system (e.g., "git", "svn", "perforce").
+	// A queue operates on exactly one VCS.
+	VCSType string
+
+	// VCSRepo identifies the repository in the version control system.
+	// The format is VCS-specific:
+	//   - Git: remote URL (e.g., "git@github.com:uber/submitqueue.git")
+	//   - Perforce: depot path (e.g., "//depot/project")
+	//   - SVN: repository URL (e.g., "https://svn.example.com/repos/project")
+	VCSRepo string
```
behinddwalls: should we call it VCSURI?

kevinlnew: Named to `VCSAddress`

extension/queueconfig/queueconfig.go
kevinlnew: Added Create

extension/queueconfig/queueconfig.go
```diff
@@ -0,0 +1,29 @@
...
+
+// ErrAlreadyExists is returned when a queue configuration with the same name already exists.
+var ErrAlreadyExists = errors.New("queue config already exists")
+
+// Store loads and provides queue configurations.
+// Implementations may read from YAML files, databases, remote services, etc.
+type Store interface {
+	// Create adds a new queue configuration.
+	// Returns ErrAlreadyExists if a configuration with the same name already exists.
+	Create(ctx context.Context, config entity.QueueConfig) error
```
sbalabanov: should be external to the system

sbalabanov (approved):

behinddwalls (approved):

PR #52, opened 2026-02-24
(entities) Adds change provider entity and mysql backed store for it

manjari25:
## Summary
(entities) Adds change provider entity and mysql backed store for it

## Stack
1. @ #52
1. #48


manjari25 (commented):
These changes have already landed on main. Not sure why this PR is getting re-opened. https://github.com/uber/submitqueue/commit/3c25a2f018b8396ba943883bae9345232060fcda

PR #53, opened 2026-02-24
feat(entities) Adds batch entity and batch sql schema

manjari25:
## Summary
feat(entities) Adds batch entity and batch sql schema

## Why?

This sets up batch entity and its corresponding sql schema. This establishes the core concept of a batch that will be used through the system.

## What?
* Add a `Batch` entity
* Add `batch` mysql schema
* Note: `BatchStore` in upcoming PR


manjari25:
Same PR as https://github.com/uber/submitqueue/pull/46 since there is some weird behavior in the stack set up.

behinddwalls (approved):

PR #56, opened 2026-02-24
feat(storage) Adds batch store interface and mysql implementation

manjari25:
## Summary
feat(storage) Adds batch store interface and mysql implementation

## Why?

This sets up a batch store interface (and implementation) interact with the mysql backend for this table.

## What?
* Adds `BatchStore`

### DIFF
--- a/extension/storage/mysql/batch_store.go
+++ b/extension/storage/mysql/batch_store.go
@@ -1,0 +1,111 @@
+package mysql
+
+import (
+	"context"
+	"database/sql"
+	"encoding/json"
+	"errors"
+	"fmt"
+
+	"github.com/go-sql-driver/mysql"
+
+	"github.com/uber/submitqueue/entity"
+	"github.com/uber/submitqueue/extension/storage"
+)
+
+type batchStore struct {
+	db *sql.DB
+}
+
+// NewBatchStore creates a new MySQL-backed BatchStore.
+func NewBatchStore(db *sql.DB) storage.BatchStore {
+	return &batchStore{db: db}
+}
+
+// Get retrieves a batch by ID. Returns ErrNotFound if the batch is not found.
+func (s *batchStore) Get(ctx context.Context, id string) (entity.Batch, error) {
+	var batch entity.Batch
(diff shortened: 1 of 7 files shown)


gateway/controller/land_test.go
```diff
@@ -51,9 +51,37 @@ func (m *mockChangeProviderStore) Create(ctx context.Context, changeProvider ent
 	return m.createFunc(ctx, changeProvider)
 }
 
+type mockBatchStore struct {
```
behinddwalls, line 54: use gomocks for all the mock generation.

extension/storage/mysql/batch_store.go
```diff
@@ -0,0 +1,111 @@
...
+
+	rowsAffected, err := result.RowsAffected()
+	if err != nil {
+		return fmt.Errorf(
+			"failed to get rows affected from update for id=%q version=%d newState=%v: %w",
+			id, version, newState, err,
+		)
+	}
+
+	if rowsAffected != 1 {
```
behinddwalls, line 103: do we want to check for 0 instead and also do we want to check if more than 1 row is affected?

sbalabanov: +1

sbalabanov: VersionMismatch on 0 rows, general "unreachable" error on 2+ rows.
not blocking, good enough as is.

gateway/controller/land_test.go
sbalabanov: should we instruct Claude?

sbalabanov (approved):

gateway/controller/land_test.go
manjari25: I'll update this in a different PR so all the mocks in here are addressed

sbalabanov (approved):

extension/storage/mysql/batch_store.go
manjari25: Will update in a different PR.

PR #58, opened 2026-02-24
feat(storage) Adds build store interface and mysql implementation

manjari25:
## Summary
feat(storage) Adds build store interface and mysql implementation

## Why?

This sets up a build store interface (and implementation) interact with the mysql backend for this table.

## What?
* Adds `BuildStore`

### DIFF
--- a/extension/storage/build_store.go
+++ b/extension/storage/build_store.go
@@ -1,0 +1,20 @@
+package storage
+
+import (
+	"context"
+
+	"github.com/uber/submitqueue/entity"
+)
+
+// BuildStore is an interface that defines methods for managing builds in the database.
+type BuildStore interface {
+	// Get retrieves a build by ID. Returns ErrNotFound if the build is not found.
+	Get(ctx context.Context, id string) (entity.Build, error)
+
+	// Create creates a new build. The build must have a unique ID already assigned.
+	// Returns ErrAlreadyExists if a build with the same ID already exists.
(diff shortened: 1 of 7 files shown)


extension/storage/build_store.go
```diff
@@ -0,0 +1,20 @@
...
+	"context"
+
+	"github.com/uber/submitqueue/entity"
+)
+
+// BuildStore is an interface that defines methods for managing builds in the database.
+type BuildStore interface {
+	// Get retrieves a build by ID. Returns ErrNotFound if the build is not found.
+	Get(ctx context.Context, id string) (entity.Build, error)
+
```
behinddwalls, line 13: we may want to think about other access patterns here? do we need to read all builds for a queue? how do we prioritize builds? waiting vs running?

manjari25: > do we need to read all builds for a queue?

Maybe we have a  `GetByQueue(...)`. But, this can be easily added when the need arises.

> how do we prioritize builds? waiting vs running?

I don't think the DB needs to be aware of this priority. We should just give a list of builds with the state and the caller can determine what to do with them? We can at most have a `GetByState(...)` method here. Again - can be added as the need arises.

sbalabanov (approved):

extension/storage/build_store.go
manjari25: Resolving this conversation to unblock. Will add these methods in a different PR once we deem this access pattern is necessary.

PR #62, opened 2026-02-24
feat(storage) Adds speculation tree store interface and mysql implementation

manjari25:
## Summary
feat(storage) Adds speculation tree store interface and mysql implementation

## Why?

This sets up a speculation tree store interface (and implementation) interact with the mysql backend for this table.

## What?
* Adds `SpeculationTreeStore`


extension/storage/mysql/speculation_tree_store.go
```diff
@@ -0,0 +1,96 @@
...
+
+// UpdateSpeculations updates the speculations of a speculation tree. Returns ErrNotFound if the speculation tree is not found.
+func (s *speculationTreeStore) UpdateSpeculations(ctx context.Context, batchID string, speculations []entity.SpeculationInfo) error {
+	speculationsJSON, err := json.Marshal(speculations)
+	if err != nil {
+		return fmt.Errorf("failed to marshal speculations batchID=%s for UpdateSpeculations: %w", batchID, err)
+	}
+
+	result, err := s.db.ExecContext(ctx,
+		"UPDATE speculation_tree SET speculations = ? WHERE batch_id = ?",
```
sbalabanov, line 79: as long as we update it, should the entity(ies) be versioned?

sbalabanov (changes requested):

extension/storage/mysql/speculation_tree_store.go
manjari25: Yes - I think this needs to be fixed for a couple of other tables. I'll make a follow up PR for adding this field in other entities.

sbalabanov (approved):

behinddwalls (approved):

PR #65, opened 2026-02-24
feat(extension): add BuildManager extension for CI/CD integration

djuloori2794:
Add vendor-agnostic BuildManager interface for managing builds with external CI providers (BuildKite, Jenkins, etc.). Enables orchestrator to schedule builds, poll status, and cancel running builds.


entity/build/build_id.go
```diff
@@ -0,0 +1,68 @@
...
+// logging, and serialization.
+//
+// Examples:
+//   - "buildkite://uber/submitqueue-ci/123"
+//   - "jenkins://456"
+//   - "mock://1"
+//
+// For database storage, this single string format avoids the complexity of
+// composite keys and makes it easy to use as a primary key or index.
+type BuildID string
```
behinddwalls: we can just store string...

entity/build/build_state.go
```diff
@@ -0,0 +1,40 @@
...
+
+	// BuildStateFailed indicates the build completed with failures.
+	// This is a terminal state.
+	BuildStateFailed BuildState = "failed"
+
+	// BuildStateCancelled indicates the build was cancelled before completion.
+	// This is a terminal state.
+	BuildStateCancelled BuildState = "cancelled"
+
+	// BuildStateBlocked indicates the build is waiting for manual approval or unblocking.
```
behinddwalls: we can remove it...

entity/build/build_state.go
```diff
@@ -0,0 +1,40 @@
+package build
+
+// BuildState represents the execution state of a build.
+// This is a string enum following SubmitQueue's pattern of using string enums with sentinel values.
+type BuildState string
```
behinddwalls: https://github.com/uber/submitqueue/blob/main/entity/build.go#L8C2-L8C19 we should move it here

extension/build/build_manager.go
```diff
@@ -0,0 +1,67 @@
...
+		ctx context.Context,
+		baseSHA string,
+		speculatedBatchesToBeApplied []entity.BatchDependent,
+		batchToBeTest entity.Batch,
+		repoURL string,
+		branch string,
+		pipelineID string,
+		sqid string,
+		env map[string]string,
+		message string,
```
behinddwalls: Head - BatchID - the thing being tested
Base -  [] BatchID - order matters applied on main
JobName - pipeline/job name to be called

extension/build/build_manager.go
```diff
@@ -0,0 +1,67 @@
...
+	) (entitybuild.BuildID, error)
+
+	// Poll retrieves the current status of a build from the CI provider.
+	// This is a synchronous call that queries the provider's API.
+	//
+	// Returns:
+	//   - BuildStatus: Current state, timestamps, URLs, and metadata for the build
+	//   - error: ErrBuildNotFound if the build doesn't exist, ErrProviderUnavailable if the CI provider is unreachable
+	Poll(ctx context.Context, id entitybuild.BuildID) (entitybuild.BuildStatus, error)
```
behinddwalls:
```suggestion
	Poll(ctx context.Context, id string) (entity.BuildStatus, error)
```

extension/build/build_manager.go
```diff
@@ -0,0 +1,67 @@
...
+
+	// CancelBuild requests cancellation of a queued or running build.
+	// Builds that have already completed cannot be cancelled.
+	//
+	// Returns:
+	//   - error: ErrBuildNotFound if the build doesn't exist,
+	//            ErrBuildNotCancellable if the build has already finished,
+	//            ErrProviderUnavailable if the CI provider is unreachable
+	CancelBuild(ctx context.Context, id entitybuild.BuildID) error
```
behinddwalls: BuildID can be just string

entity/build/build_state.go
```diff
@@ -0,0 +1,40 @@
+package build
```
sbalabanov: In Go, it is not idiomatic to have each type in a separate file

entity/build/build_id.go
```diff
@@ -0,0 +1,68 @@
+package build
+
+import "fmt"
+
+// BuildID uniquely identifies a build across all CI providers.
```
sbalabanov: nit on wording: it does not have to be CI system, just something that can execute a validation (typically build+test) on a desired sha.

entity/build/build_id.go
sbalabanov: We probably want to discuss how we want to type this and similar. I am seeing typed approach and "raw string" approach used between different entities.

entity/build/build_id.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+// ParseBuildID extracts the provider and ID components from a BuildID string.
+// Returns the provider name, provider-specific ID, and an error if the format is invalid.
+//
+// Expected format: "provider://id"
+func ParseBuildID(buildID BuildID) (provider string, id string, err error) {
+	s := string(buildID)
+
+	// Find the separator
```
sbalabanov: can use `strings.index()`

entity/build/build_id_test.go
```diff
@@ -0,0 +1,132 @@
...
+		},
+		{
+			name:             "valid jenkins ID",
+			buildID:          "jenkins://456",
+			expectedProvider: "jenkins",
+			expectedID:       "456",
+			expectError:      false,
+		},
+		{
+			name:             "valid mock ID",
```
sbalabanov: from a functional standpoint, "valid jenkins ID" and "valid mock ID" are the same tests.
AI is clearly overdid on this one.

entity/build/build_state.go
```diff
@@ -0,0 +1,40 @@
...
+// This is a string enum following SubmitQueue's pattern of using string enums with sentinel values.
+type BuildState string
+
+const (
+	// BuildStateUnknown is the sentinel value for unknown or unreachable build states.
+	// This should only occur if the provider returns an unexpected state value.
+	BuildStateUnknown BuildState = ""
+
+	// BuildStateQueued indicates the build has been scheduled but not yet started.
+	BuildStateQueued BuildState = "queued"
```
sbalabanov: nit: may be "accepted". We do not really know if a Build System implementation queues or not, what we know is that it accepted it and gave back build id.

entity/build/build_state.go
```diff
@@ -0,0 +1,40 @@
...
+	// This is a terminal state.
+	BuildStatePassed BuildState = "passed"
+
+	// BuildStateFailed indicates the build completed with failures.
+	// This is a terminal state.
+	BuildStateFailed BuildState = "failed"
+
+	// BuildStateCancelled indicates the build was cancelled before completion.
+	// This is a terminal state.
+	BuildStateCancelled BuildState = "cancelled"
```
sbalabanov: I think there are technically 2 cancelled states: one indicates that the build is canceled by submitqueue, another one that the build was canceled externally (i.e. failed to finish, but the outcome is not clear).
Most likely we'll need to distinguish these two.

sbalabanov: + most likely we want "cancelling" state, too, indicating that submitqueue asked to cancel. At this moment the further destiny of the build is not something Submitqueue is to care about.
May be this "cancelling" should be a terminal state and we do not have to update the actual state returned by the build system after this.

entity/build/build_state.go
sbalabanov: the question is, which states we want to propagate from build system to here.
We probably want to design SQ states to those only relevant to SubmitQueue. which are: "accepted", "failed", "passed", "cancelled".
Everything else is not required by logic and serves purely for user's entertainment (i.e. UI).
Should this just be a build metadata then, updated by Build System but not carrying transitions?

sbalabanov: thoughts?

entity/build/build_state_test.go
```diff
@@ -0,0 +1,159 @@
+package build
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// TestBuildState_IsTerminal tests the IsTerminal method for all build states.
+func TestBuildState_IsTerminal(t *testing.T) {
```
sbalabanov: change detector test. likely not needed at all.

entity/build/build_state_test.go
```diff
@@ -0,0 +1,159 @@
...
+			name:     "blocked",
+			state:    BuildStateBlocked,
+			expected: "blocked",
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			assert.Equal(t, tc.expected, string(tc.state),
+				"BuildState constant should have expected string value")
```
sbalabanov: AI..... please stop....

entity/build/build_state_test.go
```diff
@@ -0,0 +1,159 @@
...
+// TestBuildState_ZeroValue verifies that the zero value is BuildStateUnknown.
+func TestBuildState_ZeroValue(t *testing.T) {
+	var state BuildState
+	assert.Equal(t, BuildStateUnknown, state, "Zero value should be BuildStateUnknown")
+	assert.Equal(t, "", string(state), "Zero value should be empty string")
+	assert.False(t, state.IsTerminal(), "Zero value should not be terminal")
+}
+
+// TestBuildState_TerminalStates verifies all terminal states are accounted for.
+func TestBuildState_TerminalStates(t *testing.T) {
```
sbalabanov: this change detector test repeats another change detector test `TestBuildState_IsTerminal`

entity/build/build_status.go
```diff
@@ -0,0 +1,40 @@
...
+	// Zero if the build has not been queued yet.
+	QueuedAt int64
+
+	// StartedAt is the Unix timestamp in milliseconds when the build started executing.
+	// Zero if the build has not started yet.
+	StartedAt int64
+
+	// FinishedAt is the Unix timestamp in milliseconds when the build finished (passed, failed, or cancelled).
+	// Zero if the build has not finished yet.
+	FinishedAt int64
```
sbalabanov: why do we need all these fields in SQ?

extension/build/build_manager.go
```diff
@@ -0,0 +1,62 @@
+package build
```
sbalabanov, line 1: seems my comments on prev revision do not reflect inline on this revision, but please check them as some of them are still relevant.

extension/build/build_manager.go
```diff
@@ -0,0 +1,62 @@
...
+	"context"
+
+	"github.com/uber/submitqueue/entity"
+)
+
+// BuildManager is a vendor-agnostic interface for managing builds with external CI/CD providers.
+// Implementations provide integration with specific CI systems (BuildKite, Jenkins, etc.)
+// to schedule builds, poll their status, and cancel running builds.
+//
+// All implementations must be thread-safe and support concurrent operations.
```
sbalabanov, line 19: We probably also want to dictate if the implementation can do a heavy singleton, lightweight initialize-on-demand, or either.

extension/build/build_manager.go
```diff
@@ -0,0 +1,62 @@
...
+	//   - base: List of BatchIDs (in order) that have been applied on main. Order matters.
+	//   - jobName: Pipeline/job name to be called on the CI provider
+	//
+	// Returns:
+	//   - string: Unique build ID
+	//   - error: ErrInvalidRequest if validation fails, ErrProviderUnavailable if CI provider is unreachable
+	ScheduleBuild(
+		ctx context.Context,
+		head string,
+		base []string,
```
sbalabanov: ScheduleBuild implementation should probably not rely on its knowledge of the Batches.
I think we need to translate Batches to real buildable arguments before we call this function.

extension/build/build_manager.go
```diff
@@ -0,0 +1,62 @@
...
+	// Parameters:
+	//   - id: Build ID string
+	//
+	// Returns:
+	//   - BuildStatus: Current state of the build
+	//   - error: ErrBuildNotFound if the build doesn't exist, ErrProviderUnavailable if CI provider is unreachable
+	Poll(ctx context.Context, id string) (entity.BuildStatus, error)
+
+	// CancelBuild requests cancellation of a queued or running build.
+	// Builds that have already completed cannot be cancelled.
```
sbalabanov: SQ can still request it. It's up for the implementation how it wants to process it at this point.

extension/build/build_manager.go
```diff
@@ -0,0 +1,62 @@
...
+	//
+	// Parameters:
+	//   - id: Build ID string
+	//
+	// Returns:
+	//   - BuildStatus: Current state of the build
+	//   - error: ErrBuildNotFound if the build doesn't exist, ErrProviderUnavailable if CI provider is unreachable
+	Poll(ctx context.Context, id string) (entity.BuildStatus, error)
+
+	// CancelBuild requests cancellation of a queued or running build.
```
sbalabanov: Worth saying that the implementation is asynchronous and does not need to wait for the cancellation to complete.

extension/build/build_manager.go
```diff
@@ -0,0 +1,62 @@
...
+	// Builds that have already completed cannot be cancelled.
+	//
+	// Parameters:
+	//   - id: Build ID string
+	//
+	// Returns:
+	//   - error: ErrBuildNotFound if the build doesn't exist,
+	//            ErrBuildNotCancellable if the build has already finished,
+	//            ErrProviderUnavailable if the CI provider is unreachable
```
sbalabanov: Do we need a special error for this, if yes, why?
The provider may be reachable but error out this request for another reason.

sbalabanov (changes requested):

extension/build/build_manager.go
```diff
@@ -0,0 +1,62 @@
...
+	//   - jobName: Pipeline/job name to be called on the CI provider
+	//
+	// Returns:
+	//   - string: Unique build ID
+	//   - error: ErrInvalidRequest if validation fails, ErrProviderUnavailable if CI provider is unreachable
+	ScheduleBuild(
+		ctx context.Context,
+		head string,
+		base []string,
+		jobName string,
```
behinddwalls: seems like this is in queconfig now? What about just passing the QueName?

extension/build/build_manager.go
behinddwalls: would it make sense to have list of `Change`(s) to be applied on main and list of `Change`(s) for head?

```
ScheduleBuild(ctx context.Context, queue string, base []Change, head []Change)
```

djuloori2794: I was having a same impression, but after discussing with @behinddwalls we thought all of these could go into the implementation details.

extension/build/build_manager.go
djuloori2794: instead of jobName?

extension/build/build_manager.go
djuloori2794: Makes sense!

extension/build/build_manager.go
djuloori2794: The intent here is to treat provider availability failures differently from request failures.

entity/build.go
```diff
@@ -61,3 +59,26 @@ type Build struct {
 	// Status represents the state of the build lifecycle this build is in.
 	Status BuildStatus
 }
+
+// BuildAction defines the action to perform on a change submitted to the build system.
+type BuildAction string
```
behinddwalls: not sure if i want to call it BuildAction, but it's more of ChangeAction for checkout?

entity/build.go
```diff
@@ -10,6 +10,9 @@ const (
 	// BuildStatusQueued indicates the build has been scheduled but not yet started.
 	BuildStatusQueued BuildStatus = "queued"
```
behinddwalls: we can remove it

entity/build.go
behinddwalls: rename to `succeeded`

entity/build.go
```diff
@@ -61,3 +59,26 @@ type Build struct {
...
 	Status BuildStatus
 }
+
+// BuildAction defines the action to perform on a change submitted to the build system.
+type BuildAction string
+
+const (
+	// BuildActionUnknown is the sentinel value for uninitialized actions.
+	BuildActionUnknown BuildAction = ""
+	// BuildActionValidate runs validation/testing on the change without applying it.
```
sbalabanov: actually, it will have to apply it first

entity/build.go
sbalabanov: we need 4 states

extension/build/build_manager.go
```diff
@@ -0,0 +1,75 @@
...
+	// ignore the request, or handle it in a provider-specific way.
+	//
+	// Parameters:
+	//   - id: Build ID string
+	//
+	// Returns:
+	//   - error: ErrBuildNotFound if the build doesn't exist,
+	//            ErrBuildNotCancellable if the build cannot be cancelled (implementation-specific),
+	//            ErrProviderUnavailable if the CI provider is unreachable
+	CancelBuild(ctx context.Context, id string) error
```
behinddwalls:
```suggestion
	CancelBuild(ctx context.Context, buildID string) error
```

sbalabanov (changes requested):

extension/build/build_manager.go
```diff
@@ -0,0 +1,75 @@
...
+	// This is a synchronous call that queries the provider's API.
+	//
+	// Parameters:
+	//   - buildID: Build ID string
+	//
+	// Returns:
+	//   - BuildStatus: Current state of the build
+	//   - map[string]string: Additional metadata about the build (e.g., build URL, commit SHA, duration)
+	//   - error: ErrBuildNotFound if the build doesn't exist
+	Poll(ctx context.Context, buildID string) (entity.BuildStatus, map[string]string, error)
```
behinddwalls: for map, we can just say `BuildMetadata` as an entity and just keep it empty for now

behinddwalls (approved):

entity/build.go
```diff
@@ -24,17 +21,12 @@ const (
 	// BuildStatusCancelled indicates the build was cancelled before completion.
```
sbalabanov: Important to mention that this is SQ that canceled the build. Any Build system originated cancellation is an error in SQ's world.

entity/build.go
```diff
@@ -61,3 +53,30 @@ type Build struct {
...
+
+// ChangeAction defines the action to perform on a change submitted to the build system.
+type ChangeAction string
+
+const (
+	// ChangeActionUnknown is the sentinel value for uninitialized actions.
+	ChangeActionUnknown ChangeAction = ""
+	// ChangeActionApply applies the change to the target branch.
+	ChangeActionApply ChangeAction = "apply"
+	// ChangeActionValidate runs validation/testing on the change without applying it.
```
sbalabanov: It still has to apply the change first, then build targets changed in this change vs previous change in a chain

entity/build.go
```diff
@@ -61,3 +53,30 @@ type Build struct {
...
+	ChangeActionValidate ChangeAction = "validate"
+)
+
+// BuildChange represents a code change to be processed by the build system.
+// This is used by BuildManager to specify what changes to build and what action to perform.
+type BuildChange struct {
+	// ChangeID is the unique identifier for this change.
+	// This is typically a diff ID (e.g., "D12345") or PR number (e.g., "42"),
+	// depending on the source control provider.
+	ChangeID string
```
sbalabanov: This has to be either Change structure (already defined in entity model) or list or change URLs

extension/build/README.md
```diff
@@ -0,0 +1,345 @@
...
+    CancelBuild(ctx context.Context, buildID string) error
+
+    // Close gracefully shuts down the build manager
+    Close() error
+}
+```
+
+**Thread-safety**: All implementations must be thread-safe and support concurrent operations.
+
+**Implementation Design**: Implementations may be designed as heavy singletons (with connection pooling and caching) or lightweight instances created on-demand, depending on the CI provider's requirements and implementation strategy.
```
sbalabanov: I assume we initialize build manager only once (at least one per each build provider), which means that the implementation needs to be long living. Similar to other components like Storage.

extension/build/README.md
```diff
@@ -0,0 +1,345 @@
...
+        Schedule(gomock.Any(), "test-queue", changes).
+        Return("mock://1", nil)
+
+    // Test your code that uses the mock
+}
+```
+
+### Cancelling Builds
+
+Cancellation is **asynchronous** - the method initiates the cancellation request and returns immediately. Use `Poll` to check if the build has transitioned to `BuildStatusCancelled`.
```
sbalabanov: I think BuildStatusCancelled will be set without confirmation from the build system.

extension/build/README.md
```diff
@@ -0,0 +1,345 @@
...
+
+Potential improvements not in the current implementation:
+
+- **Webhook support**: Accept push notifications from CI providers instead of polling
+- **Build artifacts**: Track and retrieve build artifacts
+- **Retry logic**: Automatic retry for transient failures
+- **Batch operations**: Poll/cancel multiple builds at once
+- **Streaming logs**: Real-time log streaming from builds
+- **Build caching**: Cache build status to reduce API calls
+- **Partial cancellation**: Cancel individual changes in a multi-change build
```
sbalabanov: why is this?

extension/build/build_manager.go
```diff
@@ -0,0 +1,75 @@
...
+	// The implementation decides how to handle cancellation requests for builds in
+	// terminal states (succeeded, failed, cancelled). It may return an error, silently
+	// ignore the request, or handle it in a provider-specific way.
+	//
+	// Parameters:
+	//   - buildID: Build ID string
+	//
+	// Returns:
+	//   - error: ErrBuildNotFound if the build doesn't exist,
+	//            ErrBuildNotCancellable if the build cannot be cancelled (implementation-specific)
```
sbalabanov: I do not think it needs to return ErrBuildNotCancellable
From SW perspective, it should be an idempotent operation. The build needs to be stopped only to save capacity, its results are not relevant anymore, if it managed to complete or fail it should not be an error.
An error is when an infra failure occurs within the Build ecosystem processing this request. Even in this case SQ will likely only need to see this error for observability and can still proceed with the workflow.

extension/build/build_manager.go
```diff
@@ -0,0 +1,75 @@
...
+	CancelBuild(ctx context.Context, buildID string) error
+
+	// Close gracefully shuts down the build manager.
+	// Implementations should cancel pending requests, close HTTP clients, and clean up resources.
```
sbalabanov: should they cancel with error?

sbalabanov (changes requested):

extension/build/README.md
```diff
@@ -0,0 +1,347 @@
...
+Build IDs returned by `Schedule` should use a URI-like format: `"provider://id"`
+
+**Examples**:
+- `"buildkite://uber/submitqueue-ci/123"`
+- `"jenkins://456"`
+- `"mock://1"`
+
+This format allows the implementation to encode both the provider name and provider-specific build identifier in a single string.
+
+### Build Status Enum
```
sbalabanov, line 91: avoid exact domain definitions in docs as they run out of sync. Prefer samples / examples, only when needed.

entity/build.go
```diff
@@ -61,3 +55,30 @@ type Build struct {
...
+	ChangeActionApply ChangeAction = "apply"
+	// ChangeActionValidate applies the change first, and then validates the change by running respective validation/test suites.
+	ChangeActionValidate ChangeAction = "validate"
+)
+
+// BuildChange represents a code change to be processed by the build system.
+// This is used by BuildManager to specify what changes to build and what action to perform.
+type BuildChange struct {
+	// Change is the code change to process.
+	// This references the same Change entity used in Request, containing the source provider
```
sbalabanov, line 75: at this moment Change is just a list of URLs and provider is encoded into schema part of it

entity/build_test.go
```diff
@@ -28,30 +28,64 @@ func TestBuildStatus_IsTerminal(t *testing.T) {
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			assert.Equal(t, tt.expected, tt.status.IsTerminal())
+		})
+	}
+}
+
+func TestBuildChange_Creation(t *testing.T) {
```
sbalabanov, line 49: change detector

extension/build/mock/build_manager_test.go
```diff
@@ -0,0 +1,60 @@
...
+		{ChangeID: "D12345", Action: entity.ChangeActionApply},
+		{ChangeID: "D12346", Action: entity.ChangeActionValidate},
+	}
+
+	buildID, err := mockBuildMgr.Schedule(
+		context.Background(), queueName, changes,
+	)
+
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
```
sbalabanov, line 35: assert/require

extension/build/mock/build_manager_test.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"context"
+	"testing"
+
+	"go.uber.org/mock/gomock"
+	"github.com/uber/submitqueue/entity"
+)
+
+// TestMockBuildManager_Compilation verifies the mock compiles and basic setup works.
+func TestMockBuildManager_Compilation(t *testing.T) {
```
sbalabanov, line 12: I do not think we need a test for a mock :)

extension/build/mock/build_manager_test.go
```diff
@@ -0,0 +1,60 @@
+package mock
```
sbalabanov, line 1: use `gomock` bazel rule, refer to docs on how

extension/build/README.md
```diff
@@ -0,0 +1,347 @@
+# BuildManager Extension
```
sbalabanov, line 1: plz consider significantly reducing this doc, only leave what's applicable to Build Manager specifically.

extension/build/README.md
```diff
@@ -0,0 +1,347 @@
...
+5. **Long-lived singletons**: Implementations are initialized once per build provider at service startup
+6. **Asynchronous cancellation**: CancelBuild initiates request and returns immediately
+
+## Future Enhancements
+
+Potential improvements not in the current implementation:
+
+- **Webhook support**: Accept push notifications from CI providers instead of polling
+- **Build artifacts**: Track and retrieve build artifacts
+- **Retry logic**: Automatic retry for transient failures
```
sbalabanov, line 344: I think we'll handle retries on the Queue/Controller

extension/build/README.md
```diff
@@ -0,0 +1,347 @@
...
+- **Batch operations**: Poll/cancel multiple builds at once
```
sbalabanov, line 345: why need?

extension/build/README.md
```diff
@@ -0,0 +1,347 @@
...
+- **Streaming logs**: Real-time log streaming from builds
```
sbalabanov, line 346: also not sure if really needed.
a URL for the build may be good enough?
I suggest to avoid documenting speculations, which AI very likes to do.

sbalabanov (changes requested):

extension/build/README.md
djuloori2794: Hmm, definitely not a critical requirement. I can remove them!

extension/build/README.md
djuloori2794: I can remove it, but if we have to cancel multiple builds at once. Can't we do it in a single call if the provider supports it?

entity/build.go
djuloori2794: isn't line number 76 good enough?

djuloori2794:
Will just redo the PR!

PR #68, opened 2026-02-25
feat(queueconfig): Add a BuildRunner to the queueconfig

kevinlnew:
## Summary
feat(queueconfig): Add a BuildRunner to the queueconfig

What?
Adding a string field for BuildRunner which should identify the build that the queue will run, i.e. a Buildkite pipeline URL. 
also removed the constructor for the entity, and just following the pattern to directly build it.

- also, adding both json and yaml tags to the queueconfig struct using snake_case.

## Test Plan
unit test, our ci suite.


entity/queue_config.go
```diff
@@ -7,38 +7,30 @@ package entity
...
-		Name:       name,
-		VCSType:    vcsType,
-		VCSAddress: vcsAddress,
-		Target:     target,
-	}
+	// BuildRunner identifies the CI pipeline or job that runs builds for this queue.
+	// Opaque to the system; meaningful only to the build extension implementation.
+	// Examples:
+	//   - Buildkite: "buildkite.com/uber/submitqueue-ci"
+	//   - Jenkins: "jenkins.example.com/job/submitqueue-verify"
```
sbalabanov, line 34: will probably need to update this docstring later on what the expected values could be. Are they keys in another configuration? Are they hardcoded?

sbalabanov (approved):

entity/queue_config.go
```diff
@@ -7,38 +7,30 @@ package entity
...
+	BuildRunner string `json:"build_runner" yaml:"build_runner"`
```
behinddwalls, line 35: should we have some ore attributes, may be make it a struct? how many requests we can trigger concurrently? should we have here or just leave it to the implementation specific config?

behinddwalls: Also this feels very much like change-provider? BuildRunnerProvider and BuildRunnerURL?

maybe we do a single URL `{provider-name}://{URI}`


Also, I am wondering how are these passed? say API accesspts a Github PR, BuildManager must understand the Change Provider/URI semantics and based on BuildRunner provider? we select a runner from implementations automatically?

I am trying to understand end to end flow which is not clear to me...maybe we draw it out tomorrow end to end..first things and I would like to just get everything out in one go from the config pov.

behinddwalls (approved):

PR #69, opened 2026-02-25
feat(mergechecker): adding mergechecker interface with Check

kevinlnew:
## Summary
feat(mergechecker): adding mergechecker interface with Check
What?
- basic interface with `Check`: explains how this is not a mergeability guarantee but a heuristic to advance the request or fail it fast.


extension/mergechecker/mergechecker.go
```diff
@@ -0,0 +1,15 @@
...
+
+	"github.com/uber/submitqueue/entity"
+)
+
+// MergeChecker predicts whether a request's changes can merge cleanly.
+type MergeChecker interface {
+	// Check is a fail-fast validation that optimistically assesses the
+	// mergeability of the request. A positive result does not guarantee
+	// that the changes will apply cleanly at merge finalization time.
+	Check(ctx context.Context, request entity.Request) (bool, error)
```
behinddwalls: we can get rid of the bool and have named Error something like ErrMergeConflict

extension/mergechecker/mergechecker.go
```diff
@@ -0,0 +1,20 @@
...
+// ErrUnmergeable is returned by Check when the request's changes are not mergeable.
+var ErrUnmergeable = errors.New("request is not mergeable")
+
+// MergeChecker predicts whether a request's changes can merge cleanly.
+type MergeChecker interface {
+	// Check is a fail-fast validation that optimistically assesses the
+	// mergeability of the request. A nil result does not guarantee
+	// that the changes will apply cleanly at merge finalization time.
+	// Returns ErrUnmergeable if the changes are not mergeable.
+	Check(ctx context.Context, request entity.Request) error
```
sbalabanov: let's have it return true/false + error
using errors to control code flow should in general be avoided (even though Go sdk is guilty of it sometimes).

sbalabanov (changes requested):

sbalabanov:
I see now that I am contradicting earlier Preetam's comment.
I am pretty opinionated about this one though. The primary result of such a function is a boolean value, and so it should be modeled as such. Merge conflict is not an error in the system, but an expected state.

kevinlnew:
Sounds good, bring the bool back

extension/mergechecker/mergechecker.go
sbalabanov: Most likely in the future we'll have to enhance the result with some metadata anyways. For example, return back to user with the additional info what files are conflicting against what hash/request. The return object is probably justified to be a structure of some sort.

sbalabanov (changes requested):

extension/mergechecker/mergechecker.go
behinddwalls: @kevinlnew can you leave an edit for CLAUDE.md to check for future changes on this

extension/mergechecker/mergechecker.go
```diff
@@ -0,0 +1,20 @@
...
+
+import (
+	"context"
+	"errors"
+
+	"github.com/uber/submitqueue/entity"
+)
+
+// ErrUnmergeable is returned by Check when the request's changes are not mergeable.
+var ErrUnmergeable = errors.New("request is not mergeable")
```
sbalabanov: I think this error is not needed now

extension/mergechecker/mergechecker.go
```diff
@@ -0,0 +1,20 @@
...
+
+// MergeChecker predicts whether a request's changes can merge cleanly.
+type MergeChecker interface {
+	// Check is a fail-fast validation that optimistically assesses the
+	// mergeability of the request. A positive result does not guarantee
+	// that the changes will apply cleanly at merge finalization time.
+	// Returns ErrUnmergeable if the changes are not mergeable.
```
sbalabanov: also not relevant, error should indicate that the merge checkimplementation can't do the check at all.

sbalabanov (changes requested):

extension/mergechecker/mergechecker.go
kevinlnew: Keeping the bool as per Sergey's comments

extension/mergechecker/mergechecker.go
kevinlnew: Yep, removed

extension/mergechecker/mergechecker.go
kevinlnew: I'm fine to leave this to the implementation to return a relevant error message. Do we need typed err here?

extension/mergechecker/mergechecker.go
kevinlnew: Yep I'll do that in this diff

behinddwalls (approved):

PR #70, opened 2026-02-25
feat(entities) Update batch state enum with more states

manjari25:
## Summary
feat(entities) Update batch state enum with more states

## What?
Added batch state constants (scheduled, speculating, finalizing, succeeded, failed, cancelled,
cancellationfailed) and an IsTerminal() method to BatchState with table-driven tests.

## Why?
This state information is needed to inspect batches that should be under consideration for dependency graph calculation by the batch manager.

## Test Plan
Unit tests added.

### DIFF
--- a/entity/batch.go
+++ b/entity/batch.go
@@ -6,15 +6,15 @@ type BatchState string
 const (
 	// BatchStateUnknown is the unreachable state. It is set by default when the structure is initialized. It should never be seen in the system.
 	BatchStateUnknown BatchState = ""
-	// BatchStateScheduled is the state of a batch that has been scheduled for processing.
-	BatchStateScheduled BatchState = "scheduled"
+	// BatchStateCreated is the state of a batch that has been created for processing.
+	BatchStateCreated BatchState = "created"
 	// BatchStateSpeculating is the state of a batch that is undergoing speculative execution.
 	BatchStateSpeculating BatchState = "speculating"
 	// BatchStateFinalizing is the state of a batch that is being finalized after speculative execution.
 	BatchStateFinalizing BatchState = "finalizing"
 	// BatchStateSucceeded is the terminal state of a batch that has been successfully landed.
 	BatchStateSucceeded BatchState = "succeeded"
-	// BatchStateFailed is the terminal state of a batch that has failed to land.
+	// BatchStateFailed is the terminal state of a batch that has failed.
 	BatchStateFailed BatchState = "failed"
 	// BatchStateCancelled is the terminal state of a batch that was cancelled before completion.
 	BatchStateCancelled BatchState = "cancelled"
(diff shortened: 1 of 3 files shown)


entity/batch.go
```diff
@@ -6,9 +6,33 @@ type BatchState string
...
+	// BatchStateFinalizing is the state of a batch that is being finalized after speculative execution.
+	BatchStateFinalizing BatchState = "finalizing"
+	// BatchStateSucceeded is the terminal state of a batch that has been successfully landed.
+	BatchStateSucceeded BatchState = "succeeded"
+	// BatchStateFailed is the terminal state of a batch that has failed to land.
+	BatchStateFailed BatchState = "failed"
+	// BatchStateCancelled is the terminal state of a batch that was cancelled before completion.
+	BatchStateCancelled BatchState = "cancelled"
+	// BatchStateCancellationFailed is the terminal state of a batch whose cancellation process itself failed.
+	BatchStateCancellationFailed BatchState = "cancellationfailed"
```
behinddwalls: we can remove this as it's user error message

manjari25: Updated

entity/batch.go
```diff
@@ -6,9 +6,31 @@ type BatchState string
 const (
 	// BatchStateUnknown is the unreachable state. It is set by default when the structure is initialized. It should never be seen in the system.
 	BatchStateUnknown BatchState = ""
-	// TODO: Add comprehensive list of known batch states.
+	// BatchStateScheduled is the state of a batch that has been scheduled for processing.
+	BatchStateScheduled BatchState = "scheduled"
```
sbalabanov: the first state should probably be "new" or "created"

entity/batch.go
```diff
@@ -6,9 +6,31 @@ type BatchState string
...
+	// BatchStateSpeculating is the state of a batch that is undergoing speculative execution.
+	BatchStateSpeculating BatchState = "speculating"
+	// BatchStateFinalizing is the state of a batch that is being finalized after speculative execution.
+	BatchStateFinalizing BatchState = "finalizing"
+	// BatchStateSucceeded is the terminal state of a batch that has been successfully landed.
+	BatchStateSucceeded BatchState = "succeeded"
+	// BatchStateFailed is the terminal state of a batch that has failed to land.
```
sbalabanov: just failed to land or failed in general?

entity/batch_test.go
```diff
@@ -0,0 +1,30 @@
+package entity
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestBatchState_IsTerminal(t *testing.T) {
```
sbalabanov, line 9: change detector test. does not seem to be helpful.

sbalabanov (approved):

entity/batch.go
manjari25: Failed in general. Updated.

entity/batch.go
manjari25: Updated.

entity/batch_test.go
manjari25: Will make a new PR - I think there is one more in the repo that we should clean up.

PR #79, opened 2026-02-25
test(storage): Replace handwritten storage mocks with gomock bazel rule

manjari25:
## Summary
test(storage): Replace handwritten storage mocks with gomock bazel rule

## What?
Replace handwritten storage mocks in gateway/controller/land_test.go with generated mocks using gomock and the Bazel gomock rule. The mocks are generated at build time — no generated .go files are checked in.

## Why?
Handwritten mocks have to be manually updated every time a storage interface changes. Using gomock with the Bazel gomock rule eliminates this maintenance burden and guarantees mocks stay in sync with interfaces automatically.


extension/storage/mock/BUILD.bazel
```diff
@@ -0,0 +1,88 @@
+load("@rules_go//extras:gomock.bzl", "gomock")
```
behinddwalls, line 1: can we do this for all packages? Also, we should add instructions to claude as well

manjari25: Added claude instructions. Will change other packages in next PR.

sbalabanov (approved):

Issue #80, opened 2026-02-25
Flesh out Batch controller


manjari25:
* Create batches
* Create dependencies
* Update stores
* Publish to batched queue

PR #82, opened 2026-02-25
feat(controller) Add batch ID generation via counter to batch controller

manjari25:
## Summary
feat(controller) Add batch ID generation via counter to batch controller

## What?
The batch controller generates a unique batch ID using the counter.Counter extension (backed by MySQL). Each batch gets an ID in the format {queue}/batch/{seq} and a Batch entity is constructed from the request. Note that for now each batch maps to a single request. This will evolve in the future to have multiple requests map to a single batch with some conflict graph processing.

## Why?
This wires in the counter extension to produce monotonically increasing, globally unique batch IDs per queue — a prerequisite for persisting batches and tracking their dependencies in future work.

## Issues
https://github.com/uber/submitqueue/issues/80


orchestrator/controller/batch/batch_test.go
```diff
@@ -16,8 +17,27 @@ import (
...
+type mockCounter struct {
+	nextFunc func(ctx context.Context, domain string) (int64, error)
+}
+
+func (m *mockCounter) Next(ctx context.Context, domain string) (int64, error) {
+	return m.nextFunc(ctx, domain)
+}
+
+// newSequentialCounter returns a mock counter that returns incrementing values starting at 1.
+func newSequentialCounter() *mockCounter {
```
sbalabanov, line 30: probably should have it under the counter folder itself

sbalabanov (approved):

orchestrator/controller/batch/batch_test.go
```diff
@@ -16,8 +17,27 @@ import (
 	"go.uber.org/zap/zaptest"
 )
 
+// mockCounter implements counter.Counter for testing.
+type mockCounter struct {
```
behinddwalls, line 21: we should should gomock for mocking counter interface

manjari25: Yes - I'm making a PR for this

orchestrator/controller/batch/batch_test.go
manjari25: I'll refactor the counter mock in a different PR

PR #88, opened 2026-02-26
feat(controller) Query store and populate dependencies field in batch

manjari25:
## Summary
feat(controller) Query store and populate dependencies field in batch

##What?
Added storage.Storage to the batch controller and used GetByQueueAndStates to fetch active batches for the request's queue, setting them as dependencies on the newly created batch.

## Why?
A new batch needs to know which existing batches in its queue are still in progress (created, speculating, finalizing) so it can track its position in the dependency chain.


orchestrator/controller/batch/batch.go
```diff
@@ -93,22 +97,39 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) er
...
 		"batch_id", batch.ID,
 		"request_id", request.ID,
 		"queue", request.Queue,
 	)
 
-	// TODO:
-	// - Add batch to DB
-	// - Create batch dependent entity
-	// - Add to batch dependent DB
-
```
manjari25, line 111: Should not have deleted this. These three TODOs are still relevant.

behinddwalls (approved):

PR #90, opened 2026-02-26
feat(controller) Create batch dependents entity based on the batch created

manjari25:
## Summary
feat(controller) Create batch dependents entity based on the batch created

## What?
The batch controller now creates BatchDependent entities (the reverse of Batch.Dependencies) when a new batch is formed. For each active batch the new batch depends on, it looks up existing dependents from the store and merges them with the new batch ID.

## Why? 
Downstream pipeline stages need to know which batches are waiting on a given batch to complete. Batch.Dependencies answers "what do I depend on?" but not "who depends on me?" The BatchDependent reverse index enables efficient lookups for cascading state changes — e.g., when a batch succeeds or fails, the system can quickly find and notify all dependent batches.

## Issues
#80

### DIFF
--- a/orchestrator/controller/batch/batch.go
+++ b/orchestrator/controller/batch/batch.go
@@ -124,15 +124,38 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) er
 		})
 	}
 
+	// Create batch dependent entities (reverse relationship of batch.Dependencies).
+	// For each dependency, record the new batch as a dependent.
+	// If existing dependents are found in the store, append them.
+	for _, dep := range activeBatches {
+		bd := entity.BatchDependent{
+			BatchID:    dep.ID,
+			Dependents: []string{batch.ID},
+		}
+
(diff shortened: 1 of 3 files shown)


orchestrator/controller/batch/batch.go
```diff
@@ -124,15 +124,38 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) er
 		})
 	}
 
+	// Create batch dependent entities (reverse relationship of batch.Dependencies).
```
sbalabanov, line 127: Why does this controller need them?

sbalabanov (approved):

orchestrator/controller/batch/batch.go
manjari25: We don't need them in this controller. Since the arrival of a new batch; the batch dependents table will change; that's what this is tracking.

manjari25: I have a TODO below - this will be written to the DB so that speculation manager can track it.

PR #91, opened 2026-02-26
feat(errors): Architecture of errors

sbalabanov:
## Summary
General purpose package for error classification + refactor.
Gateway request input validation changed to User error. Will output proper GRPC error code.
Refer to core/errs/README.md for detailed design.


core/errs/errs.go
```diff
@@ -0,0 +1,131 @@
...
+// A user error is an error that is caused by the user's action or input, for example an invalid input or a merge conflict.
+func NewUserError(cause error) error {
+	return &userError{cause: cause, retryable: false}
+}
+
+// NewRetryableUserError creates a retryable user error wrapping the given cause. It is very rare for the user
+// error to be retryable and often a result of misclassification. Please know what you are doing when you use this.
+func NewRetryableUserError(cause error) error {
+	return &userError{cause: cause, retryable: true}
+}
```
kevinlnew, line 26: I understand having retryable user error for completion, but personally I can't think of a case where this would happen. Maybe consider removing it? Or provide an example

kevinlnew (approved):
LGTM. Left a nit comment for RetryableUserError* but ultimately up to you

core/errs/errs.go
behinddwalls: I think the classic example is 409 status codes i believe o.e. optimistic locking cases, and even though they belong 4xx, they are still retryable errors

core/errs/README.md
```diff
@@ -0,0 +1,71 @@
...
+**Service controllers classify errors.** The controller that calls an extension decides whether the error is user-caused or infrastructure-caused, and whether it should be retried:
+
+```go
+func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) error {
+    // Extension returns a plain error
+    result, err := c.mergeChecker.Check(ctx, queue, change)
+    if err != nil {
+        // Controller classifies: merge checker failure is infra, worth retrying
+        return errs.NewRetryableError(fmt.Errorf("merge check failed: %w", err))
+    }
```
behinddwalls, line 32: should controller do this? Say if the underlying extension or impl, classifies the actuall error and return retryable/non-retryable, can controller just propagate the same up the call chain stack

core/errs/README.md
```diff
@@ -0,0 +1,71 @@
...
+    if !result.Mergeable {
+        // Controller classifies: not mergeable is a user error, never retry
+        return errs.NewUserError(fmt.Errorf("not mergeable: %s", result.Reason))
+    }
+
+    // ...
+}
+```
+
+This separation keeps extensions reusable across contexts. The same `storage.ErrNotFound` might be a user error in one controller (user requested a non-existent resource) and an infra error in another (expected record is missing).
```
behinddwalls, line 43: this is a good point on why controller should decide whether to retry or not...

behinddwalls (approved):

core/errs/errs.go
sbalabanov: This may happen when the confidence in classification is low (i.e. misclassification). So it is categorized as a user error, but incorrectly, and retry fixes it while it should not.
In fact, I am ok to remove it altogether, it is here for completeness of all 4 states, but may be it should not.

sbalabanov: 409 may not be a good example. 409 happens when i.e. username that has to be unique already exists. The client is expected to retry, but with a different username, so this is not the "retry" in the way how we see it (same inputs).

core/errs/README.md
sbalabanov: I had some thinking around this and decided this approach is better (remap errors at higher level) than what we discussed offline (mapped at the component implementation).
Instead, component will output custom errors as per contract.

PR #96, opened 2026-02-26
feat(storage) Add upsert functionality for batch and batch dependent store

manjari25:
## Summary
feat(storage) Add upsert functionality for batch and batch dependent store

## What?
New method to the batch store and batch dependent store interfaces for encapsulating Upsert functionality.

## Why?
Batch manager will require to upsert the state of a batch and batch dependent.
* For batch - In the event of a failure in batch manager after an insert into the batch store; reprocessing of the request should trigger an overwrite.
* For batch dependent - Any new incoming requests can alter the dependents for existing batches.

### DIFF
--- a/extension/storage/mysql/batch_store.go
+++ b/extension/storage/mysql/batch_store.go
@@ -111,6 +111,61 @@ func (s *batchStore) UpdateState(ctx context.Context, id string, version int32, 
 	return nil
 }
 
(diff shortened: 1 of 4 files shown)


extension/storage/mysql/batch_store.go
behinddwalls, line 70: on duplicate key return error,

manjari25:
Discussed offline. Neither of these Upsert's are needed.

PR #97, opened 2026-02-26
feat(scorer): add scorer extension with heuristic and composite implementations

behinddwalls:
## Why?

SubmitQueue needs a way to estimate the success probability of a code change before landing it. Different heuristics (file count, dependency count, lines changed) each provide a signal, but no single metric is sufficient. We need a pluggable scoring system that can combine multiple signals into a single probability.

## What?

Adds a new `extension/scorer` package with:

- **`scorer.Scorer` interface** — defines `Score(ctx, ChangeStats) (float64, error)` returning a probability between 0.0 and 1.0
- **`scorer.ChangeStats` entity** — aggregates code change metrics (files changed, lines added/deleted/modified, build targets added/removed/changed, dependency count)
- **Heuristic scorer** (`extension/scorer/heuristic/`) — buckets a single extracted metric into probability ranges using configurable `Bucket` definitions and `StatFunc` extractors
- **Composite scorer** (`extension/scorer/composite/`) — wraps multiple `Scorer` instances and combines their results with a `ReduceFunc` (built-in: `Min`, `Max`, `Avg`)
- **Gomock setup** (`extension/scorer/mock/`) — build-time generated mocks for the `Scorer` interface

Both constructors return the `scorer.Scorer` interface and validate inputs (nil functions, empty scorers) by returning errors.

## Test Plan

- [x] `bazel build //extension/scorer/...` — all 5 targets compile
- [x] `bazel test //extension/scorer/heuristic:heuristic_test` — table-driven tests covering bucket matching, boundary conditions, multiple stat functions, no-match errors, and nil statFunc validation
- [x] `bazel test //extension/scorer/composite:composite_test` — table-driven tests covering Min/Max/Avg reduce, single scorer passthrough, child error propagation, three-scorer average, empty scorers error, and nil reduce error


extension/scorer/scorer.go
```diff
@@ -0,0 +1,33 @@
...
+	BuildTargetsChanged int
+	// DependencyCount is the number of downstream dependencies affected.
+	DependencyCount int
+}
+
+// Scorer computes a success probability score for a change based on its characteristics.
+type Scorer interface {
+	// Score returns a probability between 0.0 and 1.0 indicating the likelihood
+	// of a successful land for a change with the given statistics.
+	Score(ctx context.Context, stats ChangeStats) (float64, error)
```
sbalabanov: Potentially we want an ML model to work here. But what input features it wants - depend on the model. Which means we can't define them into something like `ChangeStats`. 
To conform to this, a more generic interface will be needed, something like `Score(ctx, Context, change Change)`

extension/scorer/composite/BUILD.bazel
```diff
@@ -0,0 +1,21 @@
+load("@rules_go//go:def.bzl", "go_library", "go_test")
```
sbalabanov, line 1: add README.md?

extension/scorer/composite/scorer.go
```diff
@@ -0,0 +1,65 @@
...
+import (
+	"context"
+	"fmt"
+	"slices"
+
+	"github.com/uber/submitqueue/extension/scorer"
+)
+
+// ReduceFunc combines multiple scores into a single score.
+type ReduceFunc func([]float64) float64
```
sbalabanov: seems a classical linear regression model, but how the client will know which scores belong to which domain (or featureset) to apply a reduction? It only sees a plain list.

extension/scorer/composite/scorer.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+// New creates a composite Scorer that evaluates all child scorers and combines
+// their results using the given reduce function.
+// Returns an error if scorers is empty or reduce is nil.
+func New(scorers []scorer.Scorer, reduce ReduceFunc) (scorer.Scorer, error) {
+	if len(scorers) == 0 {
+		return nil, fmt.Errorf("composite.New: scorers must not be empty")
+	}
+	if reduce == nil {
```
sbalabanov, line 63: can make it a part of (implicit) contract and avoid a runtime check. Panic is ok if not met.

sbalabanov (approved):

PR #98, opened 2026-02-27
implement change provider - github

rashmi-prithyani:
## Summary
implements change provider - github
  - Fetches PR metadata including changed files, patches, line counts, and author
  information
  - Supports stacked PRs (multiple PRs in a single request)
  - Handles pagination for PRs with >100 files
  - Validates PR staleness by comparing head SHA with submitted SHA
  - Validates stack consistency (same provider, org, and repo)
  - Partial success support: continues fetching if individual PRs fail

## Test Plan
CI and tests


extension/changeprovider/change_provider.go
behinddwalls: should it be called URI?

extension/changeprovider/github/provider.go
```diff
@@ -0,0 +1,213 @@
...
+)
+
+// Params holds the dependencies for the GitHub ChangeProvider.
+type Params struct {
+	// HTTPClient is a pre-configured HTTP client with auth (bearer token, GitHub App JWT, etc.).
+	// Auth is the caller's responsibility via HTTP transport/round-tripper.
+	HTTPClient *http.Client
+	// GraphQLURL is the GitHub GraphQL API endpoint
+	// (e.g., "https://api.github.com/graphql" or "https://ghe.example.com/api/graphql").
+	GraphQLURL string
```
behinddwalls: i think we don't need it, HTTPClient should encapsulate it

extension/changeprovider/github/provider.go
```diff
@@ -0,0 +1,213 @@
...
+// TODO add error codes for user errors (non-retryable) vs system errors.
+func (p *provider) Get(ctx context.Context, change entity.Change) ([]changeprovider.ChangeInfo, error) {
+	p.metrics.Counter("get_change_info_started").Inc(1)
+	startTime := time.Now()
+	defer func() {
+		p.metrics.Timer("get_change_info_latency").Record(time.Since(startTime))
+	}()
+
+	if len(change.URIs) == 0 {
+		p.logger.Errorw("no URIs provided in change")
```
behinddwalls: general thought on logging? we should only log at the top of the stack so we are duplicating them at each layer...so i would just log in controllers and skip here, unless swallowing errors

extension/changeprovider/github/validate.go
```diff
@@ -0,0 +1,84 @@
...
+// Stacked changes must have the same change provider (scheme), org, and repo.
+// Returns the org and repo if valid, or an error if any change is inconsistent.
+func validateChangeConsistency(
+	changeIDs []entitygithub.ChangeID,
+	logger *zap.SugaredLogger,
+	metrics tally.Scope,
+) (string, string, error) {
+	if len(changeIDs) == 0 {
+		return "", "", nil
+	}
```
behinddwalls: i think we can remove it, let it panic if empty..that's just a bug in the system at that point

extension/changeprovider/github/validate.go
```diff
@@ -0,0 +1,84 @@
...
+
+	for _, cid := range changeIDs {
+		// Validate same change provider (scheme)
+		if cid.Scheme != expectedScheme {
+			logger.Errorw("stacked changes must use same change provider",
+				"expected_provider", expectedScheme,
+				"got_provider", cid.Scheme,
+				"pr", cid.PRNumber,
+			)
+			metrics.Tagged(map[string]string{"error_type": "mixed_provider_stack"}).Counter("get_errors").Inc(1)
```
behinddwalls: i would remove log and metric both from validation util

extension/changeprovider/github/provider.go
```diff
@@ -0,0 +1,213 @@
...
+// Returns one ChangeInfo per URI (one per PR in stacked changes).
+// TODO add error codes for user errors (non-retryable) vs system errors.
+func (p *provider) Get(ctx context.Context, change entity.Change) ([]changeprovider.ChangeInfo, error) {
+	p.metrics.Counter("get_change_info_started").Inc(1)
+	startTime := time.Now()
+	defer func() {
+		p.metrics.Timer("get_change_info_latency").Record(time.Since(startTime))
+	}()
+
+	if len(change.URIs) == 0 {
```
behinddwalls: same as before, remove it, should never be empty, if it is, then likely a bug so let it panic

extension/changeprovider/github/provider.go
```diff
@@ -0,0 +1,213 @@
...
+			"total_prs", len(changeIDs),
+			"failed_count", len(fetchErrors),
+			"failed_prs", failedPRs,
+			"succeeded_count", len(changeInfos),
+		)
+		return changeInfos, fmt.Errorf("failed to fetch %d of %d PRs (failed: %v): %v",
+			len(fetchErrors), len(changeIDs), failedPRs, fetchErrors)
+	}
+
+	p.logger.Infow("successfully fetched PR data",
```
behinddwalls: make it debug?

behinddwalls (changes requested):

extension/changeprovider/change_provider.go
rashmi-prithyani: Yes, I updated to URI. Also, the previous code wasnt formatting this correctly. I have updated it to now have the URI as a string  that is similar to the input `github://org/repo/123/abc123` to uniquely identify each of the change.

extension/changeprovider/github/provider.go
```diff
@@ -0,0 +1,245 @@
...
+	httpClient *http.Client
+	graphQLURL string
+	logger     *zap.SugaredLogger
+	metrics    tally.Scope
+}
+
+// NewProvider creates a new GitHub ChangeProvider from configuration.
+func NewProvider(config Config, logger *zap.SugaredLogger, metrics tally.Scope) changeprovider.ChangeProvider {
+	if err := config.Validate(); err != nil {
+		panic(fmt.Sprintf("invalid GitHub config: %v", err))
```
behinddwalls: Should it panic? why not error?

extension/changeprovider/github/provider.go
```diff
@@ -0,0 +1,245 @@
...
+
+	// Use provided client or create default
+	httpClient := config.HTTPClient
+	if httpClient == nil {
+		timeout := config.Timeout
+		if timeout <= 0 {
+			timeout = DefaultTimeout
+		}
+		httpClient = createDefaultClient(config.Token, timeout)
+	}
```
behinddwalls: why not just always required correct client from the caller?

extension/changeprovider/github/config.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+const (
+	// DefaultTimeout is the default HTTP request timeout for GitHub API calls.
+	// This applies to the entire request/response cycle.
+	DefaultTimeout = 30 * time.Second
+)
+
+// Config holds configuration for connecting to a GitHub backend.
+type Config struct {
```
behinddwalls: why do we need it? can we not make it part of caller responsibility? The problem i have with this is somewhat very dependent on how token is generated, so having it outside allows one to be flexible about it. 

I could provide tokenProvider into oauth2 client which can create a httpclient that i can pass around without having to make provider be aware about how token is generated

extension/changeprovider/github/provider.go
```diff
@@ -0,0 +1,200 @@
...
+}
+
+// Get retrieves change information from GitHub for the provided Change.
+// Returns one ChangeInfo per URI (one per PR in stacked changes).
+// TODO add error codes for user errors (non-retryable) vs system errors.
+func (p *provider) Get(ctx context.Context, change entity.Change) ([]changeprovider.ChangeInfo, error) {
+	p.metrics.Counter("get_change_info_started").Inc(1)
+	startTime := time.Now()
+	defer func() {
+		p.metrics.Timer("get_change_info_latency").Record(time.Since(startTime))
```
behinddwalls: can you update to use the `core/metrics` for all the metrics

extension/changeprovider/github/provider.go
```diff
@@ -0,0 +1,200 @@
...
+	changeInfos, fetchErrors, failedPRs := p.fetchAllPRs(ctx, changeIDs)
+
+	// Return partial results if any PRs failed
+	if len(fetchErrors) > 0 {
+		p.logger.Errorw("failed to fetch some PRs",
+			"total_prs", len(changeIDs),
+			"failed_count", len(fetchErrors),
+			"failed_prs", failedPRs,
+			"succeeded_count", len(changeInfos),
+		)
```
behinddwalls: avoid logging, just make sure everything is in the returned error, we should only log once in the call stack for any error we return

extension/changeprovider/github/provider.go
```diff
@@ -0,0 +1,200 @@
...
+				"error", err,
+			)
+			p.metrics.Tagged(map[string]string{
+				"org":        cid.Org,
+				"repo":       cid.Repo,
+				"error_type": "fetch_pr",
+			}).Counter("get_errors").Inc(1)
+			fetchErrors = append(fetchErrors, fmt.Errorf("PR #%d: %w", cid.PRNumber, err))
+			failedPRs = append(failedPRs, cid.PRNumber)
+			continue // Continue to next PR
```
behinddwalls: do we need to continue?

extension/changeprovider/github/config.go
```diff
@@ -0,0 +1,59 @@
...
+import (
+	"net/http"
+)
+
+// Client is a GitHub API client that encapsulates connection details and authentication.
+// The client is protocol-agnostic - it provides helpers for both GraphQL and REST endpoints.
+type Client struct {
+	httpClient *http.Client
+	baseURL    string
+}
```
behinddwalls, line 12: any reason we need it? httpClient should be enough, it can have baseURL in it, i think we can get rid of config file completely.

extension/changeprovider/github/graphql.go
```diff
@@ -0,0 +1,199 @@
...
+) (*pullRequestData, error) {
+	if resp.StatusCode != http.StatusOK {
+		body, _ := io.ReadAll(resp.Body)
+		logger.Errorw("GitHub API error",
+			"status", resp.StatusCode,
+			"org", org,
+			"repo", repo,
+			"pr", prNumber,
+			"response", string(body),
+		)
```
behinddwalls: we don't need to log here at all...please remove

extension/changeprovider/github/graphql.go
```diff
@@ -0,0 +1,199 @@
...
+		metrics.Tagged(map[string]string{
+			"org":        org,
+			"repo":       repo,
+			"error_type": "api_error",
+		}).Counter("get_errors").Inc(1)
```
behinddwalls: can we just do in one place...seems like we end up emitting wat too many metrics? we just do it on callsite in provider?

extension/changeprovider/github/config.go
rashmi-prithyani: I created this config file in response to this previous comment - https://github.com/uber/submitqueue/pull/98#discussion_r2870728337

If we are ok to pass in the base url as a param to provider struct, I can revert to the previous version. By itself I dont think the `httpCleint` has facility to configure a baseURL.

extension/changeprovider/github/provider.go
rashmi-prithyani: Thought here is that if we fail to get change info for one of the PRs in a stack, we still continue to fetch changes for the other PRs in the stack. 

Then this would be marked as a retryable error and we could attempt to fetch in next attempt. That's why added the continue. What do you think ?

rashmi-prithyani:
Closing this one in favor of https://github.com/uber/submitqueue/pull/141 which has cherry picked these changes and also addressed code review comments.

PR #104, opened 2026-02-27
feat(landprovider): Adding land provider extension interface

kevinlnew:
## Summary
feat(landprovider): Adding land provider extension interface

Why?
LandProvider is the extension for landing changes.

What?
- Added the LandProvider interface with a basic Land method that takes `[]LandEntry`
- Added `LandEntry`, a basic tuple construction for landing strategy and the change being landed
- Added a land-specific error type `ErrLandRejected`, this is a specific type that will affect the control flow of the merge controller.

## Test Plan
Unit tests, CI.

## Stack
1. @ #104
1. #109
1. #105


entity/land_entry.go
```diff
@@ -0,0 +1,12 @@
+package landprovider
+
+import "github.com/uber/submitqueue/entity"
+
+// LandEntry pairs a land strategy with the change to land.
+// Each entry represents one request's contribution to a batch land operation.
+type LandEntry struct {
```
sbalabanov, line 5: it should probably be in entities, to avoid circular deps, as long as it is a part of the interface to be called from a controller

sbalabanov (changes requested):

entity/land_entry.go
kevinlnew: Moved to entity/

extension/landprovider/land_provider.go
```diff
@@ -0,0 +1,20 @@
...
+
+// LandProvider lands (merges) code changes into the target branch of a source
+// control repository. Each implementation is configured for a specific provider
+// (GitHub, GitLab, Phabricator).
+type LandProvider interface {
+	// Land merges the provided changes into the target branch of the given queue.
+	// The queue identifies the repository and target branch.
+	// Each entry contains a change and the strategy to use for landing it.
+	// Returns ErrLandRejected if the land was rejected due to the changes themselves.
+	Land(ctx context.Context, queue string, entries []entity.LandEntry) error
```
behinddwalls, line 20: What is the expectation on atomicity of merges? maybe they all depend on impl? say i send 10 land-entries with each stacked changes...is the expectation to land all atomically or per change? do we need to define it here?

kevinlnew: Yeah, we can say that it depends on implementation. My reasoning is that we can't really control how the implementation is handling it.. I would say that it's not a hard requirement, but it's implied given that it should be idempotent against calling Land. And if it's not atomic, the implementer should handle those cases to make it idempotent...?

behinddwalls (approved):

PR #105, opened 2026-02-27
feat(controller): hooking up merge controller

kevinlnew:
## Summary
feat(controller): hooking up merge controller

What?
Hooked up the merge controller with the land provider. 
Also wired up the example server with the merge controller along with default github land provider.

Idempotency checks:
- Batch state
  - if batch is already terminal, this means we can skip landing and store updates and publish the message.
  - otherwise, try to land.
- Landing:
  - Implementation can pass back `ErrIsAlreadyLanded`, which is a happy path signal to set the batch to succeeded.
- Errors
  - All the returned errors except bad payload and `ErrLandRejected` are retryable.
 

## Test Plan
Unit tests, CI

## Stack
1. #104
1. #109
1. @ #105


entity/batch.go
```diff
@@ -12,8 +12,12 @@ const (
...
 	// BatchStateSpeculating is the state of a batch that is undergoing speculative execution.
 	BatchStateSpeculating BatchState = "speculating"
-	// BatchStateFinalizing is the state of a batch that is being finalized after speculative execution.
-	BatchStateFinalizing BatchState = "finalizing"
+	// BatchStateLanding is the state of a batch that is being landed.
+	BatchStateLanding BatchState = "landing"
+	// BatchStateLandFailed is the state of a batch that has failed to land.
+	BatchStateLandFailed BatchState = "land_failed"
+	// BatchStateLandSucceeded is the state of a batch that has successfully landed.
+	BatchStateLandSucceeded BatchState = "land_succeeded"
```
sbalabanov: How does it differ from "succeeded" ?

example/server/orchestrator/main.go
```diff
@@ -361,6 +361,8 @@ func registerControllers(c consumer.Consumer, logger *zap.SugaredLogger, scope t
 		logger,
 		scope,
 		registry,
+		nil, // TODO: wire LandProvider implementation
+		nil, // TODO: wire Storage implementation
```
sbalabanov: prefer wiring up stabs, or nothing.
nils will lead to unexpected panics.

orchestrator/controller/merge/merge.go
```diff
@@ -31,82 +35,184 @@ func NewController(
...
 // Returns nil to ack (success), or error to nack (retry).
 func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) error {
 	c.metricsScope.Counter("received").Inc(1)
 
 	msg := delivery.Message()
 
-	// Deserialize request entity
-	request, err := entity.RequestFromBytes(msg.Payload)
+	// Deserialize batch entity
+	batch, err := entity.BatchFromBytes(msg.Payload)
```
sbalabanov: I think we've decided that payload is Batch ID?

orchestrator/controller/merge/merge.go
```diff
@@ -31,82 +35,184 @@ func NewController(
...
+	case entity.BatchStateLandSucceeded, entity.BatchStateLandFailed:
+		c.logger.Infow("batch already landed, skipping to publish",
+			"batch_id", batch.ID,
+			"state", string(batch.State),
+		)
+		c.metricsScope.Counter("idempotent_skip").Inc(1)
+	case entity.BatchStateLanding:
+		newState, err := c.land(ctx, batch)
+		if err != nil {
+			return err
```
sbalabanov: augment with batch id at least?

orchestrator/controller/merge/merge.go
```diff
@@ -31,82 +35,184 @@ func NewController(
...
+
+	switch batch.State {
+	case entity.BatchStateLandSucceeded, entity.BatchStateLandFailed:
+		c.logger.Infow("batch already landed, skipping to publish",
+			"batch_id", batch.ID,
+			"state", string(batch.State),
+		)
+		c.metricsScope.Counter("idempotent_skip").Inc(1)
+	case entity.BatchStateLanding:
+		newState, err := c.land(ctx, batch)
```
sbalabanov, line 96: It should also process the land error "already merged" for another idempotency check.
If it is for the future, add TODO comment.

orchestrator/controller/merge/merge.go
```diff
@@ -31,82 +35,184 @@ func NewController(
...
+		}
+
+		if err := c.storage.GetBatchStore().UpdateState(ctx, batch.ID, batch.Version, newState); err != nil {
+			c.logger.Errorw("failed to update batch state",
+				"batch_id", batch.ID,
+				"target_state", string(newState),
+				"error", err,
+			)
+			c.metricsScope.Counter("batch_update_errors").Inc(1)
+			return fmt.Errorf("failed to update batch state: %w", err)
```
sbalabanov: need to convert VersionMismatch to RetryableError

orchestrator/controller/merge/merge.go
```diff
@@ -31,82 +35,184 @@ func NewController(
...
+			"batch_id", batch.ID,
+			"state", string(batch.State),
+		)
+		c.metricsScope.Counter("unexpected_state").Inc(1)
+		return fmt.Errorf("unexpected batch state %s for batch %s", batch.State, batch.ID)
+	}
 
 	// Publish to merge signal topic
-	if err := c.publish(ctx, consumer.TopicKeyMergeSignal, request); err != nil {
+	if err := c.publish(ctx, consumer.TopicKeyMergeSignal, batch); err != nil {
```
sbalabanov: ditto - BatchID

orchestrator/controller/merge/merge.go
```diff
@@ -31,82 +35,184 @@ func NewController(
...
+	}
 
-	return nil // Success - message will be acked
+	landResult, err := c.landProvider.Land(ctx, batch.Queue, entries)
+
+	switch {
+	case err == nil:
+		c.logger.Infow("land succeeded",
+			"batch_id", batch.ID,
+			"sha", landResult.SHA,
```
sbalabanov: can include request ids

sbalabanov (changes requested):

orchestrator/controller/merge/merge.go
kevinlnew: Added the `IsAlreadyLanded` error to the interface in a previous PR in the stack and hooked it up to the `land` function

entity/batch.go
kevinlnew: Removed `land_failed` and `land_succeeded`, they weren't necessary. Just artifacts from me trying to understand state transitions. 
The merge controller now just updates the batch to the terminal states (`succeeded` and `failed`) for the "concluder" controller to pick up later.

example/server/orchestrator/main.go
kevinlnew: Wired up with new constructor

orchestrator/controller/merge/merge.go
kevinlnew: Fixed

orchestrator/controller/merge/merge.go
kevinlnew: Wrapped batch ID into error

orchestrator/controller/merge/merge.go
kevinlnew: Made all the store calls are retryable now

orchestrator/controller/merge/merge.go
kevinlnew: fixed

orchestrator/controller/merge/merge.go
kevinlnew: done

orchestrator/controller/merge/merge.go
```diff
@@ -31,82 +37,165 @@ func NewController(
...
 // Process processes a merge delivery from the queue.
-// Deserializes the request, performs the merge, and publishes to the merge signal topic.
+// Extracts the batch ID from the message, fetches the batch from storage,
+// lands the changes, and publishes the batch ID to downstream topics.
 // Returns nil to ack (success), or error to nack (retry).
 func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) error {
 	c.metricsScope.Counter("received").Inc(1)
 
 	msg := delivery.Message()
+	batchID := string(msg.Payload)
```
sbalabanov, line 64: We need to define a serialization format for this. Using direct string-to-byte conversion is Golang-specific.
Probably a separate serializable structure.

orchestrator/controller/merge/merge.go
```diff
@@ -31,82 +37,165 @@ func NewController(
...
-	if err := c.publish(ctx, consumer.TopicKeyMergeSignal, request); err != nil {
-		c.logger.Errorw("failed to publish output",
-			"request_id", request.ID,
-			"topic_key", consumer.TopicKeyMergeSignal,
+	// Fetch the batch from storage — this is the single source of truth.
+	// Terminal states mean the land outcome was already persisted on a
+	// previous attempt; skip straight to publish.
+	batch, err := c.storage.GetBatchStore().Get(ctx, batchID)
+	if err != nil {
+		c.logger.Errorw("failed to fetch batch",
```
sbalabanov, line 87: as long as we return error to upstream, we do not have to log. Otherwise most likely we'll log twice.
Logging should happen at error chain termination (top-most) point.

orchestrator/controller/merge/merge.go
```diff
@@ -31,82 +37,165 @@ func NewController(
...
+			"batch_id", batchID,
 			"error", err,
 		)
-		c.metricsScope.Counter("publish_errors").Inc(1)
-		return errs.NewRetryableError(fmt.Errorf("failed to publish to merge-signal: %w", err))
+		c.metricsScope.Counter("batch_fetch_errors").Inc(1)
```
sbalabanov, line 91: plz rebase and integrate with new metrics infrastructure that processes errors automagically

orchestrator/controller/merge/merge.go
```diff
@@ -31,82 +37,165 @@ func NewController(
...
+		}
+
+		if err := c.storage.GetBatchStore().UpdateState(ctx, batch.ID, batch.Version, newState); err != nil {
+			c.logger.Errorw("failed to update batch state",
+				"batch_id", batch.ID,
+				"target_state", string(newState),
+				"error", err,
+			)
+			c.metricsScope.Counter("batch_update_errors").Inc(1)
+			return errs.NewRetryableError(fmt.Errorf("failed to update batch state: %w", err))
```
sbalabanov, line 108: it should only be retryable if the error is expected type, like version mismatch, but we seem to retry on all of the errors? If the database is operational but "table does not exist" is the error, this is not retryable.

orchestrator/controller/merge/merge.go
```diff
@@ -31,82 +37,165 @@ func NewController(
...
+		}
+
+		batch.State = newState
+	} else {
+		c.logger.Infow("batch already in terminal state, skipping to publish",
```
sbalabanov, line 113: may be a warning level, as this should only happen on infra errors

orchestrator/controller/merge/merge.go
```diff
@@ -31,82 +37,165 @@ func NewController(
...
+			Strategy: request.LandStrategy,
+			Change:   request.Change,
+		})
 	}
 
-	msg := entityqueue.NewMessage(request.ID, payload, request.Queue, nil)
+	err := c.landProvider.Land(ctx, batch.Queue, entries)
+
+	switch {
+	case err == nil, landprovider.IsAlreadyLanded(err):
```
sbalabanov, line 172: may be at least log landprovider.IsAlreadyLanded, but better tally too?
this also should not happen for a normal workflow, so good to have some observability
also, comment here scenarios which could cause this

orchestrator/controller/merge/merge.go
```diff
@@ -31,82 +37,165 @@ func NewController(
...
+			"already_landed", landprovider.IsAlreadyLanded(err),
+		)
+		return entity.BatchStateSucceeded, nil
+	case landprovider.IsLandRejected(err):
+		c.logger.Errorw("land rejected",
+			"batch_id", batch.ID,
+			"error", err,
+		)
+		c.metricsScope.Counter("land_rejected").Inc(1)
+		return entity.BatchStateFailed, nil
```
sbalabanov, line 185: we'll need to propagate the error information about why land was rejected. Can be done in another diff, but then needs a TODO comment.

sbalabanov (changes requested):

PR #106, opened 2026-02-27
feat(extension): Add BuildManager interface for CI/CD integration

djuloori2794:
Add vendor-agnostic BuildManager interface for managing builds with external CI providers (BuildKite, Jenkins, etc.). Enables orchestrator to schedule builds, poll status, and cancel running builds.

### DIFF
--- a/extension/build/README.md
+++ b/extension/build/README.md
@@ -1,0 +1,72 @@
+# Build Manager
+
+The BuildManager extension provides a clean abstraction for integrating with CI/CD systems like BuildKite, Jenkins, and others. It allows the Orchestrator service to schedule builds, poll their status, and cancel running builds without being coupled to any specific CI provider.
+
+## Interface
+
+### BuildManager
+
+Schedules builds, polls their status, and cancels running builds.
+
+```go
+type BuildManager interface {
+    Schedule(ctx context.Context, queueName string, changes []entity.BuildChange) (string, error)
+    Poll(ctx context.Context, buildID string) (entity.BuildStatus, entity.BuildMetadata, error)
+    CancelBuild(ctx context.Context, buildID string) error
+    Close() error
+}
+```
+
+- **Schedule**: Submits changes to the CI provider for processing. Returns a unique build ID.
+- **Poll**: Retrieves the current status and metadata of a build.
+- **CancelBuild**: Requests cancellation of a build (asynchronous, does not wait for completion).
+- **Close**: Gracefully shuts down the build manager and releases resources.
+
+### Entities
+
+```go
+type BuildChange struct {
+    Change entity.Change   // List of URIs with provider encoded in schema
+    Action ChangeAction    // "apply" or "validate"
+}
+
+type BuildMetadata map[string]string  // Implementation-defined metadata (e.g., build URL, duration)
+```
+
+### Errors
+
+- **ErrBuildNotFound**: Returned by Poll and CancelBuild when the build doesn't exist.
+- **ErrInvalidRequest**: Returned by Schedule when validation fails.
+
+## Usage
+
+```go
+mgr := buildkite.NewBuildManager(config)
+defer mgr.Close()
+
+// Schedule a build
+changes := []entity.BuildChange{
+    {Change: entity.Change{URIs: []string{"github://uber/repo/pull/123/abc"}}, Action: entity.ChangeActionValidate},
+}
+buildID, err := mgr.Schedule(ctx, "my-queue", changes)
+
+// Poll for status
+status, metadata, err := mgr.Poll(ctx, buildID)
+if status.IsTerminal() {
+    fmt.Println("Build finished:", status)
+}
+
+// Cancel if needed
+err = mgr.CancelBuild(ctx, buildID)
+```
+
+## Implementing a Backend
+
+1. Create `extension/build/{backend}/` directory
+2. Implement the `BuildManager` interface
+3. Map `entity.BuildChange` actions to backend-specific job configurations
+4. Handle build status mapping to `entity.BuildStatus` values
+
+**Thread-safety**: All implementations must be thread-safe and support concurrent operations.
+
+**Singleton Design**: Implementations are long-lived singletons (one per build provider) initialized at service startup, similar to Storage and other extension components. They should manage connection pooling, caching, and other resources for the lifetime of the service.
--- a/extension/build/build_manager.go
+++ b/extension/build/build_manager.go
@@ -1,0 +1,72 @@
+package build
+
+//go:generate mockgen -source=build_manager.go -destination=mock/build_manager.go -package=mock
+
+import (
+	"context"
+
+	"github.com/uber/submitqueue/entity"
+)
+
+// BuildManager is a vendor-agnostic interface for managing builds with external CI/CD providers.
+// Implementations provide integration with specific CI systems (BuildKite, Jenkins, etc.)
+// to schedule builds, poll their status, and cancel running builds.
+//
+// Implementations are long-lived singletons (one per build provider) initialized at service
+// startup, similar to Storage and other extension components. They should manage connection
+// pooling, caching, and other resources for the lifetime of the service.
+//
+// All implementations must be thread-safe and support concurrent operations.
+type BuildManager interface {
+	// Schedule submits a list of changes to the CI provider for processing.
+	// Each change specifies an action (validate or apply) to perform.
+	//
+	// The implementation is responsible for:
+	//   - Looking up the job name from the queue configuration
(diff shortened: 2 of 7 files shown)


extension/build/build_manager.go
```diff
@@ -0,0 +1,72 @@
...
+	//
+	// Parameters:
+	//   - ctx: Request context for cancellation and timeouts
+	//   - queueName: Name of the queue processing these changes. Used to look up job configuration.
+	//   - changes: List of changes to process. Order may be significant for dependencies.
+	//
+	// Returns:
+	//   - string: Unique build ID that can be used with Poll and CancelBuild methods
+	//   - error: ErrInvalidRequest if validation fails
+	Schedule(ctx context.Context, queueName string, changes []entity.BuildChange) (string, error)
```
behinddwalls, line 37: does it make sense to say just "TriggerBuild", whether it's done right away vs scheduled...it's impl details?

extension/build/build_manager.go
```diff
@@ -0,0 +1,72 @@
...
+	// This is a synchronous call that queries the provider's API.
+	//
+	// Parameters:
+	//   - buildID: Build ID string
+	//
+	// Returns:
+	//   - BuildStatus: Current state of the build
+	//   - BuildMetadata: Additional metadata about the build (e.g., build URL, commit SHA, duration)
+	//   - error: ErrBuildNotFound if the build doesn't exist
+	Poll(ctx context.Context, buildID string) (entity.BuildStatus, entity.BuildMetadata, error)
```
behinddwalls, line 49: I would rename this to `BuildStatus".. poller is just an impl detail?

extension/build/build_manager.go
```diff
@@ -0,0 +1,72 @@
...
+	//
+	// SubmitQueue will mark the build as BuildStatusCancelled immediately without waiting
+	// for confirmation from the build system.
+	//
+	// Parameters:
+	//   - buildID: Build ID string
+	//
+	// Returns:
+	//   - error: ErrBuildNotFound if the build doesn't exist
+	CancelBuild(ctx context.Context, buildID string) error
```
behinddwalls, line 65: we can drop Build and say `Cancel`

extension/build/build_manager.go
```diff
@@ -0,0 +1,72 @@
...
+
+//go:generate mockgen -source=build_manager.go -destination=mock/build_manager.go -package=mock
+
+import (
+	"context"
+
+	"github.com/uber/submitqueue/entity"
+)
+
+// BuildManager is a vendor-agnostic interface for managing builds with external CI/CD providers.
```
behinddwalls, line 11: can we avoid saying CI/CD completely in this interface as it can be anything as per impl...

extension/build/build_manager.go
```diff
@@ -0,0 +1,72 @@
...
+// Implementations provide integration with specific CI systems (BuildKite, Jenkins, etc.)
+// to schedule builds, poll their status, and cancel running builds.
```
behinddwalls, line 13: poll is again a semantics of impl

extension/build/build_manager.go
```diff
@@ -0,0 +1,72 @@
...
+//
+// Implementations are long-lived singletons (one per build provider) initialized at service
+// startup, similar to Storage and other extension components. They should manage connection
+// pooling, caching, and other resources for the lifetime of the service.
+//
```
behinddwalls, line 18: i would avoid saying this, maybe simplify it.

BuildManager is responsible for triggering, cancelling and checking status of builds generated by underlying implementations.

Implementations must be thread-safe.

extension/build/build_manager.go
```diff
@@ -0,0 +1,72 @@
...
+// All implementations must be thread-safe and support concurrent operations.
+type BuildManager interface {
+	// Schedule submits a list of changes to the CI provider for processing.
```
behinddwalls, line 21: no CI in docs

extension/build/build_manager.go
```diff
@@ -0,0 +1,72 @@
...
+	// Each change specifies an action (validate or apply) to perform.
+	//
+	// The implementation is responsible for:
+	//   - Looking up the job name from the queue configuration
+	//   - Creating appropriate builds/jobs for each change based on its action
+	//   - Handling dependencies between changes (order may be significant)
```
behinddwalls, line 27: generally order is always important

extension/build/build_manager.go
```diff
@@ -0,0 +1,72 @@
...
+// pooling, caching, and other resources for the lifetime of the service.
+//
+// All implementations must be thread-safe and support concurrent operations.
+type BuildManager interface {
+	// Schedule submits a list of changes to the CI provider for processing.
+	// Each change specifies an action (validate or apply) to perform.
+	//
+	// The implementation is responsible for:
+	//   - Looking up the job name from the queue configuration
+	//   - Creating appropriate builds/jobs for each change based on its action
```
behinddwalls, line 26: jobs is very impl specific? so we should just say build

extension/build/build_manager.go
```diff
@@ -0,0 +1,72 @@
...
+	//   - Handling dependencies between changes (order may be significant)
+	//
+	// Parameters:
+	//   - ctx: Request context for cancellation and timeouts
+	//   - queueName: Name of the queue processing these changes. Used to look up job configuration.
+	//   - changes: List of changes to process. Order may be significant for dependencies.
+	//
+	// Returns:
+	//   - string: Unique build ID that can be used with Poll and CancelBuild methods
```
behinddwalls, line 35: replace Poll with BuildStatus or for checking the build status

extension/build/build_manager.go
sbalabanov: Yet we made it a part of the interface.

extension/build/build_manager.go
sbalabanov: This is what I asked for so I take this comment.
I think comments need to provide an instruction of how the implementation will be used - as it will ultimately dictate the implementation itself. Can the implementation hold heavy state in the object, or should it use a shared state elsewhere?
A requirement on just thread safety may not be enough to create efficient implementations, as someone implementing them may oversynchronize for the sake of safety.

extension/build/build_manager.go
sbalabanov: I think this is another constraint that we need to put on the implementation by explicitly stating the expected behavior in comments. Which is: implementation should be asynchronous, return fast, and Poll is to be used to check statuses.

sbalabanov: which means that "done right away" is not an option to the implementation.
If we believe that the implementation can deduce builds which already happen based on provided change, it can return the build status along with the buildID.
"123, accepted, nil"
"124, succeeded, nil"
So SQ will save on not doing the next polling call.

extension/build/build_manager.go
sbalabanov: Not going to nitpick names, but here again is where behavior matters. We can say in the comments, that unlike "Schedule" this implementation might be synchronous and potentially lengthy.
Full abstractions are not efficient unless they are implemented according to expected usage.
Unfortunately, the language itself has limited ways to contract this so we have to rely on conventions (explicit in comments).

extension/build/README.md
```diff
@@ -0,0 +1,72 @@
...
+The BuildManager extension provides a clean abstraction for integrating with CI/CD systems like BuildKite, Jenkins, and others. It allows the Orchestrator service to schedule builds, poll their status, and cancel running builds without being coupled to any specific CI provider.
+
+## Interface
+
+### BuildManager
+
+Schedules builds, polls their status, and cancels running builds.
+
+```go
+type BuildManager interface {
```
sbalabanov, line 12: plz avoid copying the code to the docs.

sbalabanov: examples on "how to" are ok

extension/build/README.md
```diff
@@ -0,0 +1,72 @@
...
+## Implementing a Backend
+
+1. Create `extension/build/{backend}/` directory
+2. Implement the `BuildManager` interface
+3. Map `entity.BuildChange` actions to backend-specific job configurations
+4. Handle build status mapping to `entity.BuildStatus` values
+
+**Thread-safety**: All implementations must be thread-safe and support concurrent operations.
+
+**Singleton Design**: Implementations are long-lived singletons (one per build provider) initialized at service startup, similar to Storage and other extension components. They should manage connection pooling, caching, and other resources for the lifetime of the service.
```
sbalabanov, line 72: Also important to note how transient errors are handled. Are the implementation expected to reestablish the connection if it is lost, and give back errors on interface requests in the interim?

sbalabanov (changes requested):

PR #108, opened 2026-03-02
feat(queue): per-partition isolation and fix subscriber shutdown race conditions

behinddwalls:
## Summary

- **Per-partition isolation in consumer**: deliveries are dispatched to per-partition goroutines so a slow message on one partition does not block others. Per-partition ordering is preserved.
- **Per-partition workers in MySQL subscriber**: each leased partition gets its own polling goroutine, managed by a supervisor that reconciles workers based on current leases.
- **Fix shutdown race conditions**: separate `workerWg` tracks partition workers independently from the supervisor. `workerWg.Wait()` blocks before closing `deliveryCh` to prevent sends on a closed channel. Workers are removed from the map immediately on stop so reconciliation can start replacements without zombie entries.

## Test plan

- [x] `TestConsumer_PerPartitionProcessing` — verifies a slow partition A does not block partition B
- [x] `TestConsumer_PartitionOrdering` — verifies messages within a single partition are processed in order
- [x] `TestConsumer_PartitionWorkerCleanup` — verifies all partition goroutines exit cleanly on Stop()
- [x] `TestSubscriber_ReconcilePartitionWorkers` — verifies workers are started/stopped based on lease changes
- [x] `TestSubscriber_PartitionWorkerPollAndDeliver` — verifies a partition worker delivers messages
- [x] `TestSubscriber_StopAllWorkers` — verifies all workers stop gracefully
- [x] All existing tests pass via `bazel test`

### DIFF
--- a/core/consumer/README.md
+++ b/core/consumer/README.md
@@ -2,6 +2,22 @@
 
 The consumer package orchestrates queue message processing. It manages subscription lifecycle, message consumption, ack/nack, and graceful shutdown.
 
+## Architecture
+
+```
+Consumer
+  ├── Controller A (topic: "request")
+  │     └── consumeLoop
+  │           ├── processPartition("part-1")  ← serial per partition
+  │           ├── processPartition("part-2")
+  │           └── processPartition("part-3")
+  └── Controller B (topic: "build")
+        └── consumeLoop
+              └── processPartition("part-1")
+```
+
+The consumer spawns one `consumeLoop` goroutine per controller. Each `consumeLoop` dispatches deliveries to per-partition goroutines, preserving ordering within each partition while processing different partitions in parallel.
+
 ## Interfaces
 
 ### Consumer
@@ -9,7 +25,11 @@ The consumer package orchestrates queue message processing. It manages subscript
 The top-level orchestrator. Register controllers, start consuming, and stop gracefully.
 
 ```go
-c := consumer.New(logger, scope, queue, "worker-hostname")
+registry, _ := consumer.NewTopicRegistry([]consumer.TopicConfig{
+    {Key: consumer.TopicKeyRequest, Name: "request", Queue: q, Subscription: subConfig},
+})
+
+c := consumer.New(logger, scope, registry)
 
 c.Register(myController)
 c.Start(ctx)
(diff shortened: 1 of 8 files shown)


core/consumer/consumer.go
```diff
@@ -192,6 +219,20 @@ func (m *consumer) consumeLoop(ctx context.Context, controller Controller, deliv
...
+	// and is never removed — partitions are stable for the lifetime of a subscription.
+	partitionChs := make(map[string]chan queue.Delivery)
+	var wg sync.WaitGroup
+	defer func() {
+		// Close all partition channels to signal processPartition goroutines to exit,
+		// then wait for them to finish draining.
+		for _, ch := range partitionChs {
+			close(ch)
+		}
+		wg.Wait()
```
sbalabanov: I am not sure if deferred functions are good for something like this - as it has to wait. An important part of the code flow does not lay in its logical place in a code if reading top down.
May be enclose it into a separate function instead?

core/consumer/consumer_test.go
```diff
@@ -586,3 +588,216 @@ func TestConsumer_ErrorMetrics(t *testing.T) {
...
+	mockDelB.EXPECT().DeliveryID().Return(msgB.ID).AnyTimes()
+	mockDelB.EXPECT().Ack(gomock.Any()).Return(nil).MaxTimes(1)
+
+	deliveryChan <- mockDelB
+
+	// Partition B should be processed within a short time
+	select {
+	case <-partBDone:
+		assert.True(t, partBProcessed.Load(), "partition B should have been processed")
+	case <-time.After(5 * time.Second):
```
sbalabanov: I'd prefer to rely on test timeout in case this happens instead of hardcoded timeouts.
At least they are adjustable on a higher level by infrastructure.

core/consumer/consumer_test.go
```diff
@@ -586,3 +588,216 @@ func TestConsumer_ErrorMetrics(t *testing.T) {
...
+	c := consumer.New(logger, tally.NoopScope, reg)
+
+	var mu sync.Mutex
+	var order []string
+	allDone := make(chan struct{})
+
+	handler := consumermock.NewMockController(ctrl)
+	setupController(handler, "test-handler", consumer.TopicKeyRequest, "test-group",
+		func(ctx context.Context, delivery consumer.Delivery) error {
+			mu.Lock()
```
sbalabanov, line 703: took me a while to understand why mutex + shared object used instead of channel to signal back, may be worth a short comment

core/consumer/README.md
```diff
@@ -28,15 +48,48 @@ Business logic for processing queue messages. Implement this interface to handle
...
 }
 ```
 
 ### Delivery
 
-A restricted view of a queue delivery exposed to controllers. Hides Ack/Nack (handled automatically by Consumer) while exposing message data and `ExtendVisibilityTimeout`.
+A restricted view of a queue delivery exposed to controllers. Hides Ack/Nack/Reject (handled automatically by Consumer) while exposing message data and `ExtendVisibilityTimeout`.
+
+```go
+type Delivery interface {
```
sbalabanov: may be retyping code is not needed in docs

extension/queue/mysql/subscriber_test.go
```diff
@@ -254,3 +255,211 @@ func TestSubscriber_Close(t *testing.T) {
...
+
+	// Stop all workers
+	s.stopAllWorkers(sub)
+
+	// Verify all done channels are closed
+	for _, doneCh := range doneChans {
+		select {
+		case <-doneCh:
+			// Worker exited
+		case <-time.After(2 * time.Second):
```
sbalabanov: ditto

extension/queue/mysql/subscriber_test.go
```diff
@@ -254,3 +255,211 @@ func TestSubscriber_Close(t *testing.T) {
...
+			t.Fatal("worker did not stop within timeout")
+		}
+	}
+
+	sub.workersMu.Lock()
```
sbalabanov: is it needed? workers are guaranteed to be done

extension/queue/README.md
```diff
@@ -7,6 +7,14 @@ Vendor-agnostic interfaces for pub/sub messaging systems.
 ### Queue
 Creates publishers and subscribers.
 
+```go
```
sbalabanov: code in docs

sbalabanov (approved):

core/consumer/consumer.go
behinddwalls: Yeah..it's a bit all mixed, let me extract it out.

PR #112, opened 2026-03-03
feat(metrics): instrument extension implementations with core/metrics

behinddwalls:
## Summary
- Add `metrics.Begin`/`Complete` lifecycle tracking to all uninstrumented extension implementations: `counter/mysql`, `storage/mysql` (all 6 stores), `scorer/heuristic`, and `scorer/composite`
- Each constructor now accepts a `tally.Scope` and every interface method emits called/succeeded/failed counters plus latency timers with error classification tags
- Update all call sites (gateway, orchestrator, integration tests, unit tests) to pass the scope parameter

## Test plan
- [x] All unit tests pass (`make test` — 24/24 passing)
- [x] Full build succeeds (`make build`)
- [x] `make gazelle` confirms BUILD files are in sync
- [x] Integration tests pass with `tally.NoopScope` (Docker-based, run in CI)

### DIFF
--- a/extension/counter/mysql/counter.go
+++ b/extension/counter/mysql/counter.go
@@ -5,21 +5,26 @@ import (
 	"database/sql"
 	"fmt"
 
+	"github.com/uber-go/tally/v4"
+	"github.com/uber/submitqueue/core/metrics"
 	"github.com/uber/submitqueue/extension/counter"
 )
 
 type mysqlCounter struct {
-	db *sql.DB
+	db    *sql.DB
+	scope tally.Scope
 }
 
 // NewCounter creates a new MySQL-backed Counter.
-func NewCounter(db *sql.DB) counter.Counter {
-	return &mysqlCounter{db: db}
+func NewCounter(db *sql.DB, scope tally.Scope) counter.Counter {
+	return &mysqlCounter{db: db, scope: scope}
 }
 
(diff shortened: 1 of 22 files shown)


extension/counter/mysql/counter.go
```diff
@@ -5,21 +5,26 @@ import (
...
+func NewCounter(db *sql.DB, scope tally.Scope) counter.Counter {
+	return &mysqlCounter{db: db, scope: scope}
 }
 
 // Next atomically increments the counter for the given domain and returns the new value.
 // Uses MySQL's LAST_INSERT_ID() to set the value atomically and read the incremented value.
-func (c *mysqlCounter) Next(ctx context.Context, domain string) (int64, error) {
+func (c *mysqlCounter) Next(ctx context.Context, domain string) (ret int64, retErr error) {
+	op := metrics.Begin(c.scope, "next")
+	defer func() { op.Complete(retErr) }()
```
sbalabanov, line 27: Is the expectation now to have named return types? Or not using deferred functions...

sbalabanov (approved):

extension/counter/mysql/counter.go
behinddwalls: Yes..that's side effect of op.compete() in defer...alternative way would be to just do it without defer in all places

PR #114, opened 2026-03-03
feat(controller) Add and scaffold Cancel API in gateway

manjari25:
## Summary
feat(controller) Add and scaffold Cancel API in gateway

## What?
* Cancel API support
* Add new topic - cancel to pass the request along to orchestrator

## Why?
* Requests once submitted for land can be cancelled. These changes set up the framework needed to do this.


gateway/controller/cancel.go
```diff
@@ -0,0 +1,83 @@
...
+// Cancel handles the cancel request and returns a response.
+// It publishes the sqid to the cancel topic for async processing.
+// No validation is performed — the orchestrator handles that.
+func (c *CancelController) Cancel(ctx context.Context, req *pb.CancelRequest) (*pb.CancelResponse, error) {
+	start := time.Now()
+	defer func() {
+		c.metricsScope.Timer("cancel_request_latency").Record(time.Since(start))
+	}()
+
+	c.metricsScope.Counter("cancel_request_count").Inc(1)
```
behinddwalls: can we use the core/metrics for these?

gateway/controller/cancel.go
```diff
@@ -0,0 +1,83 @@
...
+
+	sqid := req.Sqid
```
behinddwalls, line 42: do we need to check if it's empty? i think we said, we don't care about invalid id so it's fine for other values

behinddwalls (approved):

gateway/controller/cancel.go
manjari25: I'd prefer to keep all validations in the orchestrator.

gateway/controller/cancel.go
manjari25: Updated

gateway/proto/gateway.proto
```diff
@@ -73,6 +73,18 @@ message LandResponse {
...
 
+// CancelRequest defines a request to cancel a previously submitted land request.
+message CancelRequest {
+    // Globally unique identifier for the land request to cancel, as returned by Land.
+    string sqid = 1;
+}
+
+// CancelResponse defines the response to a cancel request.
+message CancelResponse {
+    // Globally unique identifier for the land request that was cancelled.
```
sbalabanov: "that was requested to be cancelled"
Do we need a response, however? sqid is provided by the client anyways.
CancelResponse may however return the current state of the request (i.e. "building" or already "cancelled") from the request log.

gateway/proto/gateway.proto
```diff
@@ -103,4 +115,8 @@ service SubmitQueueGateway {
     // Land lands a set of code changes into a target branch, performing the necessary validations across all other changes in the queue.
     // The processing is asynchronous and returns a LandResponse immediately. The land request is processed in the background.
     rpc Land(LandRequest) returns (LandResponse) {}
+
+    // Cancel cancels a previously submitted land request. The cancellation is asynchronous and
```
sbalabanov: may be say "requests cancellation". Also note that cancellation is not guaranteed and how to check for the actual status.

gateway/controller/cancel.go
sbalabanov: For emptyness we should probably check.
In this controller we'll have to add a record to the request log. If id is null, it would create invalid record.

sbalabanov: For emptyness we should probably check.
In this controller we'll have to add a record to the request log. If id is null, it would create invalid record.
i.e. the request id should be valid, but it may not exist, such validation is deferred.

gateway/controller/cancel.go
```diff
@@ -0,0 +1,78 @@
...
+	)
+
+	return &pb.CancelResponse{
+		Sqid: sqid,
+	}, nil
+}
+
+// publishToQueue publishes a cancel request to the cancel queue for async processing.
+func (c *CancelController) publishToQueue(ctx context.Context, sqid string) error {
+	payload := []byte(sqid)
```
sbalabanov: direct string to byte conversion is language-specific.
We need a proper serialization format, likely as a separate json-serializable structure with a single field.

gateway/controller/cancel.go
```diff
@@ -0,0 +1,78 @@
...
+
+	// Create queue message
+	// - Message ID: sqid for idempotency
+	// - Payload: sqid as bytes
+	// - Partition key: sqid (ensures ordering per request)
```
sbalabanov: how would it be ordered?
what if 2 sqids are being cancelled at the same time: "go/2" and "go/10"

sbalabanov (changes requested):

gateway/controller/cancel.go
manjari25: Updated.

gateway/proto/gateway.proto
manjari25: Addressed.

gateway/proto/gateway.proto
manjari25: Updated the comment.

gateway/controller/cancel.go
manjari25: Updated

gateway/controller/cancel.go
manjari25: On second thought, I am not sure if we care about ordering for cancellation. I don't think there is any inherent order we need to enforce. We can simply leave the partition key empty. Thoughts?

sbalabanov: I think order does not matter here

gateway/controller/cancel.go
```diff
@@ -0,0 +1,99 @@
...
+// It publishes the sqid to the cancel topic for async processing.
+// No validation is performed — the orchestrator handles that.
+func (c *CancelController) Cancel(ctx context.Context, req *pb.CancelRequest) (resp *pb.CancelResponse, retErr error) {
+	op := metrics.Begin(c.metricsScope, "cancel")
+	defer func() { op.Complete(retErr) }()
+
+	sqid := req.Sqid
+
+	if sqid == "" {
+		return &pb.CancelResponse{
```
sbalabanov, line 45: it should be INVALID_ARGUMENT + BadMessage first class GRPC error
`errdetails.BadRequest`
`code.InvalidArgument`
```
import (
	"google.golang.org/grpc/codes"
	"google.golang.org/grpc/status"
	"google.golang.org/genproto/googleapis/rpc/errdetails"
)
```

sbalabanov: The transformation from Go errors to GRPC errors is better to be done at a certain distinct function, means `Cancel()` function likely wants to wrap another `cancelRequest(...) error` function which is to return first class `validationError`.

entity/cancel.go
```diff
@@ -0,0 +1,22 @@
+package entity
+
+import "encoding/json"
+
+// Cancel represents a request to cancel a previously submitted land request.
+// The object is immutable after creation.
+type Cancel struct {
```
sbalabanov, line 7: mb `CancelRequest`?

sbalabanov: not critical

gateway/proto/gateway.proto
```diff
@@ -73,6 +73,40 @@ message LandResponse {
...
+}
+
+// CancelResponse defines the response to a cancel request.
+message CancelResponse {
+    // Globally unique identifier for the land request that was requested to be cancelled.
+    string sqid = 1;
+    // Current status of the land request at the time of cancellation request.
+    RequestStatus current_status = 2;
+    // Optional error describing why the cancellation could not be completed.
+    Error error = 3;
```
sbalabanov, line 107: should be first class grpc error

gateway/proto/gateway.proto
```diff
@@ -103,4 +137,12 @@ service SubmitQueueGateway {
     // Land lands a set of code changes into a target branch, performing the necessary validations across all other changes in the queue.
     // The processing is asynchronous and returns a LandResponse immediately. The land request is processed in the background.
     rpc Land(LandRequest) returns (LandResponse) {}
+
+    // Cancel requests cancellation of a previously submitted land request.
+    // The cancellation is asynchronous and published to a queue for
+    // processing. Note that cancellation is best effort and not guaranteed.
+    // To check for the status of the request, use Status API (TODO). A
+    // "Cancelled" status indicates that the requested cancellation was
+    // successful.
```
sbalabanov, line 146: may be update the last sentence to describe that it returns the current status, whatever it is

gateway/controller/cancel.go
```diff
@@ -0,0 +1,99 @@
...
+		metricsScope: scope,
+		publisher:    publisher,
+		topic:        topic,
+	}
+}
+
+// Cancel handles the cancel request and returns a response.
+// It publishes the sqid to the cancel topic for async processing.
+// No validation is performed — the orchestrator handles that.
+func (c *CancelController) Cancel(ctx context.Context, req *pb.CancelRequest) (resp *pb.CancelResponse, retErr error) {
```
sbalabanov, line 38: Note to future us (non blocking): we should discuss if controllers should receive proto structures directly. It seems that the final surfacing of the error should be a responsibility of a service, not a controller.

gateway/proto/gateway.proto
```diff
@@ -73,6 +73,40 @@ message LandResponse {
...
+    // Initial state of a land request. Confirmed by the system but processing has not started.
+    NEW = 1;
+    // The request has been validated (duplicate check, merge check etc.) successfully.
+    VALIDATED = 2;
+    // The request is being processed.
+    PROCESSING = 3;
+    // The request has been successfully processed and landed. Terminal state.
+    LANDED = 4;
+    // The cancellation request has been accepted and is being processed.
+    CANCELLATION_ACCEPTED = 5;
```
sbalabanov, line 95: may be just "cancelling"

sbalabanov (changes requested):

PR #116, opened 2026-03-04
feat(ci): add lint, tidy, and formatting CI jobs

behinddwalls:
## Summary
- Replace custom `yamllint` Go tool with `yamlfmt` (google/yamlfmt) via `go run`
- Replace `gofmt` with `goimports` for Go formatting + import grouping
- Remove custom coverage tool and all CI coverage enforcement
- Add `assert_clean` pattern for lint checks (format, then verify no changes)
- Integrate license header checking into lint pipeline

## Test plan
- [x] `make build` passes
- [x] `make test` — all 25 unit tests pass
- [x] `make fmt` — goimports and yamlfmt run successfully
- [x] Verify no remaining references to coverage or yamllint

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,11 +13,39 @@ on:
 
 permissions:
   contents: read
-  pull-requests: read
 
 jobs:
   # ---------------------------------------------------------------------------
-  # BUILD AND UNIT TESTS (special case - Gazelle + build + unit tests)
+  # LINT (gofmt via Go SDK, yamlfmt via go run)
+  # ---------------------------------------------------------------------------
+  lint:
+    name: Lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/setup
+
+      - name: Run linters
+        run: make lint
+
+  # ---------------------------------------------------------------------------
+  # TIDY (module files + BUILD files in sync)
+  # ---------------------------------------------------------------------------
+  tidy:
+    name: Tidy
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/setup
+
+      - name: Check module files are tidy
+        run: make check-tidy
+
+      - name: Check BUILD files are up to date
+        run: make check-gazelle
+
+  # ---------------------------------------------------------------------------
+  # BUILD AND UNIT TESTS
   # ---------------------------------------------------------------------------
   build-and-unit-test:
     name: Build and Unit Test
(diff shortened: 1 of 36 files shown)


doc/howto/COVERAGE.md
```diff
@@ -0,0 +1,86 @@
...
+## Overview
+
+CI enforces minimum code coverage thresholds on every PR. Coverage is computed from unit tests using Bazel's built-in `coverage` command, which produces LCOV reports. A custom tool (`tool/coverage/`) parses the LCOV data, computes overall and diff coverage, generates an HTML report, and enforces thresholds.
+
+## Thresholds
+
+| Metric | Default | Makefile Variable |
+|--------|---------|-------------------|
+| Overall coverage | 80% | `COVERAGE_THRESHOLD` |
+| Diff coverage (new/changed lines) | 85% | `DIFF_COVERAGE_THRESHOLD` |
```
sbalabanov: should we do diff coverage enforcement at all? it penalizes small changes

tool/coverage/BUILD.bazel
```diff
@@ -0,0 +1,24 @@
+load("@rules_go//go:def.bzl", "go_binary", "go_library", "go_test")
+
+go_library(
+    name = "coverage_lib",
+    srcs = ["main.go"],
```
sbalabanov, line 5: why are linters included as sources? Is there an origin or this is generated? In which case, we'll have to support the tool as well. I'd prefer a dependency if such option exists.

sbalabanov (changes requested):

doc/howto/COVERAGE.md
behinddwalls: I am fine to remove it and just keep it, informational. Sometimes it gets harder to match the thresholds

tool/coverage/BUILD.bazel
behinddwalls: so for yaml, we have `yamllint` which is python tool, we could set-up python and then pip install but makes it a non-hermetic unless i can get python/pip from bazel itself...so right now it used yaml go lib and then wrapper on top to do lint. 

For coverage, let me see i think there is some tool but again requires installation

doc/howto/COVERAGE.md
behinddwalls: Agreed — removed diff coverage enforcement. It's now informational only: markers (terminal `*` and markdown ⚠️) still appear for packages below the diff threshold, but the build won't fail on it.

tool/coverage/BUILD.bazel
behinddwalls: BUILD.bazel files are generated by Gazelle (`make gazelle`), not hand-written. The coverage tool and yamllint are custom Go sources we maintain — same pattern as go-code which also uses custom tools for coverage analysis (`src/code.uber.internal/devexp/code-coverage/`) and ships a pre-built yamllint PEX. There's no reusable Bazel rule for either; both require a custom wrapper on top of `bazel coverage` LCOV output or a YAML parser.

sbalabanov: Can we use `yamlfmt` redistributable binaries with bazel's platform resolution?

sbalabanov: The general concern that this repo is not a good place to have the source of truth for these tools, it confuses people and AI. If we absolutely need to implement them, better them be into separate oss repos.

behinddwalls: removed coverage tool, will add codecov later

sbalabanov (approved):

PR #117, opened 2026-03-04
feat(requestlog): Implement Request Log and state reconciliation

sbalabanov:
## Summary
feat(requestlog): Implement Request Log and state reconciliation


entity/request.go
```diff
@@ -35,6 +35,11 @@ const (
 	RequestStateError RequestState = "error"
 )
 
+// IsRequestStateTerminal returns true if the state represents a final, irreversible state (landed or error).
+func IsRequestStateTerminal(s RequestState) bool {
+	return s == RequestStateLanded || s == RequestStateError
```
behinddwalls, line 40: do we also want to extend it to "cancelled"

entity/request_log.go
```diff
@@ -0,0 +1,62 @@
...
+type RequestLog struct {
+	// RequestID is the ID of the request this log entry belongs to. References entity.Request.ID.
+	RequestID string `json:"request_id"`
+	// TimestampMs is the time this log entry was created, in milliseconds since Unix epoch.
+	TimestampMs int64 `json:"timestamp_ms"`
+	// Status is the request status at the time this log entry was created. It does not have to correspond to the request status. For example, it may contain intermediate statuses like "validated" or "processing".
+	Status string `json:"status"`
+	// RequestVersion is the version of the request at the time this log entry was created.
+	// Zero if the version is not available.
+	RequestVersion int32 `json:"request_version"`
```
behinddwalls, line 20: I assume this is for optimistic concurrency, do we want to rename `request_log_version`, seems confusing as it feels like different version of request itself

core/request/request.go
```diff
@@ -0,0 +1,76 @@
...
+	// The algorithm:
+	// 1. If any record has a terminal status (landed, error) AND a version (RequestVersion > 0),
+	//    pick the one with the highest version. Timestamp breaks ties between equal versions, even though it should not happen.
+	// 2. Otherwise, fall back to the record with the largest timestamp.
+
+	var bestTerminal *entity.RequestLog
+	var bestLatest *entity.RequestLog
+
+	for i := range logs {
+		// iterate over all log records, storage contract guarantees that the records are ordered by timestamp ascending.
```
behinddwalls, line 48: wondering if we can make this order explict in the contract, likely by taking another argument to `store.List` API or renaming it

core/request/request.go
```diff
@@ -0,0 +1,76 @@
...
+				(log.RequestVersion == bestTerminal.RequestVersion && log.TimestampMs > bestTerminal.TimestampMs) {
+				bestTerminal = log
+			}
+		}
+	}
+
+	winner := bestLatest
+	if bestTerminal != nil {
+		winner = bestTerminal
+	}
```
behinddwalls, line 69: this is probably a separate concern but when we batch and batch fails due to some request, we de-batch it, re-queue with the same requestID (or new)? i guess we want to keep the same requestID for status reporting purpose to the end user? This will not be true if mark the original failure as `error` and re-queue or unless we have different state like `requeue-on-error` or something like that.

extension/storage/mysql/schema/request_log.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE IF NOT EXISTS request_log (
+    request_id VARCHAR(255) NOT NULL,
+    timestamp_ms BIGINT NOT NULL,
+    salt BIGINT NOT NULL,
+    status VARCHAR(64) NOT NULL,
+    request_version INT NOT NULL,
+    last_error TEXT NOT NULL,
+    metadata JSON NOT NULL,
+    PRIMARY KEY (request_id, timestamp_ms, salt)
```
behinddwalls, line 9: do we need a separate index on `request_id` and another one on `request_id, timestamp_ms` or create a reverse lookup table?

entity/request.go
sbalabanov: Yes, but I am waiting for the Request true states to be updated in a separate change.

entity/request_log.go
sbalabanov: It is because it is not for optimistic concurrency. It is a copy of a version from a Request entity. Which is what comment says on top, may be need better description.
Log records may come in different order with timestamps skewed, this field is for the reconciliation of a final status.

behinddwalls: I see...i got confused...so i think that makes sense then. might need to say in the comment

core/request/request.go
sbalabanov: `store.List` interface has a comment for it, same as we contract returning errors etc. We are limited by the language what we can do here. Are you suggesting to rename the function to smth like `ListInChronologicalOrder` ?

core/request/request.go
sbalabanov: I think we requeue with the same ID. So the Request entity does not enter a terminal state, and request log will turn from i.e. "merging" back to "batching".

extension/storage/mysql/schema/request_log.sql
sbalabanov: we do not. As long as we have a filter on `request_id` in a query, the optimizer will be able to use the main PK index.
The way how we do it in `List` is even better as we read in the order of the index (by sorting on secondary index fields), so it can just read the index straight. Very fast.

core/request/request.go
behinddwalls: Yes..something along those lines

core/request/request.go
behinddwalls: Yes. I think we may need a state which is not terminal in that case, we can do that separately but this answers my question.

extension/storage/mysql/schema/request_log.sql
behinddwalls: NIce..

behinddwalls (approved):

PR #119, opened 2026-03-04
feat(controller) Persist batch and batch dependent entities in store

manjari25:
## Summary
feat(controller) Persist batch and batch dependent entities in store

## What?
* Update the batch store with newly created batch (request maps 1:1 to batch)
* Use newly added core/metrics pkg to emit metrics
* Add a new batch state ("ready") to indicate success of all store operations in this controller

## Why?
Enables sequential batch ordering per queue so that downstream stages (score, speculate, merge) can process batches in the correct dependency order.


behinddwalls (approved):

orchestrator/controller/batch/batch.go
```diff
@@ -83,15 +85,16 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) er
...
+	counterOp.Complete(err)
 	if err != nil {
 		c.logger.Errorw("failed to generate batch ID",
 			"request_id", request.ID,
 			"queue", request.Queue,
 			"error", err,
 		)
-		c.metricsScope.Counter("counter_errors").Inc(1)
-		return fmt.Errorf("failed to generate batch ID for queue=%s: %w", request.Queue, err)
+		return errs.NewRetryableError(fmt.Errorf("failed to generate batch ID for queue=%s: %w", request.Queue, err))
```
sbalabanov, line 97: plz do not default to retryables

sbalabanov: see other review for details

orchestrator/controller/batch/batch.go
```diff
@@ -83,15 +85,16 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) er
 	)
 
 	// Generate a globally unique batch ID.
+	counterOp := metrics.Begin(c.metricsScope, "counter_next")
```
sbalabanov, line 88: counter has its one metrics, so this is probably redundant?

orchestrator/controller/batch/batch.go
```diff
@@ -102,20 +105,37 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) er
 		Version:  1,
 	}
 
+	// Persist batch to storage.
+	// ErrAlreadyExists should never happen since batch IDs are generated from a unique counter.
+	batchCreateOp := metrics.Begin(c.metricsScope, "batch_store_create")
```
sbalabanov, line 110: ditto

orchestrator/controller/batch/batch.go
```diff
@@ -102,20 +105,37 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) er
...
+	if err := c.store.GetBatchStore().Create(ctx, batch); err != nil {
+		batchCreateOp.Complete(err)
+		c.logger.Errorw("failed to create batch in storage",
```
sbalabanov, line 113: as long as we surface error up, avoid logging as it would most likely result in double logging.
Logging should be done at where the error chain terminates, typically the highest function in the call stack

orchestrator/controller/batch/batch.go
```diff
@@ -102,20 +105,37 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) er
...
+			"batch_id", batch.ID,
+			"error", err,
+		)
+		if errors.Is(err, storage.ErrAlreadyExists) {
```
sbalabanov, line 117: we do not expect it to ever happen if counter is correct, do not need special processing to this error, just fire it up

orchestrator/controller/batch/batch.go
```diff
@@ -102,20 +105,37 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) er
 		Version:  1,
 	}
 
+	// Persist batch to storage.
+	// ErrAlreadyExists should never happen since batch IDs are generated from a unique counter.
+	batchCreateOp := metrics.Begin(c.metricsScope, "batch_store_create")
+	if err := c.store.GetBatchStore().Create(ctx, batch); err != nil {
+		batchCreateOp.Complete(err)
```
sbalabanov, line 112: adhoc approach to scope management is fragile
what if some code path below won't return (could be another code change that makes it such). `Complete` would be called twice.
Prefer `defer` or another function wrapper to scope the measured block into.

sbalabanov (changes requested):

orchestrator/controller/batch/batch.go
behinddwalls: For all QueueControllers, i guess we just log at the consumer.go and leave it as is. For API controllers, we just log at the handlers which would be implemented in go-code?

So i think we should probably never log anywhere unless we are swallowing some errors (ideally we never need to)

PR #120, opened 2026-03-04
feat(controller) Conclude controller updates request state to reflect batch outcome

manjari25:
## Summary
feat(controller) Conclude controller updates request state to reflect batch outcome

## What?
* Inspect the request IDs within a batch
* Map batch state to equivalent request state
* Update request state in the request store
* Use newly created metrics/core pkg to track metrics

## Why?
These changes mark the request "concluded" within the SQ system.

### DIFF
--- a/orchestrator/controller/conclude/conclude.go
+++ b/orchestrator/controller/conclude/conclude.go
@@ -6,7 +6,9 @@ import (
 
 	"github.com/uber-go/tally/v4"
 	"github.com/uber/submitqueue/core/consumer"
+	"github.com/uber/submitqueue/core/errs"
 	"github.com/uber/submitqueue/entity"
+	"github.com/uber/submitqueue/extension/storage"
 	"go.uber.org/zap"
 )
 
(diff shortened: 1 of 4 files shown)


orchestrator/controller/conclude/conclude.go
```diff
@@ -96,3 +143,15 @@ func (c *Controller) TopicKey() consumer.TopicKey {
...
 	return c.consumerGroup
 }
+
+// batchStateToRequestState maps a terminal batch state to the corresponding request state.
+func batchStateToRequestState(state entity.BatchState) (entity.RequestState, error) {
+	switch state {
+	case entity.BatchStateSucceeded:
+		return entity.RequestStateLanded, nil
+	case entity.BatchStateFailed:
+		return entity.RequestStateError, nil
```
behinddwalls, line 153: i think when we think of re-queing, we need to re-think about this part

manjari25: Yes I left a TODO for this. For the MVP without cancel workflow; this is what the conclude should look like. When cancel workflow is added; this section can be updated.

behinddwalls (approved):

PR #121, opened 2026-03-05
feat(controller) Score controller scores batches using heuristic scorer

manjari25:
## Summary
feat(controller) Score controller scores batches using heuristic scorer

## What?
* Add concept of Score to batches
  * Add Score field to batch entity 
  * Batch store schema change
  * Update existing Batch store methods to account for new schema
  * Add new UpdateScoreAndState method
* Add heuristic scorer to orchestrator service
* Use this scorer in the batch scorer controller to score a batch
* Once scored, update the batch store with the score and state
* Use newly added core/metrics pkg for metric emission

## Why?
* Having a score associated with a batch is needed so that in the speculation stage; speculation path scores can be computed based on this number. 
* A speculation path score will be computed based on the success/failure score of all batches in the path.


entity/batch.go
```diff
@@ -69,6 +71,14 @@ func (b Batch) ToBytes() ([]byte, error) {
 	return json.Marshal(b)
 }
 
+// WithScoreAndState returns a new Batch with the given score and state, incrementing the version.
+func (b Batch) WithScoreAndState(score float32, state BatchState) Batch {
+	b.Score = score
+	b.State = state
+	b.Version++
```
sbalabanov, line 80: why need to increment the version? It should only be incremented after a successful atomic persistence (i.e. a write to the database).

example/server/orchestrator/main.go
```diff
@@ -442,6 +450,18 @@ func newMergeChecker(logger *zap.Logger, scope tally.Scope) mergechecker.MergeCh
 	})
 }
 
+func newScorer(scope tally.Scope) scorer.Scorer {
+	return heuristicscorer.New(
+		[]heuristicscorer.Bucket{
+			{Min: 0, Max: 100, Score: 1.0},
```
sbalabanov, line 459: add "TODO" for the real one?

extension/storage/mysql/schema/batch.sql
```diff
@@ -3,6 +3,7 @@ CREATE TABLE IF NOT EXISTS batch (
     queue VARCHAR(255) NOT NULL,
     contains JSON NOT NULL,
     dependencies JSON NOT NULL,
+    score FLOAT NOT NULL DEFAULT 0,
```
sbalabanov: avoid db defaults, leave it to application

orchestrator/controller/score/score.go
```diff
@@ -74,18 +85,87 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) er
...
+			metrics.NamedCounter(c.metricsScope, "process", "request_not_found_errors", 1)
+			return fmt.Errorf("request %s not found: %w", batch.Contains[0], err)
+		}
+		c.logger.Errorw("failed to get request",
+			"batch_id", batch.ID,
+			"request_id", batch.Contains[0],
+			"error", err,
+		)
+		metrics.NamedCounter(c.metricsScope, "process", "storage_errors", 1)
+		return errs.NewRetryableError(fmt.Errorf("failed to get request %s: %w", batch.Contains[0], err))
```
sbalabanov, line 117: Only explicitly known errors should be converted to retryable.
For example, if database connection fails, it is retryable.
If database schema mismatch, it is not retryable.
Being not retryable is a default to avoid retry storms.
For now we can just bubble up the error unchanged with a TODO to make a classification later.

orchestrator/controller/score/score.go
```diff
@@ -74,18 +85,87 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) er
...
+
+	c.logger.Infow("scored batch",
+		"batch_id", batch.ID,
+		"score", batchScore,
+	)
+
+	// Update batch store with score and transition state to speculating
+	if err := c.store.GetBatchStore().UpdateScoreAndState(ctx, batch.ID, batch.Version, batchScore, entity.BatchStateSpeculating); err != nil {
+		if errors.Is(err, storage.ErrVersionMismatch) {
+			c.logger.Errorw("version mismatch updating batch score",
```
sbalabanov, line 142: If it is expected by some workflow path, should be warning at most

orchestrator/controller/score/score.go
```diff
@@ -74,18 +85,87 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) er
...
+	batchScore := float32(score)
+
+	c.logger.Infow("scored batch",
+		"batch_id", batch.ID,
+		"score", batchScore,
+	)
+
+	// Update batch store with score and transition state to speculating
+	if err := c.store.GetBatchStore().UpdateScoreAndState(ctx, batch.ID, batch.Version, batchScore, entity.BatchStateSpeculating); err != nil {
+		if errors.Is(err, storage.ErrVersionMismatch) {
```
sbalabanov, line 141: nit: we probably should create wrappers for it in `storage.go`

entity/batch.go
sbalabanov: Now I see where it is used (after the update), the comment could reflect that.

orchestrator/controller/score/score.go
```diff
@@ -74,18 +85,87 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) er
...
+		metrics.NamedCounter(c.metricsScope, "process", "batch_store_errors", 1)
+		return errs.NewRetryableError(fmt.Errorf("failed to update batch %s score: %w", batch.ID, err))
+	}
+
+	// Create new batch with updated state and version to reflect the store update
+	scored := batch.WithScoreAndState(batchScore, entity.BatchStateSpeculating)
 
 	// Publish to speculate topic
-	if err := c.publish(ctx, consumer.TopicKeySpeculate, batch); err != nil {
+	if err := c.publish(ctx, consumer.TopicKeySpeculate, scored); err != nil {
```
sbalabanov, line 162: I think we publish batch_id, not the whole object

sbalabanov (changes requested):

orchestrator/controller/score/score.go
manjari25: We do - that is abstracted in `func (c *Controller) publish(ctx context.Context, key consumer.TopicKey, batch entity.Batch) error` function below

example/server/orchestrator/main.go
manjari25: Done

extension/storage/mysql/schema/batch.sql
manjari25: Removed

entity/batch.go
manjari25: Added a comment.

orchestrator/controller/score/score.go
manjari25: This isn't only schema mismatch though right? Get from the store could fail for any other reason as well. Do we not want to retry in this case?

orchestrator/controller/score/score.go
manjari25: Will work on this in a different PR (so we can get this one out before the grand mock refactor).

orchestrator/controller/score/score.go
manjari25: I don't think we expect this by any workflow path...

orchestrator/controller/score/score.go
sbalabanov: Our default should be to not retry. Retryable errors need to be explicitly opted in as such.
For now we can put TODO here, return generic err and classify later, I envision there may be a general helper framework needed to transform i.e. mysql errors to retryable or not.

orchestrator/controller/score/score.go
sbalabanov: In this case just bubble the error up without processing?

orchestrator/controller/score/score.go
```diff
@@ -88,18 +85,87 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) er
...
+		if storage.IsNotFound(err) {
+			c.logger.Errorw("request not found",
+				"batch_id", batch.ID,
+				"request_id", batch.Contains[0],
+				"error", err,
+			)
+			metrics.NamedCounter(c.metricsScope, "process", "request_not_found_errors", 1)
+			return fmt.Errorf("request %s not found: %w", batch.Contains[0], err)
+		}
+		c.logger.Errorw("failed to get request",
```
sbalabanov, line 111: no need to log if bubble up

orchestrator/controller/score/score.go
```diff
@@ -88,18 +85,87 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) er
...
+		// TODO: multi-request batches will be supported later
+		metrics.NamedCounter(c.metricsScope, "process", "multi_request_batch_errors", 1)
+		return fmt.Errorf("batch %s contains %d requests, only single-request batches are supported", batch.ID, len(batch.Contains))
+	}
+
+	// Look up the request to get its Change
+	request, err := c.store.GetRequestStore().Get(ctx, batch.Contains[0])
+	if err != nil {
+		if storage.IsNotFound(err) {
+			c.logger.Errorw("request not found",
```
sbalabanov, line 103: no need to log if bubble up

orchestrator/controller/score/score.go
```diff
@@ -88,18 +85,87 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) er
...
+	if len(batch.Contains) > 1 {
+		// TODO: multi-request batches will be supported later
+		metrics.NamedCounter(c.metricsScope, "process", "multi_request_batch_errors", 1)
+		return fmt.Errorf("batch %s contains %d requests, only single-request batches are supported", batch.ID, len(batch.Contains))
+	}
+
+	// Look up the request to get its Change
+	request, err := c.store.GetRequestStore().Get(ctx, batch.Contains[0])
+	if err != nil {
+		if storage.IsNotFound(err) {
```
sbalabanov, line 102: do we need to process it separately? why not to just to bubble up? This is not an expected state, even when states are (eventually) inconsistent this should not be true.

orchestrator/controller/score/score.go
```diff
@@ -88,18 +85,87 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) er
...
+			"error", err,
+		)
+		metrics.NamedCounter(c.metricsScope, "process", "storage_errors", 1)
+		return errs.NewRetryableError(fmt.Errorf("failed to get request %s: %w", batch.Contains[0], err))
+	}
+
+	// Score the change
+	score, err := c.scorer.Score(ctx, request.Change)
+	if err != nil {
+		c.logger.Errorw("failed to score change",
```
sbalabanov, line 123: ditto
plz check other usages

orchestrator/controller/score/score.go
```diff
@@ -88,18 +85,87 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) er
...
+			)
+			metrics.NamedCounter(c.metricsScope, "process", "version_mismatch_errors", 1)
+			return fmt.Errorf("version mismatch updating batch %s: %w", batch.ID, err)
+		}
+		c.logger.Errorw("failed to update batch score",
+			"batch_id", batch.ID,
+			"error", err,
+		)
+		metrics.NamedCounter(c.metricsScope, "process", "batch_store_errors", 1)
+		return errs.NewRetryableError(fmt.Errorf("failed to update batch %s score: %w", batch.ID, err))
```
sbalabanov, line 155: ditto, plz check other usages

sbalabanov (changes requested):

entity/batch.go
behinddwalls: i am wondering if this should be in entity at all? 

to me it seems like we probably need this to be part of contract at storage layer itself which accepts, `entity, oldVersion, newVresion`... I am not fan of implicit things as they hide away and lead to confusion/issues down the line..

Thoughts?

PR #125, opened 2026-03-06
feat(requestlog): Request Log Event Bus, States and Statuses

sbalabanov:
Dedicated statuses for Request Log.
Separation between statuses and states.
Event bus via queue for request logging and a new controller to save to the database.
Gateway services use Request Log directly.
Orchestrator services use Request Log via a message bus.
A lot of statuses are created, but none are hooked up except for a couple exemplary ones. Up for extension.

### DIFF
--- a/entity/request_log.go
+++ b/entity/request_log.go
@@ -19,6 +19,64 @@ import (
 	"time"
 )
 
+// RequestLogStatus defines the possible status of a request. Status is customer-friendly and can be displayed to the user.
+// It is different from the request state, which is internal and used to implement a state machine. Request statuses can be
+// generally added freely by the system without breaking the state machine.
+// Some statuses correspond to the request state, in which case they should be supplemented with the request state version to be used for reconciliation.
+// Other statuses are purely informational and can be added freely.
+// Every status may be accompanied by a last error message and free-formmetadata in the Request Log. It will only be used for display or debugging purposes.
+type RequestStatus string
+
+const (
+	// RequestStatusUnknown is the unknown sentinel status. It is set by default when the structure is initialized. It should never be seen in the system.
+	RequestStatusUnknown RequestStatus = ""
+
+	// RequestStatusAccepted indicates that the request has been accepted by the system. Typically a gateway service will set this status when the land request is received and persisted to the logging database.
+	RequestStatusAccepted RequestStatus = "accepted"
+
+	// RequestStatusNew is the initial status of a request. It corresponds to the RequestStateNew state and typically set by the orchestrator service when the request is received and persisted to the operating database.
+	RequestStatusNew RequestStatus = "new"
+
+	// RequestStatusValidating indicates that the request is currently being validated (e.g., duplicate check, merge check, etc.).
+	RequestStatusValidating RequestStatus = "validating"
+
+	// RequestStatusValidated indicates that the request has been validated (duplicate check, merge check etc.) successfully. It corresponds to the RequestStateValidated state.
+	RequestStatusValidated RequestStatus = "validated"
+
+	// RequestStatusBatching indicates that the request is waiting to be included in a batch.
+	RequestStatusBatching RequestStatus = "batching"
+
+	// RequestStatusBatched indicates that the request has been included in a new batch and will be sent to speculation.
+	RequestStatusBatched RequestStatus = "batched"
+
+	// RequestStatusSpeculating indicates that the request is currently being speculated (e.g., speculative merge/rebase, etc.).
+	RequestStatusSpeculating RequestStatus = "speculating"
+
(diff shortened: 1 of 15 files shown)


orchestrator/controller/request/request.go
```diff
@@ -153,6 +160,32 @@ func (c *Controller) publish(ctx context.Context, key consumer.TopicKey, request
...
 }
 
+// publishLog publishes a request log entry to the log topic for async persistence.
+func (c *Controller) publishLog(ctx context.Context, logEntry entity.RequestLog, partitionKey string) error {
+	payload, err := logEntry.ToBytes()
+	if err != nil {
+		return fmt.Errorf("failed to serialize request log: %w", err)
+	}
+
+	msg := entityqueue.NewMessage(logEntry.RequestID, payload, partitionKey, nil)
```
sbalabanov, line 172: @behinddwalls what is a good partition key? we likely want a queue ordered chronologically per request id.

behinddwalls: I think we can just partition by requestID itself, we primarily need events ordered for the request. The other thing is that if we end up also sending events for batchIDs as well, then we can use the same pattern.

entity/request_log.go
```diff
@@ -19,6 +19,64 @@ import (
...
+	RequestStatusSpeculating RequestStatus = "speculating"
+
+	// RequestStatusSpeculated indicates that the request has been successfully speculated and is ready to be validated via a build system.
+	RequestStatusSpeculated RequestStatus = "speculated"
+
+	// RequestStatusBuilding indicates that the request is currently being built (e.g., CI/CD system is building the change on top of the speculation path).
+	RequestStatusBuilding RequestStatus = "building"
+
+	// RequestStatusBuilt indicates that the request has finished the build step successfully and can move to the next phase, either wait for other requests to finish or move to the land phase.
+	RequestStatusBuilt RequestStatus = "built"
```
behinddwalls, line 62: thinking out loud here,  request can transition through these few times,

speculated -> building, again re-speculating if other builds complete or get scheduled due to some tree pruning etc.. and when we say built, when do we say that, when all the paths have been built or at least one path has been built... (or maybe when we have reached to the a future we can land)...

there might be cases where between speculation paths being built we re-speculate few times?

entity/request_log.go
```diff
@@ -19,6 +19,64 @@ import (
...
+
+	// RequestStatusWaitingPath indicates that the request is waiting for other preceiding request in the same speculation path to finish.
+	RequestStatusWaitingPath RequestStatus = "waitingpath"
+
+	// RequestStatusLanding indicates that the request is actively being landed (e.g., source control operation is in progress to push the change to the target branch).
+	RequestStatusLanding RequestStatus = "landing"
+
+	// RequestStatusProcessing is the status of a request that is being processed. It corresponds to the RequestStateProcessing state.
+	RequestStatusProcessing RequestStatus = "processing"
+
```
behinddwalls, line 72: do we need this? thinking where would we use it.

entity/request_log.go
```diff
@@ -19,6 +19,64 @@ import (
...
+const (
+	// RequestStatusUnknown is the unknown sentinel status. It is set by default when the structure is initialized. It should never be seen in the system.
+	RequestStatusUnknown RequestStatus = ""
+
+	// RequestStatusAccepted indicates that the request has been accepted by the system. Typically a gateway service will set this status when the land request is received and persisted to the logging database.
+	RequestStatusAccepted RequestStatus = "accepted"
+
+	// RequestStatusNew is the initial status of a request. It corresponds to the RequestStateNew state and typically set by the orchestrator service when the request is received and persisted to the operating database.
+	RequestStatusNew RequestStatus = "new"
```
behinddwalls, line 38: could it be `starting`; new also works though

entity/request_log.go
sbalabanov: The actual statuses are not well thought out yet, there are all up for considerations. General idea was to allow multiple records with the same status, accompanying with the appropriate metadata, for example when a single build completes, the record is created.
So, multiple records with "building" or "build scheduled" and a single "built" when all are done. With that said, it can enter a build loop again due to respeculation, but then another record would be created indicating respeculation.

entity/request_log.go
sbalabanov: Copied from the state. If we do not need this state, it would be removed.

entity/request_log.go
sbalabanov: starting is better I think, because it is confusing to see "new" followed after "accepted". Does not look too "new". :)
But this is again a copy from the state, and it has to match for terminal state detection for proper reconciliation.

behinddwalls (approved):

entity/request_log.go
behinddwalls: Sounds good we can fix them later

PR #126, opened 2026-03-07
refactor(queue/mysql): immutable log, delivery state, heartbeat store, fair leasing

behinddwalls:
## Summary
- Add delivery state store for tracking per-consumer message state separately from the immutable log
- Add subscriber heartbeat store for liveness detection and fair partition leasing
- Refactor message store to append-only immutable log with contiguous watermark tracking
- Refactor partition lease store with heartbeat-based fair leasing
- Add batch query support for offset and delivery state operations
- Add GC throttling and dead code removal
- Fix make fmt to skip read-only pkg/mod directory


extension/queue/mysql/partition_lease_store.go
```diff
@@ -316,14 +323,43 @@ func (s *sqlpartitionLeaseStore) DiscoverAndAcquirePartitions(ctx context.Contex
...
 		logTopic, topic,
 		"count", len(partitions),
 	)
 
+	// Query owned partitions once before the loop to avoid N+1 queries
+	ownedCount := 0
+	if maxPartitions > 0 {
+		owned, err := s.GetLeasedPartitions(ctx, topic, subscriberName, consumerGroup)
+		if err != nil {
+			s.logger.Warnw("failed to get owned partitions for cap check",
```
sbalabanov: As the error is surfaced up, I believe it would be logged somewhere else? In which case this logging is redundant.

extension/queue/mysql/subscriber.go
```diff
@@ -812,6 +827,132 @@ func (s *subscriber) releaseAllLeases(ctx context.Context, sub *subscription) {
 	}
 }
 
+// sendHeartbeat sends a heartbeat for this subscriber
+func (s *subscriber) sendHeartbeat(ctx context.Context, sub *subscription) {
```
sbalabanov: what does a context control? If cancellation, then the function does not seem to return an error.
It's better to have the context contract explained.

extension/queue/mysql/subscriber.go
```diff
@@ -812,6 +827,132 @@ func (s *subscriber) releaseAllLeases(ctx context.Context, sub *subscription) {
...
+//
+// The formula is ceil(totalPartitions / activeSubscribers), which ensures all
+// partitions are assigned even when they don't divide evenly. At most one
+// subscriber will hold one extra partition.
+//
+// Returns 0 (unlimited) when fair share cannot be computed (error or single subscriber).
+func (s *subscriber) computeMaxPartitions(ctx context.Context, sub *subscription) int {
+	maxPart, _, err := s.fairShareCap(ctx, sub)
+	if err != nil {
+		return 0
```
sbalabanov: what if the error is infra? It is swallowed without even a warning.
If it is a graceful fallback, needs proper monitoring at least.

extension/queue/mysql/subscriber.go
```diff
@@ -812,6 +827,132 @@ func (s *subscriber) releaseAllLeases(ctx context.Context, sub *subscription) {
...
+	return maxPart
+}
+
+// rebalance checks if this subscriber holds more partitions than its fair share
+// and releases extras so other subscribers can pick them up.
+func (s *subscriber) rebalance(ctx context.Context, sub *subscription) {
+	cfg := sub.config
+
+	maxPart, owned, err := s.fairShareCap(ctx, sub)
+	if err != nil || maxPart == 0 || len(owned) <= maxPart {
```
sbalabanov: ditto

extension/queue/mysql/subscriber.go
```diff
@@ -812,6 +827,132 @@ func (s *subscriber) releaseAllLeases(ctx context.Context, sub *subscription) {
...
+
+	// Count all known partitions as the union of owned + discovered.
+	// Using max(owned, discovered) would undercount when some partitions
+	// have leases but no messages, or vice versa.
+	partitionSet := make(map[string]struct{}, len(owned))
+	for _, pk := range owned {
+		partitionSet[pk] = struct{}{}
+	}
+	allPartitions, err := s.leaseStore.DiscoverPartitions(ctx, sub.topic)
+	if err == nil {
```
sbalabanov: err != nil is not processed

sbalabanov:
I think overall error processing and expectations on reliability of the heartbeats need to be better articulated. Both contracting the interface with comments and readme.

sbalabanov (changes requested):

extension/queue/mysql/partition_lease_store.go
behinddwalls: Let me check the store logging, i thought i removed them all...

extension/queue/mysql/subscriber.go
behinddwalls: i will add that.

extension/queue/mysql/subscriber.go
behinddwalls: I think i need to revisit all in this queue and consumer...i will do that as a separate PR

extension/queue/mysql/subscriber.go
behinddwalls: updated the documentation, in this case, it's just controlling DB call...if it fails, we could be a slighly wrong about total subscribers available but nothing else. I am gonna add metrics here for monitoring

extension/queue/mysql/subscriber.go
behinddwalls: Added metrics here for gracefull fallback

behinddwalls:
> I think overall error processing and expectations on reliability of the heartbeats need to be better articulated. Both contracting the interface with comments and readme.

I need to update those, I will do them in a separate PR.

extension/queue/mysql/subscriber.go
behinddwalls: added log and metrics

PR #138, opened 2026-03-11
feat(batch): Reimplement batch controller for consistency

sbalabanov:
## Summary
BatchDependent is a secondary index stucture to Batch entity, make it
fully eventually consistent:
- Give it more comments
- Always create with batch creation so next batch won't have to guess
- Remove metadata, it won't be used, at least not for state tracking as it
  gets stale
- Ensure order of operations - dependents first, empty dependent entry
  second, batch itself to follow, and then queue
- Additional comments added explaining the further stale state
  reconciliation logic

## Stack
1. #137
1. @ #138

### DIFF
--- a/entity/batch.go
+++ b/entity/batch.go
@@ -51,28 +51,39 @@ func (s BatchState) IsTerminal() bool {
 type Batch struct {
 	// ID is the globally unique identifier for the batch. Format: "<queue>/batch/<counter_value>".
 	ID string
+
 	// Queue is the name of the queue processing the land request. Queue name is defined in the configuration and should be unique within the system.
 	Queue string
+
 	// Contains is a list of land request IDs that are part of this batch.
 	// Request IDs will always be part of the same queue.
 	//
 	// For e.g. - [queueA/1, queueA/2, queueA/3].
 	//
 	Contains []string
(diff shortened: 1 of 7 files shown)


entity/batch.go
```diff
@@ -51,28 +51,39 @@ func (s BatchState) IsTerminal() bool {
...
-	// such that - queueA/batch/2 and queueA/batch/3 depend on queueA/batch/1
+	// Example: consider batches - queueA/batch/1, queueA/batch/2, queueA/batch/3
+	// such that - queueA/batch/2 and queueA/batch/3 have overlapping targets with requests in queueA/batch/1, but queueA/batch/2 and queueA/batch/3 do not have overlapping targets with each other.
 	//
 	// In this case, the Dependencies field for -
 	// - queueA/batch/1 will be empty
 	// - queueA/batch/2 will contain queueA/batch/1
 	// - queueA/batch/3 will contain queueA/batch/1
-	//
-	Dependencies []map[string]interface{}
```
sbalabanov, line 73: we should not have interfaces in entity objects, but pure serializable DTOs

behinddwalls (approved):

PR #141, opened 2026-03-12
implement change provider

rashmi-prithyani:
## Summary
The orchestrator can now fetch rich metadata about a code change directly from GitHub — including the author, changed files (with line-level add/delete/modify counts and patches), and the head commit SHA — at validation time. This gives downstream controllers the information they need to make decisions based on what a PR actually touches, rather than just its identity.

  The implementation supports stacked PRs by returning one ChangeInfo per URI in a Change, and handles large PRs transparently via GraphQL pagination. Authentication and endpoint configuration are injected via environment variables, making it compatible with both github.com and GitHub Enterprise.

## Test Plan
Unit tests cover Client URL construction, GraphQL pagination, URI validation, and provider behavior with a mock  HTTP server.


core/httpclient/transport.go
```diff
@@ -0,0 +1,80 @@
...
+	next := t.Next
+	if next == nil {
+		next = http.DefaultTransport
+	}
+	return next.RoundTrip(newReq)
+}
+
+// BearerTransport is an http.RoundTripper that adds a Bearer token
+// Authorization header to every request.
+type BearerTransport struct {
```
behinddwalls: lets remove this, you can do this using oauth2 lib way simpler than re-creating it

rashmi-prithyani: Done

core/httpclient/transport.go
```diff
@@ -0,0 +1,62 @@
...
+		next = http.DefaultTransport
+	}
+	return next.RoundTrip(newReq)
+}
+
+// NewClient builds an *http.Client with BaseURLTransport and optionally
+// oauth2 bearer auth configured. The transport chain is:
+//
+//	oauth2.Transport (if token provided) → BaseURLTransport → DefaultTransport
+func NewClient(rawBaseURL, token string, timeout time.Duration) (*http.Client, error) {
```
behinddwalls: you can remove token from here as well, even timeout.. the only goal of this to support a BaseURL so that should be it for this specific transport..

core/httpclient/transport.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	var transport http.RoundTripper = &BaseURLTransport{
+		BaseURL: u,
+		Next:    http.DefaultTransport,
+	}
+	if token != "" {
+		ts := oauth2.StaticTokenSource(&oauth2.Token{AccessToken: token})
+		transport = &oauth2.Transport{Source: ts, Base: transport}
+	}
```
behinddwalls: you don't need..one can chain it outside..

core/httpclient/transport.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+// NewClient builds an *http.Client with BaseURLTransport and optionally
+// oauth2 bearer auth configured. The transport chain is:
+//
+//	oauth2.Transport (if token provided) → BaseURLTransport → DefaultTransport
+func NewClient(rawBaseURL, token string, timeout time.Duration) (*http.Client, error) {
+	u, err := url.Parse(rawBaseURL)
+	if err != nil {
+		return nil, err
```
behinddwalls, line 45: should we add some additional message in err via `fmt.Errorf(...)`

behinddwalls (approved):

core/httpclient/transport.go
rashmi-prithyani: url.Parse already provides a descriptive message about what failed. 
```
  // e.g. → parse "://invalid": missing protocol scheme
```

core/httpclient/transport.go
rashmi-prithyani: ok, removed those and set them directly in the main.go file now.

PR #144, opened 2026-03-17
duplicate detection via change store

rashmi-prithyani:
## Summary
Stacked on top of #152 (change store extension).

Detects duplicate land requests by reading per-URI claims from the new change store. The `start` controller writes a claim row for each URI on the request; `validate` reads the change store and rejects requests whose URIs overlap with another in-flight request in the same queue.

- **start.go** — after persisting the request, claim each URI in the change store. `INSERT IGNORE` makes this idempotent on queue redelivery (same `request_id` retry is a no-op).
- **validate.go** — query the change store for any row with an overlapping URI in the same queue (excluding self). For each unique candidate `request_id` returned, look up the request and skip terminal/orphan owners. If any live owner remains, reject as a user error naming the conflicting `request_id`.
- Two single-table queries per check, no cross-store SQL joins.
- Stop wrapping every storage error as `Retryable`; default is plain `fmt.Errorf` per the `core/errs` framework + #134.
- Wires `changeStore` into both controllers in `example/server/orchestrator/main.go`.

## Behavior

- **Overlap (not full-set equality):** the change store returns rows for any URI in the new request that matches a row from another in-flight request, so partial-overlap is rejected — addresses @sbalabanov's review comment.
- **Liveness:** terminal-state owners and orphans (whose request row is gone) are skipped, so stale claims don't block new requests indefinitely.
- **Retries:** `start` tolerates `ErrAlreadyExists` from request creation as a same-id replay; the change store's `INSERT IGNORE` is idempotent for the same reason.

## Test Plan
- Unit tests for `start` and `validate` cover happy path, overlap-with-live, overlap-with-terminal-skipped, overlap-with-orphan-skipped, multi-URI-same-owner deduped, and infra-error propagation.
- Existing integration tests on `extension/storage/mysql` continue to pass after `GetActiveByQueue` removal.
- Change store integration test (in #152) exercises the MySQL impl against a real DB.


extension/storage/mysql/request_store.go
```diff
@@ -95,6 +95,42 @@ func (r *requestStore) Create(ctx context.Context, request entity.Request) (retE
...
+	op := metrics.Begin(r.scope, "find_by_queue_and_change_uris")
+	defer func() { op.Complete(retErr) }()
+
+	urisJSON, err := json.Marshal(uris)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal change URIs for duplicate check queue=%s: %w", queue, err)
+	}
+
+	rows, err := r.db.QueryContext(ctx,
+		"SELECT id FROM request WHERE queue = ? AND id != ? AND JSON_LENGTH(change_uri) = ? AND JSON_CONTAINS(change_uri, CAST(? AS JSON))",
```
rashmi-prithyani: This query is not very efficient at the moment as it does a full table scan - [slack thread on this ](https://uber.slack.com/archives/C09N26TQLLA/p1773712982804059)

orchestrator/controller/validate/validate_test.go
```diff
@@ -259,3 +260,105 @@ func TestController_Process_MergeCheckError(t *testing.T) {
...
+			duplicates:    []string{"test-queue/100"},
+			wantErr:       true,
+			wantUserError: true,
+		},
+		{
+			// JSON_CONTAINS is order-insensitive, so a reordered stack matches. We accept
+			// this: reordering a stack changes commit hashes in practice (rebase produces
+			// new SHAs), so URIs will never actually collide. If they somehow did, rejecting
+			// is the safe behavior.
+			name: "reordered stack treated as duplicate (ordering not checked)",
```
rashmi-prithyani: Would like to verify if this assumption is correct ?

orchestrator/controller/validate/validate.go
```diff
@@ -103,6 +103,22 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) (r
 		"partition_key", msg.PartitionKey,
 	)
 
+	// Duplicate check: reject requests that share the exact same set of change URIs in the same queue.
+	duplicates, err := c.store.GetRequestStore().FindByQueueAndChangeURIs(ctx, request.Queue, request.Change.URIs, request.ID)
+	if err != nil {
+		coremetrics.NamedCounter(c.metricsScope, "process", "duplicate_check_errors", 1)
+		return errs.NewRetryableError(fmt.Errorf("failed to check for duplicate requests for %s: %w", request.ID, err))
+	}
+	if len(duplicates) > 0 {
```
manjari25: Check for == 1 specifically. It should never be more than that.

orchestrator/controller/validate/validate.go
```diff
@@ -103,6 +103,22 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) (r
...
+		c.logger.Infow("duplicate request detected",
+			"request_id", request.ID,
+			"queue", request.Queue,
+			"duplicate_ids", duplicates,
+		)
+		coremetrics.NamedCounter(c.metricsScope, "process", "duplicate_requests", 1)
+		return errs.NewUserError(fmt.Errorf("request %s is a duplicate of existing requests %v", request.ID, duplicates))
```
manjari25: Update this based on above comment. If at all, there should only ever be 1 duplicate request found.

orchestrator/controller/validate/validate.go
```diff
@@ -103,6 +103,22 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) (r
 		"partition_key", msg.PartitionKey,
 	)
 
+	// Duplicate check: reject requests that share the exact same set of change URIs in the same queue.
+	duplicates, err := c.store.GetRequestStore().FindByQueueAndChangeURIs(ctx, request.Queue, request.Change.URIs, request.ID)
```
manjari25: Discussed offline; summarizing here. I think this logic should look as follows -
```
activeRequests, err := c.store.GetRequestStore().GetActiveRequestsByQueue(..., request.Queue) // might need an index on (queue, state); similar to batch table
// err handling
for _, r := range activeRequests {
  if r.ID == rid {
    continue; // rid is the current incoming request; skip comparing with itself
  }
  // check if request.URIs set has any overlap with r.URIs
  if duplicateReqID, ok := getIntersection(r.URIs, request.URIs); ok {
    // user error; return duplicateReqID
  }
  // duplicate check passes; more logic...
} 
```

orchestrator/controller/validate/validate_test.go
rashmi-prithyani: This assumption is correct. If someone reordered the stack, it would result in totally different shas and it would be a new request at that point.

orchestrator/controller/validate/validate.go
rashmi-prithyani: Updated the code to do the duplicate check in go instead of the db query now.

extension/storage/mysql/request_store.go
rashmi-prithyani: Changed the query to filter by queue and non terminal request state - also have secondary index on these fields. 

Now, we do the duplication check in go instead of the JSON_CONTAINS in the db

extension/storage/mysql/schema/request.sql
```diff
@@ -5,5 +5,6 @@ CREATE TABLE IF NOT EXISTS request (
     land_strategy VARCHAR(64) NOT NULL,
     state VARCHAR(64) NOT NULL,
     version INT NOT NULL,
-    PRIMARY KEY (id)
+    PRIMARY KEY (id),
+    INDEX idx_queue_state (queue, state)
```
sbalabanov: this table should whatsoever has no secondary indexes

orchestrator/controller/validate/validate.go
```diff
@@ -103,6 +103,28 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) (r
 		"partition_key", msg.PartitionKey,
 	)
 
+	// Duplicate check: fetch all active requests in the queue and compare URI sets in Go.
+	// Exclude terminal requests — only in-flight requests can be duplicates.
+	activeRequests, err := c.store.GetRequestStore().GetActiveByQueue(ctx, request.Queue)
+	if err != nil {
+		coremetrics.NamedCounter(c.metricsScope, "process", "duplicate_check_errors", 1)
+		return errs.NewRetryableError(fmt.Errorf("failed to fetch active requests for duplicate check %s: %w", request.ID, err))
```
sbalabanov: not every db error is ultimately retryable. See error framework and https://github.com/uber/submitqueue/pull/134 for details.

orchestrator/controller/validate/validate.go
```diff
@@ -103,6 +103,28 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) (r
 		"partition_key", msg.PartitionKey,
 	)
 
+	// Duplicate check: fetch all active requests in the queue and compare URI sets in Go.
+	// Exclude terminal requests — only in-flight requests can be duplicates.
+	activeRequests, err := c.store.GetRequestStore().GetActiveByQueue(ctx, request.Queue)
```
sbalabanov: we could do better than retrieving all of the in-flight requests then iterating them on a client

orchestrator/controller/validate/validate.go
```diff
@@ -187,6 +209,23 @@ func totalFiles(infos []changeprovider.ChangeInfo) int {
 	return total
 }
 
+// sameURIs returns true if a and b contain exactly the same set of URIs (order-insensitive).
```
sbalabanov: I think we should reject a request is any change specified in the request overlaps with any change already in flight, not just full sets

sbalabanov (changes requested):

behinddwalls (approved):

PR #148, opened 2026-04-29
feat(speculate): naive happy-path speculation in speculate controller

sbalabanov:
## Summary

Replaces the speculate controller's unconditional fan-out (always publishing to **both** `build` and `merge`) with a state-machine driver that advances a batch one step per invocation along a single happy-path speculation chain (`batch.Dependencies + [batch.ID]`).

Per call, based on `batch.State`:

| State | Action |
|---|---|
| `Created` / `Scored` | publish to `build`, transition to `Speculating` |
| `Speculating` + all deps `Succeeded` | publish to `merge`, transition to `Finalizing` |
| `Speculating`, deps still in flight | no-op (re-evaluated on next event) |
| `Speculating`, dep `Failed`/`Cancelled` | no-op + warn (TODO: replan) |
| `Finalizing` / terminal | no-op |

Inspired by the legacy Java `SpeculationTree`/`StrategyImpl` (under `dev-platform/cicd/submitqueue`). Deliberately omitted from this PR — these are TODOs for follow-ups:

- Score-driven prioritization (the legacy `NodeScorer` / `PrioritizerImpl`).
- 2^N tree of orderings with true/false branches per batch (`ExhaustiveEnumerator`).
- Failure replanning when a dependency fails.
- Conflict resolution and capacity limits.

### Implementation notes

- Publishes first, then CAS-updates state via the existing optimistic-locking `UpdateState`. A publish failure leaves the state unchanged for the retry to re-attempt cleanly.
- Speculation chain is currently only logged; the `build` controller does not yet receive it as part of the message (TODO: thread `SpeculationPathInfo` through to `build`/`buildsignal`).
- Re-triggering across batches still relies on the existing `merge → speculate` and `buildsignal → speculate` hops; downstream batches getting re-evaluated when an upstream batch concludes is a separate gap (not addressed here).

## Test plan

- [x] `go test ./orchestrator/controller/speculate/...` — passes
- [x] `go test ./orchestrator/controller/...` — passes (no regressions in adjacent controllers)
- [x] `go vet ./...` — clean
- [x] New table/state tests cover: start-from-Created, start-from-Scored, finalize-no-deps, finalize-all-deps-succeeded, waiting-on-deps, failed-dep halts, terminal/finalizing no-op, storage failures (primary + dep fetch), publish failures (start + finalize), bad payload.

### DIFF
--- a/orchestrator/controller/speculate/speculate.go
+++ b/orchestrator/controller/speculate/speculate.go
@@ -27,8 +27,23 @@ import (
 )
 
 // Controller handles speculate queue messages.
-// It consumes batches, performs speculation, and publishes to both build and merge stages.
-// Implements consumer.Controller interface for integration with the consumer.
+//
+// Naive happy-path algorithm: assume every in-flight build will pass and
+// treat batch.Dependencies + [batch.ID] as the single speculation chain.
+// Per invocation, the controller advances the batch one step in the
+// state machine:
+//
+//   - Created or Scored → publish to build, transition to Speculating.
+//   - Speculating       → if all deps are Succeeded, publish to merge and
+//     transition to Merging; otherwise no-op (or fail-fast if a dep is
+//     in a non-succeeding terminal state).
+//   - Merging           → no-op (owned by the merge controller).
+//   - Terminal          → re-fan-out to conclude for self-healing in case a
+//     prior publish was lost.
+//
+// The controller is re-triggered on every relevant downstream event
+// (buildsignal, merge), so each call simply re-evaluates the current
+// state and either advances or waits.
 type Controller struct {
 	logger        *zap.SugaredLogger
 	metricsScope  tally.Scope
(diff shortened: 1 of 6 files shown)


behinddwalls (approved):

orchestrator/controller/speculate/speculate.go
```diff
@@ -91,37 +100,119 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) er
...
+			"state", string(batch.State),
+		)
+		return nil
+	}
+}
+
+// startSpeculation kicks off CI for this batch on top of the speculative head
+// (batch.Dependencies assumed to all pass), then transitions to Speculating.
+func (c *Controller) startSpeculation(ctx context.Context, batch entity.Batch) error {
+	c.logger.Infow("starting speculation",
```
behinddwalls, line 126: we probably want to use the context aware logger for tracing/ maybe we deal it with separately to track request journey in the logs end to end

sbalabanov: not today, I just want to have something working first

PR #151, opened 2026-05-06
feat(pusher): add Pusher extension and wire it into the merge controller

sbalabanov:
## Summary

- New `extension/pusher` interface with all-or-nothing `Push` semantics, per-change outcomes (committed vs already-existed), and an `ErrConflict` sentinel for user-caused failures.
- Git-backed implementation that fetches → resets → cherry-picks → pushes, serialized via mutex on the shared checkout. Detects redundant picks and rolls back genuinely empty commits so they don't reach the remote.
- Concurrent-push contention is detected by re-checking the remote tip after a push failure (rather than parsing git error output, which varies across versions and rejection sources). The full cycle is retried up to `Params.MaxPushAttempts` (default 10).
- Merge controller now calls `Pusher.Push`, transitions the batch to `Succeeded`/`Failed`, and nacks transient infra errors. Three outcome cases are inlined (success, `ErrConflict` → Failed, generic error → return) — retryability classification will live in separate infra.
- `example/server/orchestrator` wires a git pusher from `PUSHER_CHECKOUT_PATH` / `PUSHER_REMOTE` / `PUSHER_TARGET` env vars.

## Test plan

- [x] `make test` — 30 unit tests pass, including new pusher and rewritten merge controller suites
- [x] Real-git fixture tests for the pusher cover single/stacked URIs, already-existed, mixed-partial, conflict, recovery-after-conflict, reset-between-calls, retry-on-contention, and giveup-after-cap
- [x] Pre-receive race hook drives the retry loop end-to-end with real git mechanics (unsets `GIT_QUARANTINE_PATH`/`GIT_OBJECT_DIRECTORY`/`GIT_ALTERNATE_OBJECT_DIRECTORIES` to mutate refs from inside pre-receive)
- [x] `make fmt`, `make gazelle`, `make tidy` — tree clean


extension/pusher/pusher.go
```diff
@@ -0,0 +1,88 @@
...
+// either roll back any local state or arrange for the push to never happen
+// when any change fails to apply. Callers can treat a non-nil error as
+// "the remote is exactly as it was before the call".
+//
+// On success, len(Result.Outcomes) == len(changes) and Outcomes[i] describes
+// what happened to changes[i]. A change can produce multiple commits
+// (OutcomeStatusCommitted, CommitSHAs populated in apply order) or none at
+// all (OutcomeStatusAlreadyExisted, CommitSHAs empty) — the latter happens
+// when the change's content is already present on the target branch.
+type Pusher interface {
```
behinddwalls, line 84: do we want to call it `Merge` or `Merger` as we can probably have Github Merge API as one of the impl/extensions for the merge?

extension/pusher/git/git_pusher.go
```diff
@@ -0,0 +1,432 @@
...
+// defaultMaxPushAttempts caps the retry loop in Push when the remote tip
+// keeps moving under us. Bounded to prevent an infinite loop against a
+// pathologically busy remote.
+const defaultMaxPushAttempts = 10
+
+// Params holds the dependencies for the git Pusher.
+type Params struct {
+	// CheckoutPath is the absolute path to an existing git checkout that the
+	// Pusher will operate against. The Pusher owns this working tree.
+	CheckoutPath string
```
behinddwalls, line 79: this would require some git workspace semantics to be built in the project to support it, my initial thought was to just use github apis in the open for MVP

extension/pusher/git/git_pusher.go
```diff
@@ -0,0 +1,432 @@
...
+		remote:          params.Remote,
+		target:          params.Target,
+		logger:          params.Logger.Named("git_pusher"),
+		metricsScope:    params.MetricsScope.SubScope("git_pusher"),
+		maxPushAttempts: maxAttempts,
+	}
+}
+
+// Push fulfils the pusher.Pusher contract.
+func (p *gitPusher) Push(ctx context.Context, changes []entity.Change) (pusher.Result, error) {
```
behinddwalls: should we use metrics utility to capture the metrics?

extension/pusher/pusher.go
sbalabanov: I'd like to keep it as "Pusher". Merger is the controller that does more logic. Pusher is abstracted to literally just push the change to the server.

extension/pusher/git/git_pusher.go
sbalabanov: I only did the actual implementation to fill the void and have any ability for full e2e tests, we can implement another pusher if needed, but I found this implementation easier to implement with claude and test locally.

extension/pusher/git/git_pusher.go
sbalabanov: added

extension/pusher/git/git_pusher.go
sbalabanov: Also, APIs would not guarantee atomicity, so a strict implementation of the requirements is not possible.

behinddwalls (approved):

PR #152, opened 2026-05-07
add change store extension for per-URI request claims

behinddwalls:
## Summary
Introduce a dedicated extension that tracks (URI, request_id) claims with mutable
metadata, scoped to queue. Foundation for cross-request URI overlap detection;
no consumers in this commit.

- entity/change_record.go: immutable identity (URI, RequestID), mutable metadata
- extension/changestore/: ChangeStore interface, mysql impl, mock, schema
- INSERT IGNORE on writes for retry idempotency on (uri, request_id)
- FindOverlapping is single-table; callers do their own liveness check
- integration test under test/integration/extension/changestore/mysql/
- e2e suite applies the change schema (harmless no-op until consumers land)

## Stack
1. @ #152
1. #144


entity/change_record.go
```diff
@@ -0,0 +1,45 @@
...
+	// URI identifies the change (RFC 3986). Same scheme/format as entity.Change.URIs.
+	// Example: "github://uber/submitqueue/pull/123/abc123def".
+	URI string `json:"uri"`
+
+	// RequestID is the owning land request that claimed this URI.
+	// Format matches entity.Request.ID: "<queue>/<counter_value>".
+	RequestID string `json:"request_id"`
+
+	// Queue is the queue scope for the owning request. Denormalized from the request
+	// to allow queue-scoped duplicate checks without a join.
```
sbalabanov: Do they need to be queue-scoped?

extension/changestore/mysql/schema/change.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE IF NOT EXISTS `change` (
+    uri         VARCHAR(255) NOT NULL,
+    request_id  VARCHAR(255) NOT NULL,
+    queue       VARCHAR(255) NOT NULL,
+    metadata    JSON,
```
sbalabanov: also not null?

extension/changestore/mysql/schema/change.sql
```diff
@@ -0,0 +1,9 @@
...
+    created_at  BIGINT NOT NULL,
+    updated_at  BIGINT NOT NULL,
+    PRIMARY KEY (uri, request_id)
```
sbalabanov: If queries are queue-scoped, queue should be the first in a PK?

extension/changestore/change_store.go
```diff
@@ -0,0 +1,51 @@
...
+	// excluding any records belonging to excludeRequestID (so callers can skip self when checking
+	// for duplicates of a freshly-claimed request). Returns an empty slice when there is no overlap.
+	//
+	// Liveness of the returned records' owning requests is NOT filtered here — the caller is
+	// responsible for consulting RequestStore to skip terminal owners.
+	FindOverlapping(
+		ctx context.Context,
+		queue string,
+		uris []string,
+		excludeRequestID string,
```
sbalabanov: may be the API is cleaner if exclusion happens at the callsite, not the implementation

extension/changestore/change_store.go
```diff
@@ -0,0 +1,51 @@
...
+// ChangeStore manages per-URI claim records for in-flight land requests.
+// Each row records that a specific URI was claimed by a specific request, scoped to a queue.
+// The (URI, RequestID) pair is the immutable identity of a record. Metadata may evolve over time.
+//
+// The store is the source of truth for "which URIs are or were associated with which requests".
+// Liveness of an owning request is NOT tracked here — callers must consult RequestStore separately
+// to determine whether an owner is in a terminal state.
+type ChangeStore interface {
+	// Create persists a batch of ChangeRecords. Conflicts on the (URI, RequestID) primary key
+	// are silently ignored, making the call idempotent under queue redeliveries of the same request.
```
sbalabanov: Plz also indicate a requirement for atomicity: i.e. should they all fail or succeed at the same time or not.

extension/changestore/mysql/change_store.go
```diff
@@ -0,0 +1,118 @@
...
+		return nil
+	}
+
+	const cols = 6
+	placeholders := strings.Repeat("(?, ?, ?, ?, ?, ?), ", len(records))
+	placeholders = placeholders[:len(placeholders)-2] // trim trailing ", "
+
+	args := make([]any, 0, len(records)*cols)
+	for _, r := range records {
+		// Pass empty Metadata as NULL — JSON column rejects empty string but accepts NULL.
```
sbalabanov: it should accept "{}" though as a right centinel

extension/changestore/mysql/change_store.go
```diff
@@ -0,0 +1,118 @@
...
+	op := metrics.Begin(s.scope, "find_overlapping")
+	defer func() { op.Complete(retErr) }()
+
+	if len(uris) == 0 {
+		return nil, nil
+	}
+
+	uriPlaceholders := "?" + strings.Repeat(", ?", len(uris)-1)
+	query := "SELECT uri, request_id, queue, metadata, created_at, updated_at FROM `change` " +
+		"WHERE uri IN (" + uriPlaceholders + ") AND queue = ? AND request_id != ?"
```
sbalabanov: this request needs `queue` in PK to be fast and shardable.

extension/changestore/mysql/schema/change.sql
sbalabanov: I think MYSQL limit on varchar index is around 1K. URL could be 2K (platform-dependent). We probably need to enforce URL language at the entrypoint, before it hits here.

sbalabanov (changes requested):

entity/change_record.go
behinddwalls: i think we can remove it..given requestID is encoding queue as well...completely fogot about that

extension/changestore/mysql/schema/change.sql
behinddwalls: I will update it.

extension/changestore/mysql/schema/change.sql
behinddwalls: I will make that change to the next PR in stack which does the validation

extension/changestore/mysql/change_store.go
behinddwalls: Ack

extension/changestore/mysql/change_store.go
behinddwalls: Currently I went with URI+requestID where requestID encodes queue as well...so yes now thinking, queue would make it fast and also, maybe some sort of expiry/cleanup on this data set...should keep it small... we don't have state in here i was concerned about being it inconsistent with request and just keeping it there...

Let me make `PK(queue, uri)` instead...

entity/change_record.go
behinddwalls: Yes — kept, with expanded comment in entity/change_record.go explaining that Queue is the leading column of the change-store PK so queue-scoped lookups are PK-prefix scans and the table is shardable by queue. The questions in #3204577916 / #3204651213 led me to the same conclusion.

extension/changestore/mysql/schema/change.sql
behinddwalls: Done — `metadata JSON NOT NULL`. The mysql impl substitutes `'{}'` when callers pass an empty Go string, so callers don't need to know about the column constraint.

extension/changestore/mysql/schema/change.sql
behinddwalls: Done — `PRIMARY KEY (queue, uri, request_id)`. Kept `request_id` in the PK so concurrent claims by different requests on the same URI coexist as distinct rows: same-request retries are absorbed idempotently via the PK + INSERT IGNORE, and cross-request collisions surface as additional rows that callers detect via FindOverlapping. Schema and entity comments now explain this.

extension/changestore/change_store.go
behinddwalls: Done — removed `excludeRequestID` from the interface. The store returns every matching row; callers that want to skip self filter the result by RequestID themselves. Cascaded the change to validate.go in #144.

extension/changestore/mysql/change_store.go
behinddwalls: Done — empty Metadata is written as `'{}'` instead of NULL. Integration test `TestCreate_EmptyMetadataStoredAsObject` asserts the round-trip.

extension/changestore/mysql/change_store.go
behinddwalls: Done — query is now `WHERE queue = ? AND uri IN (...)`, leading with queue to align with the new `(queue, uri, request_id)` PK. PK-prefix scan, no secondary index needed.

extension/changestore/mysql/schema/change.sql
behinddwalls: Agreed — URI length/format enforcement belongs upstream (gateway/validation), not in the storage schema. Out of scope for this PR; will follow up separately.

extension/changestore/change_store.go
behinddwalls: making it per URI so that it's one by one.

sbalabanov (approved):

PR #157, opened 2026-05-22
feat(cancel): add Cancel RPC + orchestrator cancel pipeline

sbalabanov:
## Summary

- New gateway `Cancel(sqid)` RPC that publishes a `CancelRequest` to `TopicKeyCancel`; processing is asynchronous and cancellation is not guaranteed (a request that has already merged, or that races to completion, may still land or error — callers should check the actual outcome via the separate status / request-log API).
- New orchestrator cancel controller with a two-step transition: `RequestStateCancelling` (non-terminal intent) → either `RequestStateCancelled` directly (un-batched) or via batch cancel + conclude fan-out (batched). The Cancelling intent is non-terminal so concurrent merge / failure may still win and prevail with `Landed` / `Error`.
- Every stage controller (validate, batch, score, build, buildsignal, merge, conclude, speculate) is now cancellation-aware via a single `IsRequestStateHalted` / batch-terminal short-circuit. Speculate treats cancelled deps as out-of-the-way (drop and continue) instead of cascade-failing; conclude maps `BatchStateCancelled` → `RequestStateCancelled`.
- Cancel controller does not classify storage errors itself — `storage.ErrNotFound` and `storage.ErrVersionMismatch` propagate as-is so the base controller layer can map them to retryable, matching the pattern in every other orchestrator controller.

## Test plan

- [ ] `make fmt && make lint && make check-tidy && make check-gazelle && make check-mocks` clean
- [ ] `bazel test //... --test_tag_filters=-integration,-e2e` (32 unit-test targets) all pass
- [ ] Manual e2e via `make local-start`: Land → Cancel before batch → confirm `RequestStateCancelled` and no batch created
- [ ] Manual e2e: Land → wait for batch → Cancel → confirm batch goes `Cancelled`, contained requests reconcile to `Cancelled`, dependent batches respeculate
- [ ] Integration test exercising both paths through real MySQL extension

## Out of scope (deferred)

- Re-enqueuing non-cancelled requests from a cancelled batch (TODO in cancel controller).
- Cancelling a real CI build via external API (no-op today; hook is the existing `BuildStatusCancelled` write).


albertywu:
> Cancel controller does not classify storage errors itself — storage.ErrNotFound and storage.ErrVersionMismatch propagate as-is so the base controller layer can map them to retryable, matching the pattern in every other orchestrator controller.

What base controller layer? Is this referring to [core/errs](https://github.com/uber/submitqueue/blob/164917b58fc8031549e3e558f840cb651162699a/core/errs/errs.go)? If so, it looks like the only retryable error types are context-cancellation and infra-errors -- I don't see any special handing of these specific storage errors.

behinddwalls:
> > Cancel controller does not classify storage errors itself — storage.ErrNotFound and storage.ErrVersionMismatch propagate as-is so the base controller layer can map them to retryable, matching the pattern in every other orchestrator controller.
> 
> 
> 
> What base controller layer? Is this referring to [core/errs](https://github.com/uber/submitqueue/blob/164917b58fc8031549e3e558f840cb651162699a/core/errs/errs.go)? If so, it looks like the only retryable error types are context-cancellation and infra-errors -- I don't see any special handing of these specific storage errors.

It's referring to the ConsumerController aka BaseController for all queue consumers

albertywu:
> > > Cancel controller does not classify storage errors itself — storage.ErrNotFound and storage.ErrVersionMismatch propagate as-is so the base controller layer can map them to retryable, matching the pattern in every other orchestrator controller.
> > 
> > 
> > What base controller layer? Is this referring to [core/errs](https://github.com/uber/submitqueue/blob/164917b58fc8031549e3e558f840cb651162699a/core/errs/errs.go)? If so, it looks like the only retryable error types are context-cancellation and infra-errors -- I don't see any special handing of these specific storage errors.
> 
> It's referring to the ConsumerController aka BaseController for all queue consumers

I see -- it looks like there is no base controller implemented. So as things stand today (unless I'm missing something), there is no code that maps `storage.ErrNotFound` and `storage.ErrVersionMismatch` to retry-able errors as the PR comment states.

orchestrator/controller/build/build.go
```diff
@@ -95,6 +95,30 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) (r
 		"partition_key", msg.PartitionKey,
 	)
 
+	// If the batch is halted (terminal OR cancelling), publish a cancelled
+	// Build to buildsignal instead of triggering CI. This is the "cancel the
+	// (would-be) build" behaviour for the build stage — no external CI is ever
+	// kicked off, and buildsignal's own halted short-circuit takes over. Both
+	// truly-terminal and mid-cancellation batches reach the same observable
+	// behaviour (no build performed).
+	if entity.IsBatchStateHalted(batch.State) {
```
behinddwalls, line 108: just thinking if this can be wired up into the Consumer Controller in some way,

Basically if each controller provides a way to get the deserialized payload, then consumer can just make a callback using the same entity before calling the process and short-circuit for all the consumer controllers.

orchestrator/controller/merge/merge.go
```diff
@@ -102,6 +102,16 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) (r
 		"partition_key", msg.PartitionKey,
 	)
 
+	// Cancelling intent: the cancel controller marked this batch as not landing
+	// and handed it off to speculate. Silently ack — do not push (the inherent
+	// push-already-committed race is acknowledged elsewhere) and do not fan out
+	// (speculate owns the terminal write to Cancelled and the downstream
+	// dependent / conclude publishes).
+	if batch.State == entity.BatchStateCancelling {
```
behinddwalls, line 110: should this use the IsHalted utility?

behinddwalls (approved):

behinddwalls:
> > > > Cancel controller does not classify storage errors itself — storage.ErrNotFound and storage.ErrVersionMismatch propagate as-is so the base controller layer can map them to retryable, matching the pattern in every other orchestrator controller.
> > > 
> > > 
> > > What base controller layer? Is this referring to [core/errs](https://github.com/uber/submitqueue/blob/164917b58fc8031549e3e558f840cb651162699a/core/errs/errs.go)? If so, it looks like the only retryable error types are context-cancellation and infra-errors -- I don't see any special handing of these specific storage errors.
> > 
> > 
> > It's referring to the ConsumerController aka BaseController for all queue consumers
> 
> I see -- it looks like there is no base controller implemented. So as things stand today (unless I'm missing something), there is no code that maps `storage.ErrNotFound` and `storage.ErrVersionMismatch` to retry-able errors as the PR comment states.

https://github.com/uber/submitqueue/blob/main/core/consumer/controller.go
https://github.com/uber/submitqueue/blob/main/core/consumer/consumer.go

albertywu:
> > > > > Cancel controller does not classify storage errors itself — storage.ErrNotFound and storage.ErrVersionMismatch propagate as-is so the base controller layer can map them to retryable, matching the pattern in every other orchestrator controller.
> > > > 
> > > > 
> > > > What base controller layer? Is this referring to [core/errs](https://github.com/uber/submitqueue/blob/164917b58fc8031549e3e558f840cb651162699a/core/errs/errs.go)? If so, it looks like the only retryable error types are context-cancellation and infra-errors -- I don't see any special handing of these specific storage errors.
> > > 
> > > 
> > > It's referring to the ConsumerController aka BaseController for all queue consumers
> > 
> > 
> > I see -- it looks like there is no base controller implemented. So as things stand today (unless I'm missing something), there is no code that maps `storage.ErrNotFound` and `storage.ErrVersionMismatch` to retry-able errors as the PR comment states.
> 
> https://github.com/uber/submitqueue/blob/main/core/consumer/controller.go https://github.com/uber/submitqueue/blob/main/core/consumer/consumer.go

Should we make sure these error types are retriable? Otherwise we fall into the DLQ code-path here instead of the retry code-path: https://github.com/uber/submitqueue/blob/main/core/consumer/consumer.go#L374

PR #159, opened 2026-05-27
fix(github): align codebase to use a consistent github uri

albertywu:
## Summary

Aligns proto documentation and internal implementation to use a consistent canonical github uri shape:

```
github://uber/repo/pull/123/<sha>
```

### Before
`ParseChangeID` silently misparsed input URLs because it wasn't handling `/pull/` in the URI:

For example: 
`github://uber/repo/pull/123/<sha>` would parse to `Org="uber/repo" Repo="pull"` instead of erroring.

---

### After
- `github://uber/repo/pull/123/<sha>` parses correctly to `Org="uber" Repo="repo"`
- parser ensures that `<sha>` is canonical 40-char lowercase hex and returns error if not. Parsing happens at edge of application such that downstream internals can be assured that the `<sha>` is always in standard format.

## Test Plan
✅  `make test`

### DIFF
--- a/entity/github/change_id.go
+++ b/entity/github/change_id.go
@@ -21,12 +21,25 @@ import (
 )
 
 // changeIDFormat is the expected format for change IDs, included in error messages.
-const changeIDFormat = "{scheme}://{owner}/{repo}/{pr_number}/{head_commit_sha}"
+const changeIDFormat = "{scheme}://{owner}/{repo}/pull/{pr_number}/{head_commit_sha}"
+
+// pullSegment is the literal segment separating the repo path from the pull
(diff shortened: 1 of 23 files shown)


entity/github/change_id.go
```diff
@@ -21,12 +21,25 @@ import (
 )
 
 // changeIDFormat is the expected format for change IDs, included in error messages.
-const changeIDFormat = "{scheme}://{owner}/{repo}/{pr_number}/{head_commit_sha}"
```
albertywu, line 24: previous changeIDFormat was missing `/pull/` here

behinddwalls (approved):

sbalabanov (approved):

PR #161, opened 2026-05-28
feat(phabricator): add ChangeID entity for parsing phab:// URIs

albertywu:
## Summary
Mirrors entity/github.ChangeID with a Phabricator-shaped identifier: `phab://D{revision_id}/{diff_id}`. 

Parser validates scheme, the "D" prefix on the revision segment, positive integer revision/diff IDs, and exact
segment count. Adds a Revision() accessor that returns the canonical "D{n}" form used everywhere in the Phabricator UI and CLI.

No downstream consumers yet — changeprovider/mergechecker/pusher wiring will come in follow-up PRs.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

## Test Plan
✅`make test`

### DIFF
--- a/entity/phabricator/change_id.go
+++ b/entity/phabricator/change_id.go
@@ -16,6 +16,7 @@ package phabricator
 
 import (
 	"fmt"
+	"regexp"
 	"strconv"
 	"strings"
 )
@@ -33,6 +34,15 @@ const scheme = "phab"
 // Phabricator UI and CLI ("arc diff", "arc patch D12345"), so URIs can be
 // constructed by trivial substitution rather than reshaping.
 const revisionPrefix = "D"
+
+// revisionPattern matches a revision segment: the literal "D" followed by a
+// positive integer with no leading zero (matches the strict round-trip form
+// produced by String()).
+var revisionPattern = regexp.MustCompile(`^D([1-9]\d*)$`)
+
+// diffPattern matches a diff segment: a positive integer with no leading zero
+// (matches the strict round-trip form produced by String()).
+var diffPattern = regexp.MustCompile(`^[1-9]\d*$`)
 
 // ChangeID represents a parsed Phabricator change identifier.
 // Format: phab://D{revision_id}/{diff_id}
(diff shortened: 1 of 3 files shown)


entity/phabricator/change_id.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	revisionSegment := segments[0]
+	diffSegment := segments[1]
+
+	revisionMatch := revisionPattern.FindStringSubmatch(revisionSegment)
+	if revisionMatch == nil {
+		return ChangeID{}, fmt.Errorf("invalid change ID %q: revision %q must match D{positive_int} (expected format: %s)", raw, revisionSegment, changeIDFormat)
+	}
+	revisionID, _ := strconv.Atoi(revisionMatch[1])
```
behinddwalls: we likely need to handle the error for overflow cases to avoid setting revisionID to 0

entity/phabricator/change_id.go
```diff
@@ -0,0 +1,120 @@
...
+
+	if !diffPattern.MatchString(diffSegment) {
+		return ChangeID{}, fmt.Errorf("invalid change ID %q: diff %q must be a positive integer (expected format: %s)", raw, diffSegment, changeIDFormat)
+	}
+	diffID, _ := strconv.Atoi(diffSegment)
```
behinddwalls: same as above

entity/phabricator/change_id.go
albertywu: Good catch — fixed in 6b7cd85. The regex pre-filter guarantees only `[1-9]\d*` reaches `Atoi`, so the only remaining failure mode is int overflow on >19-digit inputs. That used to silently fall through to `revisionID = 0`, producing a ChangeID the regex was specifically designed to forbid. Now returns the overflow error. Added a 20-digit test case for each segment.

entity/phabricator/change_id.go
albertywu: Same fix applied to the diff segment in 6b7cd85.

behinddwalls (approved):

PR #175, opened 2026-06-02
feat(buildrunner): add vendor-agnostic BuildRunner interface + RFC

behinddwalls:
### Why?

The build stage needs a vendor-agnostic abstraction for talking to a
Build Runner — one that fits the existing extension family (storage,
queue, conflict) and that does not have to change when the first real
backend lands. Design rationale lives in `doc/rfc/build-runner.md`.

### What?

Introduces the `BuildRunner` contract — three verbs, all keyed by an `entity.BuildID`: `Trigger` submits a build (ordered `base` + `head` change sets plus a free-form metadata map) and returns its ID; `Status` fetches the current `BuildStatus` and runner-defined metadata; `Cancel` requests cancellation. See `extension/buildrunner/build_runner.go` for the signatures.

Highlights:

- `Trigger` takes ordered `base` (assumed-good prefix from dependency
  batches) and `head` (changes being verified) — a runner can cache or
  short-circuit a prefix it has validated before, and attribute terminal
  failures to base vs head via `BuildMetadata`.
- Build identifiers are the typed `entity.BuildID` (a lightweight
  `{ID string}` value, also the queue payload envelope) rather than a
  bare `string`, so the runner contract and the on-queue messages share
  one identifier type and the compiler distinguishes a build ID from
  other string IDs.
- `metadata` is a free-form `map[string]string` for caller-supplied
  attributes (requester, ticket ID, trace ID) the runner MAY persist or
  echo back via `Status`.
- `Trigger` and `Cancel` return promptly; `Status` MAY be lengthy.
- The `BuildStatus` enum is narrowed to `Accepted` / `Running` /
  `Succeeded` / `Failed` / `Cancelled` so every backend can map its
  native lifecycle without leaking runner-specific stages.
- `BuildMetadata` is added as a free-form map for round-tripping
  runner-defined attributes (build URL, duration, SHA, etc.).

Naming: the interface is `BuildRunner` (not `BuildManager` — the
"Manager" suffix doesn't describe what it does, and "Build" alone would
collide cognitively with `entity.Build`). The package is
`extension/buildrunner` to follow the repo convention used by
`mergechecker`, `changeprovider`, `pusher`, `counter`, and `storage`.

Implementation (noop backend, queue `PublishAfter` primitive, controller
wiring, and the poll-driven `buildsignal` loop) lands in the follow-up
PRs stacked on top of this one.

## Test Plan

- ✅ `make test` — entity/build tests pass; mock-only consumers of the
  interface compile and pass.
- ✅ `make fmt lint check-tidy check-gazelle check-mocks` — clean.
- ✅ `make build` — all targets compile.

## Stack
1. @ #175
1. #179
1. #176
1. #177


doc/rfc/build-runner.md
```diff
@@ -0,0 +1,240 @@
...
+
+## Interface
+
+```go
+type BuildRunner interface {
+    Trigger(
+        ctx context.Context,
+        queueName string,
+        base []entity.Change,
+        head []entity.Change,
```
JamyDev: Based on [request.go](https://github.com/uber/submitqueue/blob/main/entity/request.go#L58-L68) Change is already a list of changes. While I think the buildmanager is likely best to take a list of changes (given that we may speculate multiple changesets at once); is the idea that `Change` itself could encompass something like a stack of PRs/Diffs? And then the set of base/heads provided here, would be the full speculation stack?

Is there an upper bound of the length of `len(flatten(base)) + len(flatten(head))`? I think the higher this gets, the more likely we would hit merge conflicts (unless that is already considered before hitting the BuildManager)

doc/rfc/build-runner.md
```diff
@@ -0,0 +1,240 @@
...
+        metadata entity.BuildMetadata,
+    ) (buildID string, err error)
```
JamyDev: Trigger being fire and forget requires the BuildManager to track builds internally, so we can't refer to the buildID we get back from the downstream. I think that this is fine, but wanted to call it out.

doc/rfc/build-runner.md
JamyDev: Do we need an indicator of the initial change to apply to, or should the assumption be that base[0][0] will be the base commit/immutable to apply everything to?

I'm thinking about the following scenario:

```
base = [Change{URIs: [pr1_1, pr1_2]}, Change{URIs: [pr2]}]
head = [Change{URIs: [pr3]}]
```

Is the expectation that pr1_1 gets applied on top of main (as far as CI sees it), or on top of `pr1_1` base? Should we be explicit here and see "apply everything to base identifier x"? Should we support specifying that in the `base` list? (Currently `Change` only supports referencing PRs based on the URI format).

extension/buildrunner/README.md
```diff
@@ -0,0 +1,20 @@
...
+itself.
+
+## Adding a new backend
+
+1. Create `extension/buildrunner/{backend}/` with a `BuildRunner`
+   implementation bound to its runner configuration at construction.
+2. Map the `base` and `head` change slices onto the backend's build
+   primitives (apply `base`, apply `head`, validate the result).
+3. Map the runner's lifecycle states down to the `BuildStatus` values:
+   `Accepted` (accepted for execution), `Running` (executing), and the
```
JamyDev: Let's say we accept a build, but then downstream is failing. How long should buildmanager retry until we mark it as failed? Or should buildmanager retry infinitely, and wait for a cancel signal from upstream?

doc/rfc/build-runner.md
behinddwalls: I think for Build system, it's just a list of change (it's unaware of stack in general), as stacks themselves could be batched with other changes, so everything just flattened in the order it should be applied.

For base question, I think in general the test is against the target branch where the merge is to happen, in this case it's somewhat implicit based on the queue, build system should be able infer what it means in VCS system through some config or implicitly

doc/rfc/build-runner.md
behinddwalls: Yes.. i think that's the responsibility of the caller in this SQ to be track and store. I have follow-up change to wire up things

extension/buildrunner/README.md
behinddwalls: i think in general, 2-3 retries would be enough and then just give-up/fail, it mostly helps with transient issues, in general if that happens, it might be best to pause the queue rather than failing everything

doc/rfc/build-runner.md
JamyDev: > the test is against the target branch where the merge is to happen

So de-facto configured main, makes sense. That does mean that if you send me two build requests (for speculation) where `base = [A B C], head = [D]`, and `base = [A B C D], head = [E]`, they may be tested on distinctly different main commits depending on the upstream target branch (eg for the first build `main = x` and for the second build `main = y`). I feel like that reduces determinism of tests?

behinddwalls: True. As we are speculating on future states, we are assuming we have those futures covered which may become true by the time this is landed, so from that perspective, we already accounted for that in speculation paths.

albertywu:
When would the orchestration engine call the synchronous build `Status` API vs use the [buildsignal topic](https://github.com/uber/submitqueue/blob/main/orchestrator/controller/buildsignal/buildsignal.go) to handle reading build results asynchronously?

edit: nvm i see your 4th PR here: https://github.com/uber/submitqueue/pull/177

JamyDev (approved):

PR #178, opened 2026-06-02
docs(storage): clarify BatchDependentStore invariant and Get semantics

sbalabanov:
## Summary
- Document on the `BatchDependentStore` interface that every active `Batch` has a paired `BatchDependent` row, since the batch-creation flow calls `Create` here before creating the `Batch` itself.
- Make explicit that `Get` is only called for batch IDs returned from the active-batch set, so `ErrNotFound` is part of the contract for completeness but signals an invariant violation rather than an expected outcome.
- Note on `Get` that an empty `Dependents` list is the normal "no dependents" result.

## Test plan
- [ ] CI passes (docs-only change; no behavior modified)

### DIFF
--- a/extension/storage/batch_dependent_store.go
+++ b/extension/storage/batch_dependent_store.go
@@ -23,13 +23,22 @@ import (
 )
 
 // BatchDependentStore is an interface that defines methods for managing batch dependent information in the database.
+//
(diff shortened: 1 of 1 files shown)


behinddwalls (approved):

sbalabanov:
signed CLA but GH does not pick up. Bypassing.

PR #182, opened 2026-06-02
feat(stovepipe): add orchestrator service

mnoah1:
Set up stovepipe orchestrator service and ensure that the stack can be launched.

## Why?
Similar to submitqueue, we will use the gateway service for API handling, and an orchestrator service to process work.  This will give us more flexibility in deployment.

## What?
Add separate orchestrator layer which will mainly be used to provide a controller that can be injected via fx and deployed.

## Test Plan
 - make local-stovepipe-start
 - CI


behinddwalls (commented):
do you want to add extension, entity to the stovepipe/orchestrator as well?

mnoah1:
> do you want to add extension, entity to the stovepipe/orchestrator as well?

We currently have them one level above (e.g. [here](https://github.com/uber/submitqueue/tree/main/stovepipe)), to be shared between orchestrator/gateway.  I think this should be ok - we'll probably be more likely to want to share between stovepipe orchestrator and gateway than across the boundary between submitqueue and stovepipe.

behinddwalls (approved):

PR #183, opened 2026-06-03
refactor(errs): introduce per-backend Classifier framework

sbalabanov:
## Summary

Implements an error classification framework in `core/errs`: per-backend `Classifier` implementations recognise raw driver/library errors and the consumer runs a single `errs.Classify(err, classifiers...)` pass after each delivery. Downstream `IsRetryable` / `IsUserError` / `IsDependencyError` become plain type checks; an explicit framework wrap from a controller always wins over any classifier.

- Adds `errs.Classifier` interface, `Verdict` enum, and `errs.Classify` (two-pass: framework-wrap short-circuit, then classifier walk).
- New `core/errs/generic` — `context.Canceled` → retryable infra.
- New `core/errs/mysql` — server error numbers, `driver.ErrBadConn`, `sql.ErrConnDone`/`TxDone`, `net.Error`. Never returns `User`; controllers wrap constraint violations with `NewUserError` explicitly.
- Orchestrator's `consumer.New` wires both classifiers.
- `core/errs/README.md` documents the framework, adding classifiers, and controller overrides.

## Test plan

- [ ] CI green


core/errs/mysql/mysql.go
```diff
@@ -0,0 +1,167 @@
...
+//
+// The classifier is stateless; this package-level singleton is the canonical
+// handle. Pass it into errs.Classify (typically as a vararg to consumer.New).
+var Classifier errs.Classifier = classifier{}
+
+type classifier struct{}
+
+// Classify dispatches a single error node to one of the recognisers above.
+// See the Classifier var docs for the full list.
+func (classifier) Classify(err error) errs.Verdict {
```
behinddwalls, line 65: should this be used in extensions itself or controllers? the one thing which i don't see here is how do we go about wrapped errors...if say stoarge throws MySQLError but then it gets wrapped in controller with a generic fmt.Error

EDIT:
I see it's in errs.go file

core/errs/errs.go
```diff
@@ -107,34 +106,147 @@ func (e *infraError) Is(target error) bool {
...
+// tagging; intentionally deferred.
+func Classify(err error, classifiers ...Classifier) error {
+	if err == nil {
+		return nil
+	}
+
+	// Pass 1 — cheap framework-wrap check. If any node already carries a
+	// framework type, the chain is interpretable as-is and classifiers are
+	// never invoked.
+	for cur := err; cur != nil; cur = errors.Unwrap(cur) {
```
behinddwalls, line 196: do we expect mutli-errors to work in this case?

behinddwalls (approved):

core/errs/errs.go
sbalabanov: what are multi-errors? This function inspects the whole error chain, same as Errors.As/Is does

behinddwalls: errors.Join(...) stuff which returns array of errors... i don't think we are doing it anywhere yet, just making sure if we do it later at some point, i can be calssified..

PR #190, opened 2026-06-04
feat(buildrunner): add Buildkite BuildRunner implementation

JamyDev:
## What

  Adds `submitqueue/extension/buildrunner/buildkite` — a `BuildRunner` implementation backed by the Buildkite CI platform. The package is four files:

  - **`buildkite.go`** — `runner` struct implementing `BuildRunner.Trigger`, `.Status`, and `.Cancel`
  - **`client.go`** — thin HTTP wrapper over the three Buildkite REST endpoints (create/get/cancel build)
  - **`buildkite_test.go`** — 16 unit tests backed by an in-process `httptest` server
  - **`BUILD.bazel`** — Bazel build file

  `Trigger` calls the Buildkite API to create a build, encoding the base and head change URIs as JSON-encoded environment variables (`SQ_BASE_URIS`, `SQ_HEAD_URIS`, `SQ_QUEUE`). The Buildkite
  pipeline script uses these to fetch diffs via the GitHub API, apply them in order, and run CI. `Status` and `Cancel` derive the Buildkite build number from the `BuildID` string — no local
  state is needed.

  Auth and base URL are injected via HTTP transport layers on the caller-provided `*http.Client` (using `httpclient.BaseURLTransport` + an auth transport), following the existing `httpclient`
  pattern and keeping credentials out of the struct.

  ## Why

  The `BuildRunner` extension interface was designed to support multiple CI platforms as pluggable backends. This adds the first real (non-noop) implementation. Buildkite is the target CI
  platform for speculative merges: its pipeline model maps cleanly onto the base/head separation in `BuildRunner.Trigger` — the pipeline applies the base layer first (potentially cacheable
  across batches), then the head layer, then runs the full test suite.

  Using environment variables to pass change URIs keeps the SQ→Buildkite contract simple and auditable; the pipeline script is responsible for materialising the diffs, isolating SQ from
  Buildkite's pipeline YAML.

  ## Tests

  All three interface methods are tested against an in-process `httptest` server — no external dependencies or mocks:

  - **`Trigger`**: validates HTTP method, request payload (env var keys/values), URI flattening across multiple `Change` entries, and that a nil base produces `[]` (not `null`) in JSON
  - **`Status`**: state mapping — exhaustive table over all documented Buildkite states (`creating`, `scheduled`, `running`, `blocked`, `passed`, `failed`, `not_run`, `skipped`, `canceling`,
  `canceled`) plus unknown future states (mapped to non-terminal `Unknown` so the poll loop keeps waiting rather than failing a batch on an unrecognised state); also validates build URL is
  returned in metadata
  - **`Cancel`**: verifies the cancel endpoint is called; verifies 422 (already-terminal build) is treated as a no-op
  - **Helpers**: encode/parse round-trip, invalid build ID rejection


submitqueue/extension/buildrunner/buildkite/config.go
```diff
@@ -0,0 +1,73 @@
...
+import (
+	"net/http"
+	"time"
+)
+
+// Config holds the per-queue settings for a Buildkite BuildRunner. It is
+// constructed by Factory.For for each queue and bound to a single pipeline.
+type Config struct {
+	// APIToken is the Buildkite personal or agent API token. Required.
+	APIToken string
```
behinddwalls: should this be left to be injected through http transport? similar to changeprovider & mergechecker model?

submitqueue/extension/buildrunner/buildkite/config.go
```diff
@@ -0,0 +1,73 @@
...
+	// Resolved by Factory from its Pipelines map. Required.
+	PipelineSlug string
+
+	// QueueName is the SQ queue this runner serves. Passed as SQ_QUEUE in
+	// the build environment so the pipeline script can select queue-specific
+	// test targets.
+	QueueName string
+
+	// Branch is the target branch builds run against (e.g. "main"). Required.
+	Branch string
```
behinddwalls: do we need this when have QueueName as the main thing being passed around? should underlying imp infer things based on that queue however they configure or store..rather than we passing something as queueName uniquely represents the repo+target

submitqueue/extension/buildrunner/buildkite/config.go
```diff
@@ -0,0 +1,73 @@
...
+
+	// TriggerQueueSize is the buffer capacity of the async trigger channel.
+	// Trigger returns an error when the channel is full (the build controller
+	// will nack and retry). Defaults to 256.
+	TriggerQueueSize int
+
+	// CancelQueueSize is the buffer capacity of the async cancel channel.
+	// Defaults to 256.
+	CancelQueueSize int
```
behinddwalls: i think we can get rid of buffers

submitqueue/extension/buildrunner/buildkite/config.go
```diff
@@ -0,0 +1,73 @@
...
+	SubmitTimeout time.Duration
+
+	// MaxSubmitAttempts is the number of times the background worker tries a
+	// Buildkite create/cancel call before giving up, to ride out transient
+	// failures. Defaults to 5.
+	MaxSubmitAttempts int
+
+	// SubmitBackoff is the base delay between background-worker retries; the
+	// delay grows linearly with the attempt number. Defaults to 1s.
+	SubmitBackoff time.Duration
```
behinddwalls: should we just accept a retry-policy instead?

submitqueue/extension/buildrunner/buildkite/config.go
```diff
@@ -0,0 +1,73 @@
...
+
+	// HTTPClient overrides the HTTP client used for Buildkite API calls.
+	// If nil, http.DefaultClient is used. Intended for testing.
+	HTTPClient *http.Client
+
+	// BaseURL overrides the Buildkite API base URL
+	// (default: "https://api.buildkite.com/v2"). Intended for testing.
+	BaseURL string
```
behinddwalls: can this be encoded into http transporter as well?

submitqueue/extension/buildrunner/buildkite/config.go
```diff
@@ -0,0 +1,73 @@
...
+	// APIToken is the Buildkite personal or agent API token. Required.
+	APIToken string
+
+	// OrgSlug is the Buildkite organisation slug (URL segment after
+	// buildkite.com/). Required.
+	OrgSlug string
+
+	// PipelineSlug is the Buildkite pipeline that runs builds for this queue.
+	// Resolved by Factory from its Pipelines map. Required.
+	PipelineSlug string
```
behinddwalls: can these be in the transporter itself?

submitqueue/extension/buildrunner/buildkite/factory.go
```diff
@@ -0,0 +1,78 @@
...
+	// BaseURL overrides the Buildkite API base URL for all produced runners.
+	// Intended for testing.
+	BaseURL string
+}
+
+var _ buildrunner.Factory = Factory{}
+
+// For returns a BuildRunner bound to the Buildkite pipeline configured for
+// cfg.QueueName. Returns an error if no pipeline is configured for the queue.
+func (f Factory) For(cfg buildrunner.Config) (buildrunner.BuildRunner, error) {
```
behinddwalls: I was hoping it's implemented outside of the extension, basically in example server we would define this and let implementer decide however they want to wire the implementation per queue, so say one queue could use buildkite, another one can use gh-actions

submitqueue/extension/buildrunner/buildkite/config.go
JamyDev: Yeah I think that makes sense, also wanted to add some logging. Will update.

submitqueue/extension/buildrunner/buildkite/config.go
```diff
@@ -0,0 +1,23 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package buildkite
+
+// Config holds the per-queue settings for a Buildkite BuildRunner.
+type Config struct {
```
behinddwalls: wondering if we need it? or we can use this: https://github.com/uber/submitqueue/blob/main/submitqueue/extension/buildrunner/build_runner.go#L86

Unless we plan to add more things down the line..might be redundant... but i can see we adding things to it once we start integrating i guess

submitqueue/extension/buildrunner/buildkite/buildkite.go
```diff
@@ -0,0 +1,297 @@
...
+		client: c,
+		refs:   make(map[string]string),
+	}
+}
+
+// Trigger generates a unique build ID, calls the Buildkite API to create the
+// build, caches the Buildkite reference, and returns the build ID. Errors are
+// propagated to the caller so the queue consumer can nack and retry.
+func (r *runner) Trigger(ctx context.Context, base, head []entity.Change, _ entity.BuildMetadata) (entity.BuildID, error) {
+	id := newBuildID()
```
behinddwalls: can we not just return buildkite's buildID as is?

submitqueue/extension/buildrunner/buildkite/buildkite.go
```diff
@@ -0,0 +1,297 @@
...
+	// EnvKeyQueue carries the SQ queue name so the pipeline script can select
+	// queue-specific test targets.
+	EnvKeyQueue = "SQ_QUEUE"
+)
+
+// metaKeyBuildID is the Buildkite build-metadata key under which the SQ build
+// ID is stored at create time. Status and Cancel filter builds by this key to
+// recover the Buildkite reference when the in-memory cache misses (e.g. after
+// a restart), which keeps that cache from being a durable source of truth.
+const metaKeyBuildID = "sq_build_id"
```
behinddwalls: do we need in-mem at all? if we returned Buildkite's buildID as is? we can just call Status/Cancel on the same buildID? any reason we don't want to do that. Also, caching si somewhat not optimal in this case, as the SQ V2 doesn't have any single host affinity in general so queues would be across multiple hosts in the fleet all trying to get same BuildID from different workflows. Especially Status Poll loop is partitioned by buildID itself, so each buildID poll could live in any host in the fleet.

JamyDev: Yeah haven't had a chance yet to review this part of it yet, will clean up

submitqueue/extension/buildrunner/buildkite/buildkite.go
```diff
@@ -0,0 +1,212 @@
...
+// pipeline.
+//
+// The HTTPClient must have BaseURLTransport configured to the pipeline's API
+// root (e.g. "https://api.buildkite.com/v2/organizations/{org}/pipelines/{slug}"),
+// and an auth transport that injects the Authorization header.
+func NewBuildRunner(params Params) (buildrunner.BuildRunner, error) {
+	httpClient := params.HTTPClient
+	if httpClient == nil {
+		httpClient = http.DefaultClient
+	}
```
behinddwalls: should we fail given we expect transports to be wired up and be passed and if nil, there is no point is continueing?

behinddwalls (approved):

submitqueue/extension/buildrunner/buildkite/buildkite.go
```diff
@@ -0,0 +1,214 @@
...
+// flattenURIs concatenates the URI lists from all changes into a single slice.
+func flattenURIs(changes []entity.Change) []string {
+	uris := make([]string, 0, len(changes))
+	for _, c := range changes {
+		uris = append(uris, c.URIs...)
+	}
+	return uris
+}
+
+// encodeBuildNumber encodes a Buildkite build number as the SQ build ID.
```
albertywu, line 175: does this build id need to be namespaced somehow to prevent conflicts if there are multiple build systems used that might use the same id number?

submitqueue/extension/buildrunner/buildkite/buildkite.go
```diff
@@ -0,0 +1,214 @@
...
+		client: c,
+		logger: logger,
+	}
+}
+
+// Trigger calls the Buildkite API to create the build and returns the Buildkite
+// build number as the build ID. Errors are propagated to the caller so the
+// queue consumer can nack and retry.
+func (r *runner) Trigger(ctx context.Context, base, head []entity.Change, _ entity.BuildMetadata) (entity.BuildID, error) {
+	baseJSON, _ := json.Marshal(flattenURIs(base))
```
albertywu, line 110: just curious (outside scope of this PR) -- will we actually apply multiple base changes, and if so why? This is different from SQV1 right?

submitqueue/extension/buildrunner/buildkite/buildkite.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+
+	r.logger.Debugw("triggered Buildkite build",
+		"buildkite_number", resp.Number,
+	)
+	return entity.BuildID{ID: encodeBuildNumber(resp.Number)}, nil
+}
+
+// Status fetches the current state of the build from Buildkite and returns it
+// with the build URL in BuildMetadata["url"].
```
albertywu, line 134: Can we check w/CI team to see if any concerns with the polling approach mainly w.r.t rate limits -- are we are planning on using polling in production or a webhook-based approach?

submitqueue/extension/buildrunner/buildkite/buildkite.go
```diff
@@ -0,0 +1,214 @@
...
+// Params holds the dependencies for a Buildkite BuildRunner. The caller is
+// responsible for configuring HTTPClient with the base URL (via
+// httpclient.BaseURLTransport) and auth (via an Authorization-header transport).
+type Params struct {
+	// Config holds the per-queue identity for this BuildRunner.
+	Config buildrunner.Config
+	// HTTPClient is a pre-configured HTTP client. The caller is responsible
+	// for the base URL (via httpclient.BaseURLTransport) and auth (via a
+	// transport layer). If nil, http.DefaultClient is used.
+	HTTPClient *http.Client
```
albertywu, line 76: will we be able to use this extension as-is for our on-prem deployment or would we need one that does everything here except uses CI Gateway?

submitqueue/extension/buildrunner/buildkite/buildkite.go
JamyDev: I think it doesn't matter since inside the orchestrator they are all bounded by their batch ids, but we can always switch it to uuid if we want to fully avoid conflicts. Build numbers on BK are unique per pipeline.

submitqueue/extension/buildrunner/buildkite/buildkite.go
JamyDev: v1 did apply multiple changes as bases, only the last change in the list was considered the "head change". I think distinctly in v2 we allow for more than a single test change to be applied to the batch, but the bases follow the same pattern as v1.

submitqueue/extension/buildrunner/buildkite/buildkite.go
JamyDev: Will have to double check, this was just to have a reference implementation for OSS and just to test the design; I would likely have a completely separate internal one for ci-gw

PR #191, opened 2026-06-04
refactor(storage): fold ChangeStore into storage, drop dead store

behinddwalls:
### Why?

Two "change"-related stores were in the wrong shape. `ChangeStore` — the
real, used store that records per-URI claims for in-flight requests and
backs `start`'s URI claiming and `validate`'s overlap detection — lived
as its own top-level extension and was injected into controllers as a
separate dependency, bypassing the `storage.Storage` aggregator that owns
every other entity store. Meanwhile `ChangeProviderStore` (exposed via
`Storage.GetChangeProviderStore()`, with a mysql impl, mock, and schema)
was dead code: no controller ever called it, and its `entity.ChangeProvider`
was orphaned alongside it.

### What?

- Move `ChangeStore` into `package storage`: the interface, mysql impl,
  and `change.sql` schema now live under `extension/storage[/mysql]`, and
  `ChangeStore` is a first-class member of the `Storage` aggregator via
  `GetChangeStore()` — matching every other store.
- `start` and `validate` controllers drop their separate `changeStore`
  constructor param and read `store.GetChangeStore()`; the example
  orchestrator no longer constructs/injects it separately.
- Delete the dead `ChangeProviderStore` (interface, mysql, mock, schema),
  `GetChangeProviderStore()`, and the orphaned `entity/change_provider.go`.
- Fold the standalone changestore integration suite into the shared
  `StorageContractSuite` (driven through `GetChangeStore()`); e2e drops the
  now-redundant changestore schema apply.

## Test Plan
- ✅ `make build`, `make test` (start/validate controllers pass against the
  storage-package mock)
- ✅ `make lint`, `make check-gazelle`, `make check-mocks`, `make check-tidy`
  (no drift; `go.mod` / `MODULE.bazel` unchanged)
- ✅ `make integration-test` (storage mysql suite now exercises the change
  store via `GetChangeStore()`)
- ✅ `make e2e-test` (full land→validate flow: URI claim + overlap detection,
  `change` table applied from the storage schema dir)

## Stack
1. @ #191
1. #195
1. #196
1. #197


submitqueue/extension/storage/README.md
```diff
@@ -1,6 +1,12 @@
 # Storage
 
-Pluggable persistence interfaces for SubmitQueue entities (requests, batches, dependents, logs, etc.). Implementations live under `extension/storage/<impl>/`.
+Pluggable persistence interfaces for SubmitQueue entities (requests, changes, batches, dependents, logs, etc.). Implementations live under `extension/storage/<impl>/`.
+
+The `Storage` interface aggregates every entity store behind a single injectable dependency; callers obtain a specific store through its accessor (e.g. `GetRequestStore`, `GetChangeStore`).
+
+## Change store
```
sbalabanov: should not be in a general description

sbalabanov: Probably no need to change README at all for this change

sbalabanov (changes requested):

submitqueue/extension/storage/README.md
behinddwalls: Reverted `submitqueue/extension/storage/README.md` entirely — it now matches main, so this PR no longer touches the general storage description. (The only remaining README changes are the deletion of the folded-away `changestore/README.md` and a `change` table entry in the schema README, following that file's existing per-table pattern.)

JamyDev: It's still here tho

behinddwalls: claude i guess missed pushing it..let me do it again

behinddwalls: removed

JamyDev (approved):

PR #192, opened 2026-06-04
feat(dlq): per-topic DLQ reconciliation + ErrorProcessor split

sbalabanov:
## Summary

- Adds a `submitqueue/orchestrator/controller/dlq/` package that drains every primary pipeline topic's `{topic}_dlq` companion and transitions the affected request or batch to a terminal failed state. Without this, a request that exhausts its retry budget in the primary pipeline stays stuck in a non-terminal state and the gateway keeps reporting it as "in progress".
- Splits error-classification policy from consumer transport via a new `errs.ErrorProcessor` interface with two implementations: `NewClassifierProcessor` for primary pipeline consumers (framework-wrap short-circuit + per-node classifier walk) and `AlwaysRetryableProcessor` for DLQ consumers (every failure redelivered). DLQ subscriptions disable their own DLQ and run with `Retry.MaxAttempts=1000` to avoid a `_dlq_dlq` cascade while preserving convergence.
- Documentation: new `dlq/README.md` (DLQ-is-final-destination design constraints), updated `core/errs/README.md` (processor choices), updated `submitqueue/core/consumer/README.md` (`ErrorProcessor` argument + error-handling rewrite), new DLQ reconciliation section in `doc/rfc/submitqueue/workflow.md`, and a sentence in `CLAUDE.md` on the two default classifiers.

## Design notes

- **DLQ is the final destination.** A genuinely unprocessable DLQ message (e.g. malformed payload) must be removed by an operator. The controllers do not attempt to recover; they only reconcile request/batch state to terminal failed states with the same optimistic-locking CAS the primary pipeline uses.
- **Two controller shapes.** `RequestController` for request-scoped topics (start, validate, batch, cancel, log); `BatchController` for batch-scoped topics (score, speculate, build, merge, conclude) with fan-out to member requests; small dedicated controller for `buildsignal` payloads.
- **Error handling in `dlq.go`.** `storage.ErrNotFound` → warn+skip; `storage.ErrVersionMismatch` → wrapped retryable; everything else → plain (then forced retryable by `AlwaysRetryableProcessor`).

## Test plan
- [x] `bazel test //core/errs/... //submitqueue/core/consumer:consumer_test //submitqueue/orchestrator/controller/dlq:dlq_test` — all PASSED locally.
- [x] `bazel build //example/submitqueue/... //example/stovepipe/...` — clean locally.
- [ ] Watch CI on the draft PR.


submitqueue/orchestrator/controller/dlq/log.go
```diff
@@ -0,0 +1,92 @@
...
+	return &logController{
+		logger:        logger.Named(name),
+		metricsScope:  scope.SubScope(name),
+		topicKey:      topicKey,
+		consumerGroup: consumerGroup,
+	}
+}
+
+// Process records that a log message landed in the DLQ and acks it.
+func (c *logController) Process(_ context.Context, delivery consumer.Delivery) (retErr error) {
```
behinddwalls, line 60: ideally the request log is consumed by gateway and orchestrator is just a publisher to it...

submitqueue/orchestrator/controller/dlq/dlq.go
```diff
@@ -0,0 +1,176 @@
...
+	// The log entry is written unconditionally — even when the state was already
+	// terminal on entry — so a previous DLQ attempt that succeeded in flipping
+	// the state but then failed to insert the log is repaired on redelivery.
+	// A duplicate log entry from such a retry is the accepted trade-off; a
+	// missing one would leave the gateway-visible status divergent from the
+	// entity state.
+	logEntry := entity.NewRequestLog(requestID, entity.RequestStatusError, logVersion, "", nil)
+	if err := store.GetRequestLogStore().Insert(ctx, logEntry); err != nil {
+		return fmt.Errorf("failed to insert request log for %s: %w", requestID, err)
+	}
```
behinddwalls, line 122: should this publish to log topic and let gateway manage all the writes to it as they should be written to gateway DB anyway..with this, they will get written to orcgstrator DB.

We can probably do a separate PR to fix it

behinddwalls (approved):

submitqueue/orchestrator/controller/dlq/buildsignal.go
```diff
@@ -0,0 +1,143 @@
...
+		logger:        logger.Named(name),
+		metricsScope:  scope.SubScope(name),
+		store:         store,
+		topicKey:      topicKey,
+		consumerGroup: consumerGroup,
+	}
+}
+
+// Process reconciles a single DLQ delivery for the buildsignal topic.
+func (c *buildSignalController) Process(ctx context.Context, delivery consumer.Delivery) (retErr error) {
```
behinddwalls, line 68: we probably want to add build topic as well...so if we fail to trigger builds say buildkite is unavailable then they can be transitioned to failed as well

submitqueue/orchestrator/controller/dlq/log.go
sbalabanov: Do you recommend to not do DLQ at all for it?

behinddwalls: For this one, i think we should be able republish back..if they failed due to mysql issues, if poison pills (ideally we shouldn't, given its free form) then we could mark it in the log itself with some generic event that we lost something... 

But from request state perspective, this likely has no meaning. Because to derive the request state we need this on gateway end all the times.

submitqueue/orchestrator/controller/dlq/dlq.go
sbalabanov: Request Log controller is in orchestrator right now. Gateway does not have queue-based controllers. If the controller is at the gateway, it probably makes sense to publish, otherwise I'd like to avoid republishing from dlq layer to primary layer.

submitqueue/orchestrator/controller/dlq/buildsignal.go
sbalabanov: Every new topic will need to have a corresponding dlq controller added... not part of this diff though?

submitqueue/orchestrator/controller/dlq/log.go
sbalabanov: Not sure I understood. DLQ controllers process those that regular controllers could not, possibly because of programmer's bug. 
The other option is to not have dlq log controller at all, monitor dlq,  and defer to the operator to fix any outstanding messages in it.
What I want to avoid is fixing things in DLQ controllers, i.e. republishing back to primary ones. Error processing layers should only transition one way.

submitqueue/orchestrator/controller/dlq/dlq.go
behinddwalls: https://github.com/uber/submitqueue/pull/205 made the change to move it to gateway

PR #193, opened 2026-06-04
feat(example): per-queue extension wiring; retire buildrunner/noop

behinddwalls:
### Why?

Demonstrate SubmitQueue's core premise — per-queue pluggable extensions —
in the example orchestrator, rather than wiring one implementation for
every queue, and make the fakes drivable end-to-end. Also document where
factory implementations and queue routing belong.

Stacked on the fakes PR, which adds the extension/*/fake packages this PR
wires in.

### What?

- Queue-major registry in example main.go: queueExtensions (one queue's
  full impl set) + queueRegistry (queue -> profile, with a default). Six
  thin Factory adapters resolve impls by Config.QueueName — all routing
  lives here in the wiring layer.
- Per-queue profiles: test-queue (heuristic over lines changed),
  e2e-test-queue (composite scorer, no-conflict), plus a new
  e2e-conflict-error-queue that always fails conflict analysis. Edge
  integrations + build runner share a baseline; scorer/analyzer are wrapped
  by the decorator fakes (score-error marker / failing predicate).
- Retire buildrunner/noop (subsumed by buildrunner/fake); switch the build
  controller unit test to buildfake.
- queues.yaml: register e2e-conflict-error-queue.
- CLAUDE.md: extensions hold contracts + impls only — factory
  implementations and routing live in the wiring layer; plus a TODO to
  evaluate promoting the registry into submitqueue/core later.

## Test Plan
- ✅ `make test` — 44/44 unit tests
- ✅ `make e2e-test` — 1/1
- ✅ `make integration-test` — 7/7
- ✅ `make check-gazelle` / `make check-tidy` — clean
- ✅ Live stack (`make local-submitqueue-start`): Ping OK; gateway rejects
  unknown queues; validated end-to-end — scorer passthrough (test-queue),
  scorer error (`?sq-fake=score-error` -> "fake: marked score error"), and
  conflict error (e2e-conflict-error-queue -> "fake: injected analyze
  error").

## Stack
1. #197
1. @ #193
1. #202

PR #195, opened 2026-06-04
refactor(change): persist typed change details from the change provider

behinddwalls:
### Why?

The change provider already produces rich per-URI facts (author, changed
files, line counts), but its value types lived in the extension layer and
the data was thrown away — validate fetched ChangeInfo only to log a file
count, and ChangeRecord stored an opaque Metadata JSON string that was never
written. Nothing downstream could read typed change facts.

### What?

- Move the change value types into entities: entity.User, entity.ChangedFile
  (now with LinesModified), entity.ChangeDetails (the facts), and
  entity.ChangeInfo (URI -> Details), with aggregation helpers. The
  changeprovider extension and GitHub impl now produce these.
- Replace ChangeRecord.Metadata (opaque string) with typed Details
  (ChangeDetails); the change table's metadata JSON column becomes details.
- Add ChangeStore.UpdateDetails — a version-guarded conditional write,
  following the optimistic-locking contract (arithmetic in the controller).
- validate now persists each fetched ChangeInfo onto the request's change
  records (per-URI, idempotent; ErrVersionMismatch is a benign no-op).

This is the producer half: typed details now exist and are persisted. The
score controller consumes them in a follow-up.

## Test Plan
- ✅ `make build`, `make test`, `make lint`, `make check-mocks/gazelle/tidy`
- ✅ `make integration-test` (storage contract suite round-trips Details and
  covers UpdateDetails create/update/version-mismatch)

## Stack
1. @ #195
1. #196
1. #197
1. #193
1. #202

### DIFF
--- a/submitqueue/entity/change_provider.go
+++ b/submitqueue/entity/change_provider.go
@@ -1,0 +1,76 @@
+// Copyright (c) 2025 Uber Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package entity
+
+// Author represents the author of a change.
+type Author struct {
+	// Name is the display name of the author.
+	Name string `json:"name"`
+	// Email is the email address of the author.
+	Email string `json:"email"`
+}
+
+// ChangedFile represents a single file modification in a change.
+type ChangedFile struct {
+	// Path is the file path relative to the repository root.
+	Path string `json:"path"`
+	// LinesAdded is the number of lines added in this file.
+	LinesAdded int `json:"lines_added"`
+	// LinesDeleted is the number of lines deleted in this file.
+	LinesDeleted int `json:"lines_deleted"`
+	// LinesModified is the number of lines modified in this file. Some providers
+	// (e.g. GitHub) report only additions and deletions and leave this zero.
+	LinesModified int `json:"lines_modified"`
+}
+
+// TotalLines returns the total number of lines touched in this file.
+func (f ChangedFile) TotalLines() int {
+	return f.LinesAdded + f.LinesDeleted + f.LinesModified
+}
+
+// ChangeDetails holds the provider-supplied facts about a single change (author,
+// modified files, line counts). It carries no identity — the owning URI lives on
+// ChangeInfo (provider correlation) and ChangeRecord (persisted claim).
+type ChangeDetails struct {
+	// Author is the author of the change.
+	Author Author `json:"author"`
+	// ChangedFiles is the list of files modified in this change. Order is unspecified.
+	ChangedFiles []ChangedFile `json:"changed_files,omitempty"`
+}
+
+// TotalLinesChanged returns the total number of lines touched across all files in the change.
+func (d ChangeDetails) TotalLinesChanged() int {
+	total := 0
+	for _, f := range d.ChangedFiles {
+		total += f.TotalLines()
+	}
+	return total
+}
+
+// FileCount returns the number of files touched in the change.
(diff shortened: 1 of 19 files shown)


submitqueue/entity/change_provider.go
```diff
@@ -0,0 +1,78 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package entity
+
+// User represents the author of a change.
+type User struct {
```
sbalabanov: User is very generic name. May be "ChangeAuthor" or "Author" ?

submitqueue/entity/change_provider.go
```diff
@@ -0,0 +1,78 @@
...
+	Name string `json:"name"`
+	// Email is the email address of the user.
+	Email string `json:"email"`
+}
+
+// ChangedFile represents a single file modification in a change.
+type ChangedFile struct {
+	// Path is the file path relative to the repository root.
+	Path string `json:"path"`
+	// Patch is the diff patch content for this file.
```
sbalabanov: Do we always store it? If not, may be give expectations.

submitqueue/extension/storage/mysql/change_store.go
```diff
@@ -75,13 +74,57 @@ func (s *changeStore) GetByURI(ctx context.Context, queue string, uri string) (r
...
+func (s *changeStore) Update(ctx context.Context, record entity.ChangeRecord, oldVersion int32, newVersion int32) (retErr error) {
+	op := metrics.Begin(s.scope, "update")
+	defer func() { op.Complete(retErr) }()
+
+	detailsJSON, err := marshalDetails(record.Details)
+	if err != nil {
+		return fmt.Errorf("failed to marshal details for change record queue=%s uri=%s request_id=%s: %w", record.Queue, record.URI, record.RequestID, err)
+	}
+
+	const query = "UPDATE `change` SET details = ?, updated_at = ?, version = ? WHERE queue = ? AND uri = ? AND request_id = ? AND version = ?"
```
sbalabanov: If details contain a patch, wouldn't it be expensive to rewrite it every time?

submitqueue/orchestrator/controller/validate/validate.go
```diff
@@ -254,11 +264,57 @@ func (c *Controller) publish(ctx context.Context, key consumer.TopicKey, request
 	return nil
 }
 
+// enrichChangeRecords persists each fetched ChangeInfo onto the matching change
+// record for this request. For each info it reads the record at (queue, info.URI,
+// request.ID) — the row start.claimURIs created — and conditionally writes the
+// details, bumping the version. A missing record or a lost version race surfaces
+// as ErrVersionMismatch, treated as a benign no-op so redelivery stays idempotent.
```
sbalabanov: why don't we make it immutable and fixed at Change info creation, or have separate entities and datastores for immutable/mutable fields?

sbalabanov (approved):

submitqueue/entity/change_provider.go
behinddwalls: Renamed `User` -> `Author` (type, the `ChangeDetails.Author` field, and the `json:"author"` tag).

submitqueue/entity/change_provider.go
behinddwalls: Removed `Patch` entirely — it was set by the GitHub provider but read by nothing downstream, so we no longer fetch or store it. (Easy to re-add, with clear expectations, once a consumer needs it.)

submitqueue/extension/storage/mysql/change_store.go
behinddwalls: Good catch — `Patch` is gone (unused), so it's no longer in the stored details. And records are now immutable (see the next thread): details are written once at claim time, never rewritten.

submitqueue/orchestrator/controller/validate/validate.go
behinddwalls: Done — the claim now happens in the validate controller after the provider fetch, so the change record is created once with its details in a single immutable Create. Removed the empty-claim-in-start + the enrich step, and dropped ChangeStore.Update entirely (records are immutable). Duplicate detection still reads other requests' claims before this request claims its own; per-queue partition leasing keeps the read-then-claim race-free.

PR #197, opened 2026-06-04
feat(extensions): fake implementations with error injection

behinddwalls:
### Why?

Extensions over external systems (changeprovider, buildrunner, pusher,
mergechecker) had no runnable stub to exercise their success — and
especially failure — paths without standing up the real backend
(GitHub/CI/git). scorer and conflict had deterministic stubs (heuristic,
composite, all, none) but no way to inject errors. These fakes let tests
drive both happy and error paths end-to-end from a land request.

### What?

- buildrunner/fake, changeprovider/fake, pusher/fake, mergechecker/fake:
  best-case by default; inject failures via an `sq-fake=<token>` marker in
  a change URI (e.g. build-fail, conflict, unmergeable, provider-error).
- scorer/fake, conflict/fake: decorators that wrap an existing impl (the
  "pick") and overlay error injection — scorer via the URI marker
  (score-error) over entity.BatchChanges, analyzer via a predicate, since
  Analyze sees batches, not change URIs.
- Each fake exposes only New(...) (decorator constructors for scorer/
  conflict). No factory implementations live in extension/* — those belong
  in the wiring layer.
- pusher/README.md documents the fake.

These packages are test/example stubs, never production. They are wired
into the example orchestrator in a follow-up PR.

## Test Plan
- ✅ `make test` — fake package tests
- ✅ `bazel build //...`

## Stack
1. @ #197
1. #193
1. #202


submitqueue/extension/buildrunner/fake/fake.go
```diff
@@ -0,0 +1,134 @@
...
+
+// fakeToken returns the marker token embedded in the first URI that carries
+// one, or "" if none do. The token ends at the first "&" or "#" delimiter.
+func fakeToken(uris []string) string {
+	for _, u := range uris {
+		if i := strings.Index(u, markerPrefix); i >= 0 {
+			rest := u[i+len(markerPrefix):]
+			if j := strings.IndexAny(rest, "&#"); j >= 0 {
+				rest = rest[:j]
+			}
```
JamyDev: Doesn't seem like your test cases exercise this. Probably good to add just to have an example.

submitqueue/extension/changeprovider/fake/fake.go
```diff
@@ -0,0 +1,80 @@
...
+		if i := strings.Index(u, markerPrefix); i >= 0 {
+			rest := u[i+len(markerPrefix):]
+			if j := strings.IndexAny(rest, "&#"); j >= 0 {
+				rest = rest[:j]
+			}
+			return rest
+		}
+	}
+	return ""
+}
```
JamyDev: Can we util this? It's duplicated across 5 different fakes..

JamyDev (changes requested):

submitqueue/extension/changeprovider/fake/fake.go
behinddwalls: Done — extracted the marker parsing into a shared `submitqueue/core/fakemarker` package (`Token([]string)` + `TokenInChanges([]entity.Change)`). All five fakes (changeprovider, buildrunner, mergechecker, pusher, scorer) now delegate to it instead of each carrying its own copy of `fakeToken`/`markerPrefix`. Put it under `core/` rather than `extension/` since it is shared infra, not an extension.

submitqueue/extension/buildrunner/fake/fake.go
behinddwalls: Added a buildrunner `Status` case with the marker among other query params (`...?ref=main&sq-fake=build-fail&attempt=2`) so the `&`/`#` trimming is exercised end-to-end. The trimming logic now lives in `core/fakemarker`, which has dedicated table tests covering the `&` delimiter, the `#` fragment, both together, and the marker-on-a-later-URI case (`fakemarker_test.go`).

submitqueue/core/fakemarker/fakemarker.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/uber/submitqueue/submitqueue/entity"
+)
+
+// Prefix introduces a marker token in a change URI: "sq-fake=<token>".
+const Prefix = "sq-fake="
+
+// Token returns the marker token embedded in the first URI that carries one, or
+// "" if none do. The token ends at the first "&" or "#" delimiter, so a marker
+// may sit among other query parameters or a fragment (e.g.
+// "github://o/r/pull/1/a?sq-fake=build-fail&attempt=2").
```
sbalabanov, line 35: should we build a generic augmentation (metadata) into a change URI language to be used by these fakes and similar?

submitqueue/core/fakemarker/fakemarker.go
```diff
@@ -0,0 +1,58 @@
...
+			}
+			return rest
+		}
+	}
+	return ""
+}
+
+// TokenInChanges returns the first marker token found across all changes' URIs,
+// or "" if none carry one.
+func TokenInChanges(changes []entity.Change) string {
```
sbalabanov, line 51: this seems to better belong to Change API

submitqueue/extension/buildrunner/fake/fake.go
```diff
@@ -0,0 +1,106 @@
...
+		return entity.BuildID{}, fmt.Errorf("fake: marked trigger error")
+	case tokenFail:
+		outcome = tokenFail
+	case tokenError:
+		outcome = tokenError
+	}
+
+	// Encode the outcome in the ID (e.g. "fake-build-fail-7") so Status is
+	// stateless. The counter just keeps IDs unique.
+	id := fmt.Sprintf("fake-%s-%d", outcome, r.counter.Add(1))
```
sbalabanov: is it ok to have a local counter? It violates the contract of uniquiness of  Build ID.

sbalabanov (approved):

submitqueue/core/fakemarker/fakemarker.go
behinddwalls: Let me think about it, i might take it up as separate change. The one thing this might be handy if & when we want to simulate through the actual logic but mock the outermost outbound calls...with fakes.

submitqueue/extension/buildrunner/fake/fake.go
behinddwalls: Good catch — fixed. Dropped the per-instance `atomic.Uint64` counter (two runner instances would each mint `fake-ok-1`, `fake-build-fail-1`, … violating the global BuildID uniqueness contract, which the package doc explicitly relies on for cross-process Trigger/Status). IDs now carry a `crypto/rand` hex suffix — `fake-<outcome>-<8hex>` — globally unique across instances and processes with no shared state. Hex digits can't spell the outcome tokens, so `Status`'s substring decode is unaffected. Added a cross-instance uniqueness assertion to `TestRunner_Trigger_UniqueIDs`. (f7b3eba)

submitqueue/core/fakemarker/fakemarker.go
behinddwalls: Kept this in the test-only `core/fakemarker` deliberately: moving marker parsing onto `entity.Change` would couple the production Change type to a test-only convention (every fake's doc is explicit that this is "never production"). The production change-URI parsers are also strict — `entity/github/change_id.go` requires a 40-char hex head SHA and `changeprovider/github/validate.go` rejects unexpected schemes — so a generic URI accessor on Change needs careful design to not collide with them. I captured the "promote a neutral accessor to the Change API" option in #209 as a follow-up rather than expand scope on this stub PR.

submitqueue/core/fakemarker/fakemarker.go
behinddwalls: Good idea, and worth doing properly rather than inline here — this PR is just test-only stub infra. Filed #209 to track a generic change-URI metadata/augmentation facility usable beyond the fakes, including the coexistence concerns with the strict github/phab change-ID parsers. Keeping the `sq-fake=` marker test-only for now.

behinddwalls:
Pushed an update (force-push, history rewritten):

- **Addressed @sbalabanov's BuildID-uniqueness comment**: replaced the per-instance `atomic.Uint64` counter with a `crypto/rand` suffix so IDs are globally unique across instances/processes (details in-thread).
- **Cleaned history**: folded the fix into the single feature commit and dropped the temporary `Merge main` commit, so the PR is back to one commit. Restacked #193/#202 on top.
- **Deferred** the "generic change-URI metadata language / move to Change API" discussion to #209 (rationale in-thread) — keeping the `sq-fake=` marker test-only here.

@JamyDev your two comments (shared `core/fakemarker` extraction + the buildrunner `Status` example exercising the marker among query params) were already addressed earlier; re-requested your review.

PR #202, opened 2026-06-05
refactor(conflict): take a batch of changes as analyzer input

behinddwalls:
### Why?

The conflict analyzer received entity.Batch values (request IDs only), so a real
analyzer — eventually backed by uber/tango ChangedTargetsAndEdges — could not see
the actual changes without storage lookups, which extensions must not do. The
scorer already solved this by moving to entity.BatchChanges; mirror it here.

### What?

- conflict.Analyzer.Analyze input changes from entity.Batch to entity.BatchChanges
  for both candidate and in-flight; the output ([]Conflict) is unchanged. No new
  entity types — a Tango-backed analyzer derives targets from the change URIs/files
  itself, with the target branch injected per queue at construction.
- New submitqueue/core/batchchanges.Collect: shared assembler that resolves a
  batch's request IDs -> change records -> entity.BatchChanges. The score and batch
  controllers both use it (extracted from the score controller).
- batch controller assembles candidate + in-flight BatchChanges before Analyze.
- all/none updated to the new input; conflict/fake switched to the
  sq-fake=conflict-error URI marker (drops FailOn/FailAlways); mock regenerated.
- example: conflictfake.New(all/none) without a predicate; removed the dedicated
  e2e-conflict-error-queue (the marker now reaches the analyzer on any queue).

## Test Plan
- ✅ make test (45/45 unit), make e2e-test (1/1), make integration-test (7/7)
- ✅ bazel build //...; check-gazelle / check-tidy / check-mocks clean once committed

## Stack
1. #197
1. #193
1. @ #202

### DIFF
--- a/example/submitqueue/gateway/server/queues.yaml
+++ b/example/submitqueue/gateway/server/queues.yaml
@@ -7,7 +7,3 @@ queues:
   - name: test-queue
   - name: e2e-test-queue
   - name: e2e-cancel-queue
-  # Routes to an analyzer that always errors (conflictfake.FailAlways) so e2e can
-  # exercise the conflict-analysis error path. See newQueueRegistry in the
-  # orchestrator example server.
-  - name: e2e-conflict-error-queue
--- a/example/submitqueue/orchestrator/server/main.go
+++ b/example/submitqueue/orchestrator/server/main.go
@@ -830,10 +830,9 @@ func newQueueRegistry(logger *zap.Logger, scope tally.Scope) (queueRegistry, err
 	//
 	// The scorer is wrapped by scorerfake so a change URI carrying
 	// "sq-fake=score-error" forces a scoring error end-to-end; it is a pure
-	// passthrough otherwise. The analyzer is wrapped by conflictfake with a nil
-	// predicate (passthrough) — swap the predicate (e.g. conflictfake.FailAlways)
-	// on a queue to exercise the analyzer error path, as e2e-conflict-error-queue
-	// below does.
+	// passthrough otherwise. The analyzer is wrapped by conflictfake so a change
+	// URI carrying "sq-fake=conflict-error" forces a conflict-analysis error;
+	// passthrough otherwise.
 	base := queueExtensions{
 		mergeChecker:   mc,
 		changeProvider: cp,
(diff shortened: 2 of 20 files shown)


sbalabanov (changes requested):
Analyzer API was specifically left lean and depending on IDs only with the expectation to use anything it needs including storage as a part of implementation.

This diff changes the expectation, trying to inject all of the data from the upstream eliminating the need for the implementation to cross-reference other extensions.
I think it may not work well. Quick reasons:
- Controller is technically not aware of what the implementation of Analyzer is. It will supply all the inputs unconditionally.
- Analyzer may not really use these inputs, or not use some parts of inputs. Some inputs may be fetched and not used. Some of them might be expensive to fetch or compute.
- Every future possible Analyzer implementation has to conform to only use the inputs defined for previous implementations (as they are a part of a common interface). If a future analyzer wants an additional data, the contract needs to be changed for all the analyzers + see previous issue for existing analyzers.

Let's brainstorm on this one before proceeding.

behinddwalls:
> Analyzer API was specifically left lean and depending on IDs only with the expectation to use anything it needs including storage as a part of implementation.
> 
> This diff changes the expectation, trying to inject all of the data from the upstream eliminating the need for the implementation to cross-reference other extensions. I think it may not work well. Quick reasons:
> 
> * Controller is technically not aware of what the implementation of Analyzer is. It will supply all the inputs unconditionally.
> * Analyzer may not really use these inputs, or not use some parts of inputs. Some inputs may be fetched and not used. Some of them might be expensive to fetch or compute.
> * Every future possible Analyzer implementation has to conform to only use the inputs defined for previous implementations (as they are a part of a common interface). If a future analyzer wants an additional data, the contract needs to be changed for all the analyzers + see previous issue for existing analyzers.
> 
> Let's brainstorm on this one before proceeding.

Sure,

I have been thinking about this and how much of extension (cross-dependency), we should allow or not at all... my first instinct is not allow cross extension aware-ness and keep extensions free of internal storage access or other extentions and have them be implemented solely on it's own. If we do allow cross-dependency, we need to figure out what could depend on each other and not end up in cyclic dependency.

behinddwalls:
Closing this PR for now until we decide the right contract. For now, I will use the existing mock for tests.

PR #205, opened 2026-06-05
refactor(request-log): gateway is sole writer of request log

behinddwalls:
### Why?

The request log had two persistence paths: the gateway wrote some entries
directly, while the orchestrator ran the `log`-topic consumer that wrote all
downstream entries to storage. Having the orchestrator persist the request log
blurs ownership — the orchestrator is a pipeline that should only emit events,
not own the request-log table. Concentrating all writes in the gateway gives a
single owner for the request log and keeps the orchestrator free of request-log
storage writes.

### What?

The request-log persistence consumer moves from the orchestrator to the gateway:

- Move `submitqueue/orchestrator/controller/log/` → `submitqueue/gateway/controller/log/`
  (importpath, doc comment, and default consumer group `orchestrator-log` →
  `gateway-log`). Logic is unchanged.
- Orchestrator: `TopicKeyLog` becomes publish-only (subscription dropped), the
  log controller registration and import are removed, controller count 11 → 10.
  It still publishes via `submitqueue/core/request.PublishLog`.
- Gateway: builds a consumer (generic + mysql classifiers), registers the moved
  log controller on `TopicKeyLog` with a subscription (group `gateway-log`),
  starts it, and drains it with `Stop(30000)` on shutdown — preserving the
  128+SIGTERM graceful-exit contract.
- Add `HOSTNAME=gateway-dev` to both gateway compose files for a stable
  subscriber name; update the workflow RFC and gateway README.
- Tests: add a gateway integration test that publishes to the log topic (as the
  orchestrator does) and asserts the gateway consumer persists it, and an e2e
  test that lands a request and asserts Status advances to `started` —
  exercising the publish→consume→persist path across both services.

The gateway keeps its two synchronous direct writes (`accepted` on Land,
`cancelling` on Cancel) for read-your-write visibility at RPC return. Both are
gateway writes, so the invariant holds: only the gateway persists the request
log; the orchestrator only publishes. This works because gateway and
orchestrator already share the same queue and app databases.

## Test Plan
- ✅ `bazel build` of both servers + the moved package
- ✅ `make test` — unit tests pass (incl. the moved log_test)
- ✅ `make check-gazelle`, `make check-tidy`, `make lint` (fmt + license)
- ✅ `make integration-test-submitqueue-gateway` — new `TestRequestLogConsumer`
  verifies the gateway consumer persists a log entry published to the log topic
- ✅ `make e2e-test` — new `TestLandRequest_PersistsStartedLogViaGatewayConsumer`
  verifies an orchestrator-published `started` log is persisted by the gateway
  and readable via Status; both services still exit 128+SIGTERM on shutdown

## Stack
1. @ #205
1. #213
1. #214


example/submitqueue/gateway/server/main.go
```diff
@@ -174,12 +178,29 @@ func run() error {
 		zap.String("queue_dsn", queueDSN),
 	)
 
-	// Build a publish-only topic registry: gateway only feeds the start of the
-	// orchestrator pipeline (TopicKeyStart) and the cancel topic (TopicKeyCancel).
-	// No subscription is configured because the gateway never consumes from the queue.
+	// Stable subscriber name for the log-topic consumer. Falls back to a
+	// time-seeded name when HOSTNAME is unset (e.g. local runs).
+	subscriberName := os.Getenv("HOSTNAME")
```
sbalabanov: what if we run multiple gateway instance on the same host

example/submitqueue/gateway/server/main.go
```diff
@@ -273,10 +320,25 @@ func run() error {
...
+	// Stop the consumer with a 30s timeout; by this time the context should be
+	// cancelled and the processing threads may already be exiting; recollect them.
+	errStop := logConsumer.Stop(30000)
+	if errStop != nil {
+		errStop = fmt.Errorf("failed to stop consumer: %w", errStop)
+	}
+
+	if errStop != nil || serverErr != nil {
+		// Override context cancellation error with the shutdown error
+		err = errors.Join(errStop, serverErr)
```
sbalabanov: should server_err go first?

submitqueue/gateway/README.md
```diff
@@ -1 +1,20 @@
-SubmitQueue Gateway
+# SubmitQueue Gateway
+
+The gateway is the RPC entry point to SubmitQueue. It accepts `Land`, `Cancel`,
+`Status`, and `Ping` calls, validates them at the edge, and hands work off to the
+orchestrator pipeline asynchronously via the message queue.
+
+## Request log ownership
+
+The gateway is the **sole writer of the request log**. No other service persists
```
sbalabanov: and technically, a sole reader, too

test/integration/submitqueue/gateway/suite_test.go
```diff
@@ -144,3 +151,47 @@ func (s *GatewayIntegrationSuite) TestLandAPI() {
...
+	require.NoError(t, err, "failed to create topic registry")
+
+	const sqid = "log-consumer-test/1"
+	logEntry := entity.NewRequestLog(sqid, entity.RequestStatusStarted, 1, "", nil)
+	require.NoError(t, corerequest.PublishLog(s.ctx, registry, logEntry, sqid),
+		"failed to publish request log to log topic")
+
+	s.log.Logf("Published 'started' log for sqid=%s; waiting for gateway consumer to persist it", sqid)
+
+	require.Eventually(t, func() bool {
```
sbalabanov, line 198: I think we do not do "Eventually()" with hard-coded timeouts even in integration tests?

sbalabanov (changes requested):
Eventually() in tests comment is the only blocking comment.
Side note: we should probably describe in a design doc the ownership of queues/databases, tiering of queues, and the flow of messages across queues.
Side note 2: dlq orchestrator controllers need to change to PublishLog()

example/submitqueue/gateway/server/main.go
behinddwalls: Good catch. Two gateway processes on one host would share `HOSTNAME` and thus the same partition-lease identity. Fixed in 43670be by making the subscriber name unique per instance (`hostname-pid`); offset tracking stays on the shared `gateway-log` consumer group.

example/submitqueue/gateway/server/main.go
behinddwalls: Agreed — the gRPC server error is the primary failure and the consumer-stop error is secondary cleanup. Swapped to `errors.Join(serverErr, errStop)` in 43670be.

submitqueue/gateway/README.md
behinddwalls: True. Updated the README in 43670be to call the gateway the sole *owner* — writer and reader — of the request log: `Status`/`Cancel` read it directly, and the orchestrator only publishes.

test/integration/submitqueue/gateway/suite_test.go
behinddwalls: The log consumer runs inside the gateway-service container, so this suite can only observe persistence black-box via the Status RPC — there's no in-process channel/HookSignal to wait on across the container boundary (unlike the messagequeue suite). In 43670be I extracted named constants (`persistTimeout`/`persistPollInterval`) and added a comment explaining the bounded poll is the deterministic-enough analog here. If you'd prefer, I can instead restructure the test to run the consumer in-process for a signal-based wait — happy to do that if you think it's worth the added complexity (it would need a distinct consumer group to avoid contending with the container's consumer).

albertywu (approved):

PR #208, opened 2026-06-05
feat(buildrunner): add GitHub Actions backend

albertywu:
## Summary

This PR adds a GitHub Actions implementation of BuildRunner.

They key detail to note is that starting with API version 2026-03-10, the workflow dispatch API synchronously returns a `workflow_run_id` as documented [here](https://docs.github.com/en/rest/actions/workflows?apiVersion=2026-03-10#create-a-workflow-dispatch-event). Prior versions did not return this workflow id.

With this `workflow_run_id`, we can poll for status and cancel workflow runs via API to implement BuildRunner.

## Validation

### DIFF
--- a/submitqueue/extension/buildrunner/README.md
+++ b/submitqueue/extension/buildrunner/README.md
@@ -13,7 +13,7 @@ See [`doc/rfc/submitqueue/build-runner.md`](../../../doc/rfc/submitqueue/build-r
 
 ## Backends
 
-- `noop`: local-development backend that immediately succeeds every build.
+- `fake`: local-development backend; every build succeeds unless a head change URI carries a failure marker (see `fake` package doc).
 - `githubactions`: proof-of-architecture backend that dispatches a GitHub
   Actions workflow. See [`githubactions/README.md`](githubactions/README.md)
   for the workflow inputs and example orchestrator environment variables.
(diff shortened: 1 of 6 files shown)


submitqueue/extension/buildrunner/README.md
```diff
@@ -10,3 +10,11 @@ See [`doc/rfc/submitqueue/build-runner.md`](../../../doc/rfc/submitqueue/build-r
 2. Map the `base` and `head` change slices onto the backend's build primitives (apply `base`, apply `head`, validate the result).
 3. Map the runner's lifecycle states down to the `BuildStatus` values: `Accepted` (accepted for execution), `Running` (executing), and the terminal `Succeeded` / `Failed` / `Cancelled`.
 4. Implement internal reconnect / retry so transient failures surface as plain errors without blocking the caller.
+
+## Backends
+
+- `noop`: local-development backend that immediately succeeds every build.
```
behinddwalls: this is removed, may need to update the text here

submitqueue/extension/buildrunner/githubactions/githubactions.go
```diff
@@ -0,0 +1,332 @@
...
+}
+
+type factory struct {
+	params FactoryParams
+}
+
+var _ buildrunner.Factory = (*factory)(nil)
+
+// NewFactory constructs a queue-aware GitHub Actions BuildRunner factory.
+func NewFactory(params FactoryParams) (buildrunner.Factory, error) {
```
behinddwalls: factory shouldn't be implemented here but rather be left for implementer/integrator

albertywu: Good call — removed `NewFactory`/`FactoryParams` from the `githubactions` package in 453829c. Per-queue factory wiring now lives in the integrator/wiring layer (matching the `buildkite` backend); the extension package keeps only the impl + `Params`. Reworked the corresponding test to construct the runner via `NewBuildRunner`.

submitqueue/extension/buildrunner/README.md
albertywu: Updated in 453829c — replaced the retired `noop` bullet with `fake` (the local-development backend; succeeds unless a head change URI carries a failure marker).

behinddwalls (approved):

PR #215, opened 2026-06-08
docs(rfc): stovepipe post-merge trunk-validation workflow

behinddwalls:
### Why?

SubmitQueue can no longer prove every change green before merge at the monorepo's current throughput, so it now merges directly to a single `main` that may be temporarily broken. Stovepipe is the post-merge service that validates `main`, records per-commit health, and drives recovery. There was no design doc describing its pipeline; this RFC fills that gap, mirroring `doc/rfc/submitqueue/workflow.md` so the two services read as siblings.

### What?

Adds `doc/rfc/stovepipe/workflow.md` describing the end-to-end pipeline: ingest trunk push events (external webhooks plus a fallback reconciliation poller, deduped on commit SHA) → start → validate → batch commits since the last known green → speculate / build / buildsignal → on green record `succeeded`; on failure bisect to the offending commit → invoke a pluggable remediation extension whose external backend lands a fix or revert via SQ.

Documents the three commit states (`unknown` / `succeeded` / `failed`), the SHA-as-identity / batch-as-validation-unit tracking model (bisect owns termination), and two gateway-owned sinks the orchestrator publishes to and the gateway consumes: `status` (the commit-status store callers query) and `log` (an append-only event log, the analogue of SQ's request log). Also covers fail-closed DLQ reconciliation, ownership by service, and the status/log ownership invariant. Links the new doc from `doc/rfc/index.md` under a new `## Stovepipe` section.

### DIFF
--- a/doc/rfc/stovepipe/workflow.md
+++ b/doc/rfc/stovepipe/workflow.md
@@ -1,0 +1,169 @@
+# Stovepipe Workflow
+
+Stovepipe is the post-merge trunk-validation service: it consumes a stream of commits pushed to the trunk (`main`), validates them in batches *after* they land, and records a per-commit health state that downstream systems gate on. It exists because SubmitQueue (SQ) can no longer afford to prove every change green *before* merge at the throughput the monorepo now sees — so SQ merges directly to a `main` that may be temporarily broken, and Stovepipe is the system that finds the breakages, names the offending commit, and drives recovery.
+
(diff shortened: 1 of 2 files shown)


doc/rfc/stovepipe/workflow.md
```diff
@@ -0,0 +1,169 @@
...
+   (poll)           │ poll backfills missed SHAs  ││  ┌──────────────────────┐    tooling query it
+                    └──────────────┬──────────────┘│  │ gateway: log         │
+                                   │ PushEvent     ├─►│ Append-only event log│
+                                   ▼               │  └──────────────────────┘
+                    ┌─ orchestrator ──────────────┐│
+                    │ start                       ├┘  orchestrator publishes status
+                    │ Record Commit (unknown) by  │   + log events (any stage)
+                    │ SHA; emit status + log      │
+                    └──────────────┬──────────────┘
+                                   │ SHA
```
mnoah1, line 55: Should be `ChangeInfo` entity here if we want to keep this agnostic to VCS

mnoah1 (approved):

PR #220, opened 2026-06-08
fix(speculate): gate finalize on the batch's own build result

albertywu:
## Problem

`speculate.tryFinalize` only consulted **dependency** batches, never the batch's **own** build. A batch whose build failed still passed finalize and merged, so a `sq-fake=build-fail` request reached `landed` instead of `error` (verified live: `build.status=failed`, `batch=succeeded`, `request=landed`).

## Fix

`tryFinalize` now fetches the batch's own build first:
- own build **failed** → fail the batch (CAS `Speculating → Failed` + publish `conclude`), via a shared `failBatch` helper also used by the existing dependency-failure path;
- own build **succeeded** → fall through to the existing dependency evaluation → merge;
- own build **not yet persisted / still in-flight** → no-op and wait for the next event.

Net merge condition is now **own build Succeeded AND all deps Succeeded**.

### Enabling change: `BuildStore.GetByBatchID`

`Build.ID` is the **runner-assigned** ID (e.g. `fake-build-fail-…`), not `batch.ID`, so a caller holding only a batch ID could not fetch its build via `Get`. Added `GetByBatchID(ctx, batchID) (entity.Build, error)` (interface + MySQL impl `WHERE batch_id = ? LIMIT 1`, indexed; regenerated mock).

This also repairs **`cancelBuild`**, which called `Get(ctx, batch.ID)` and was a silent no-op (always `ErrNotFound`); it now resolves the build by batch ID and updates it by build ID.

## Tests

### DIFF
--- a/submitqueue/extension/storage/mysql/schema/build.sql
+++ b/submitqueue/extension/storage/mysql/schema/build.sql
@@ -4,6 +4,5 @@ CREATE TABLE IF NOT EXISTS build (
     speculation_path JSON NOT NULL,
     score FLOAT NOT NULL,
     status VARCHAR(64) NOT NULL,
-    PRIMARY KEY (id),
-    UNIQUE KEY uniq_batch_id (batch_id)
+    PRIMARY KEY (id)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
--- a/submitqueue/extension/storage/mysql/schema/build.sql
+++ b/submitqueue/extension/storage/mysql/schema/build.sql
@@ -5,6 +5,5 @@ CREATE TABLE IF NOT EXISTS build (
     score FLOAT NOT NULL,
     status VARCHAR(64) NOT NULL,
     PRIMARY KEY (id),
-    -- Supports GetByBatchID: SELECT ... WHERE batch_id = ?
-    INDEX idx_batch_id (batch_id)
+    UNIQUE KEY uniq_batch_id (batch_id)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
--- a/submitqueue/extension/storage/mysql/schema/build.sql
+++ b/submitqueue/extension/storage/mysql/schema/build.sql
@@ -4,5 +4,7 @@ CREATE TABLE IF NOT EXISTS build (
     speculation_path JSON NOT NULL,
     score FLOAT NOT NULL,
     status VARCHAR(64) NOT NULL,
-    PRIMARY KEY (id)
+    PRIMARY KEY (id),
+    -- Supports GetByBatchID: SELECT ... WHERE batch_id = ?
+    INDEX idx_batch_id (batch_id)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
--- a/submitqueue/orchestrator/controller/speculate/speculate.go
+++ b/submitqueue/orchestrator/controller/speculate/speculate.go
@@ -170,8 +170,11 @@ func (c *Controller) startSpeculation(ctx context.Context, batch entity.Batch) e
 	return nil
 }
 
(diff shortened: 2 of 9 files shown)


submitqueue/extension/storage/mysql/schema/build.sql
```diff
@@ -4,5 +4,6 @@ CREATE TABLE IF NOT EXISTS build (
     speculation_path JSON NOT NULL,
     score FLOAT NOT NULL,
     status VARCHAR(64) NOT NULL,
-    PRIMARY KEY (id)
+    PRIMARY KEY (id),
+    UNIQUE KEY uniq_batch_id (batch_id)
```
sbalabanov: we should not do enforcements or secondary indexes on storage layer

submitqueue/orchestrator/controller/speculate/speculate.go
```diff
@@ -115,12 +115,13 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) (r
...
-	// between the terminal CAS and the dependent publish would otherwise
-	// leave them stuck waiting on a Cancelled dep.
+	// For terminal states that unblock downstream dependencies by being observed
+	// as terminal non-success, also re-publish to dependents. A crash between
+	// the terminal CAS and the dependent publish would otherwise leave them
+	// stuck waiting.
 	if batch.State.IsTerminal() {
 		metrics.NamedCounter(c.metricsScope, opName, "self_heal_terminal", 1)
-		if batch.State == entity.BatchStateCancelled {
+		if batch.State == entity.BatchStateCancelled || batch.State == entity.BatchStateFailed {
```
sbalabanov, line 124: if batch is successful, should we respeculate dependents too? i.e. drop speculation paths where this batch is predicted to fail.

test/e2e/submitqueue/suite_test.go
```diff
@@ -167,31 +167,77 @@ func (s *E2EIntegrationSuite) TestPingOrchestrator() {
...
+	assert.Equal(t, wantStatus, final)
+}
+
+// waitForStatus polls the gateway Status RPC until the request reaches a
+// terminal status (landed / error / cancelled) and returns it, or fails the
+// test if no terminal status is observed within timeout.
+func (s *E2EIntegrationSuite) waitForStatus(sqid string, timeout time.Duration) string {
+	t := s.T()
+	var final string
+	require.Eventually(t, func() bool {
```
sbalabanov: We have not come up with e2e framework really, so this test might be postponed. We likely do not want require.Eventually at all or have a framework around them to work with fat mocks, otherwise the test becomes flaky/fragile.

sbalabanov: @behinddwalls

sbalabanov (changes requested):

submitqueue/orchestrator/controller/speculate/speculate.go
albertywu: No re-speculation needed on success in the current naive model. Dependents already speculate assuming this batch passes — the chain is `Dependencies + [self]` with every in-flight build assumed to succeed — so a *success* changes none of their assumptions. Re-speculation is only needed on terminal **non-success** (Failed/Cancelled), so dependents can drop this batch from their chain; that's exactly the `Cancelled || Failed` branch here. Dropping speculation paths predicted to fail is the broader respeculation work tracked by the `tryFinalize` TODO.

submitqueue/extension/storage/mysql/schema/build.sql
albertywu: fixed

test/e2e/submitqueue/suite_test.go
albertywu: rm'd

PR #221, opened 2026-06-08
refactor(buildrunner): trigger on batches, resolve changes internally

behinddwalls:
## Summary
Change BuildRunner.Trigger to take batch identity — base []entity.Batch (the dependency batches) and head entity.Batch (the batch under test) — instead of controller-pre-resolved base/head []entity.Change, per the extension contract. Each implementation (buildkite, githubactions, fake) gains an injected changeset.Resolver and resolves the base and head batches' changes itself; the build controller drops its private collectChanges walk and loads the dependency batches as identity.

Status, Cancel, and the build id/status outputs are unchanged. The wiring injects the resolver into the fake build runner; the buildkite/githubactions Params gain a Resolver field.

Revises build-runner.md, which had deliberately kept batches out of the boundary — the base/head split survives, expressed as batch identity.

## Stack
1. @ #221
1. #222
1. #223
1. #227

### DIFF
--- a/submitqueue/extension/buildrunner/githubactions/githubactions.go
+++ b/submitqueue/extension/buildrunner/githubactions/githubactions.go
@@ -29,6 +29,7 @@ import (
 
 	"go.uber.org/zap"
 
+	"github.com/uber/submitqueue/submitqueue/core/changeset"
 	"github.com/uber/submitqueue/submitqueue/entity"
 	"github.com/uber/submitqueue/submitqueue/extension/buildrunner"
 )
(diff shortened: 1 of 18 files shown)


albertywu (approved):

submitqueue/extension/buildrunner/githubactions/githubactions.go
```diff
@@ -280,6 +295,19 @@ func flattenURIs(changes []entity.Change) []string {
...
+	for _, b := range batches {
+		cs, err := resolver.ChangesForBatch(ctx, b)
+		if err != nil {
+			return nil, err
+		}
+		changes = append(changes, cs...)
+	}
+	return changes, nil
+}
+
```
JamyDev: nit: extract this to avoid duplication between runners?

JamyDev (changes requested):

submitqueue/extension/buildrunner/githubactions/githubactions.go
behinddwalls: Done — extracted the shared resolution loop into `buildrunner.ResolveBatches` (new `extension/buildrunner/resolve.go`); both the Buildkite and GitHub Actions runners now call it instead of each holding a private copy.

PR #223, opened 2026-06-08
feat(conflict): target-overlap analyzer using changeset resolver

behinddwalls:
## Summary
Add submitqueue/extension/conflict/targetoverlap, a conflict.Analyzer that flags two batches as conflicting when they change a common file. It is the first analyzer to use the capability the extension contract unblocks: it takes only batch identity and resolves each batch's changed files itself through an injected changeset.Resolver, derived from each change's provider details.

ConflictTypeTargetOverlap was already named in the contract but had no implementation that could be written against an identity-only batch — this is that implementation. No change to the conflict.Analyzer interface. The example wires a target-overlap-queue to it.

## Stack
1. #221
1. #222
1. @ #223
1. #227

### DIFF
--- a/submitqueue/extension/conflict/fileoverlap/fileoverlap.go
+++ b/submitqueue/extension/conflict/fileoverlap/fileoverlap.go
@@ -1,0 +1,105 @@
+// Copyright (c) 2025 Uber Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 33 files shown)


submitqueue/extension/conflict/fileoverlap/fileoverlap.go
albertywu, line 1: Should this implementation be called `fileoverlap`? Since it seems to be detecting overlapping files, not targets. And `targetoverlap` a separate implementation that uses tango?

albertywu (approved):

submitqueue/extension/conflict/fileoverlap/fileoverlap.go
behinddwalls: seems like calude generated and published PR without me getting a change to review..but yes, it should be fileoverlap

behinddwalls: Renamed the package to `fileoverlap` — it detects overlapping changed files and reports `ConflictTypeTargetOverlap` (a shared file being the concrete notion of target overlap). A tango/build-target analyzer can land later as a separate `targetoverlap` implementation.

PR #225, opened 2026-06-09
docs: add gateway list api rfc

albertywu:
## Summary
Adds gateway List API rfc

Main idea here is the storage for list api is colocated / materialized at the same time request logs are upserted and the list api results should match up with what the status endpoint returns (both Status and List data is derived from immutable request logs stream and data is stored in gateway layer, not orchestrator).

## Stack
1. @ #225
1. #226

### DIFF
--- a/doc/rfc/submitqueue/list-api.md
+++ b/doc/rfc/submitqueue/list-api.md
@@ -1,0 +1,227 @@
+# Gateway List API
+
+Design notes for a gateway `List` API that powers a queue-scoped UX for
+observing SubmitQueue requests over a time window.
+
+This document captures **design decisions and rationale only**.
+
+## Problem
+
+Users need to inspect what happened in a queue during a time window: which
+requests were still running, which reached a terminal state, and what useful
+state each request is currently in. The existing `Status` API answers that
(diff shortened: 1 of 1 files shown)


JamyDev (approved):

doc/rfc/submitqueue/list-api.md
```diff
@@ -0,0 +1,227 @@
...
+
+The first retention target is 30 days after completion. Non-terminal requests
+must never be purged by age alone; a request that started 40 days ago and is
+still running must appear in a current overlap query.
+
+Terminal summaries and detailed logs can expire 30 days after completion.
+Detailed logs may have a separate policy later only if the UX no longer needs
+timeline/debug information for the same period.
+
+## Flow
```
behinddwalls, line 225: we need to add sorting as well? generally FIFO/time-based would be fine..but often we need to know or say what at the head of the queue and time-based listing may not reflect that clearly?

albertywu: Added a `Sorting` section above and how it would work. To start, we'd support sorting by admitted asc / dsc and later when BE supports queue-positioning info we can also add support for that.

behinddwalls (approved):

PR #228, opened 2026-06-09
refactor(consumer): lift queue consumer framework to be shared across submitqueue + stovepipe

mnoah1:
Move the shared queue consumer framework to top-level core/consumer so both SubmitQueue and Stovepipe can reuse it. Pipeline stage constants now live in per-domain submitqueue/core/topickey and stovepipe/core/topickey packages.

## Why?
Queue management will follow a similar framework for stovepipe, this is infra that can be shared.

## What?
Move core/consumer package up a level and update all reference and docs.  Topic keys move to their own package since they are domain specific

## Test Plan
- All checks below pass


sbalabanov (approved):
Do we want to have "global" core and "per service" core folders? Might be confusing.
May be rename the top one to something else.

behinddwalls:
> Do we want to have "global" core and "per service" core folders? Might be confusing. May be rename the top one to something else.

we could rename, right now core & entity are at global layer and also at per domain

behinddwalls (approved):

mnoah1:
Sure makes sense - maybe a `shared` or `infra` top level folder.  We already have some other stuff in the top level core, so I'll merge this first to move over the queue related items and then we can do separate bulk rename of the core folder.

PR #230, opened 2026-06-09
docs(rfc): add speculation design

behinddwalls:
## Summary

Capture how SubmitQueue speculates before the interfaces land. Speculation is modeled as two layers: **decision seams** — enumeration and **scoring** *describe* the tree of possible bets, selection and prioritization *act* on it — and **signal-driven limit policies** that decide *how much* to allow at each step (dependency limit, selection limit, prioritization limit), scaling with the build system's resources. A path's score is dynamic — recomputed as dependencies land, fail, or have their builds pass — not frozen at enumeration.

Covers: the problem (dependent batches serialize behind multi-minute builds), the Base/Head path-and-tree model, where speculation sits in the orchestrator pipeline, the status-vs-action ownership split, the dependency limit as an eligibility gate (batches wait for predecessors to land rather than trimming their base), dynamic rescoring, and how prioritization rations the queue's builds against its budget.

Design only; entities, store, and extension interfaces follow in later PRs in this stack. Adds doc/rfc/submitqueue/speculation.md and links it from the RFC index.

### Changes since review

Reframed from two seams (enumeration + selection) to describe/act decisions + limits, addressing @sbalabanov's review:
- selector resource-awareness → the new **prioritization** layer owns the queue-wide build budget; the selector stays a pure per-batch policy
- "variable depth, not a single knob" → the **dependency limit** is a signal-driven policy, not a constant
- clarified the dependency limit is an **eligibility gate** on active in-flight dependencies (a batch waits until predecessors land), not a base trim
- **dynamic scoring**: split scoring out of enumeration into a **scorer** that recomputes each path's score as dependencies land/fail/pass — enumeration is now structure-only
- entire tree persisted (the selector reads all path statuses + scores, so it can cancel a path it earlier asked to build); the redundant fallback path is cancelled on a winning bet; slash-free example names; trimmed impl-detail bullets
- recorded as open: extending rescoring to weigh how long a batch has waited

## Test Plan

Doc-only. Rendered locally; ASCII diagrams and tables verified; no hard-wrapped prose.

## Stack
1. @ #230
1. #231
1. #232


doc/rfc/submitqueue/speculation.md
```diff
@@ -0,0 +1,179 @@
...
+                    │  1. enumerate tree   2. persist tree        │              │
+                    │  3. select actions   4. enact + write status│◀──BatchID────┘
+                    └───┬────────────────────────────▲────────────┘  (a merge advances
+            per path    │ BatchID                    │ build result   the next batch)
+                        ▼                            │
+                      build ───Build───▶ buildsignal
+                  (trigger build)       (poll build status)
+```
+
+Each speculative path becomes its own build; build results flow back through `buildsignal` into `speculate`, which re-evaluates against the new reality. The controller is a **thin driver**: it trims the dependency set by speculation depth, asks the enumerator for the tree, persists it, reconciles each path's status from the latest builds and dependency states, asks the selector for actions, then enacts those actions (dispatching builds, cancelling, advancing to merge) and writes the resulting statuses back to the store.
```
sbalabanov: I think it is important to indicate that each speculation path only builds targets changed in a leaf node vs parent node, with the assumption that other speculation paths cover the rest of the chain.

doc/rfc/submitqueue/speculation.md
```diff
@@ -0,0 +1,179 @@
...
+                                                                          writes new status to store
+```
+
+### Enumeration
+
+Enumeration is deliberately **dumb**. Given a batch and a list of dependency batches, it mechanically lists the candidate paths and attaches a score to each. It is **pure and deterministic** — the same inputs always yield the same tree — so the controller can re-enumerate freely whenever the DAG changes, without the enumerator holding state. It does **not** decide what to build, it does **not** set status, and it does **not** decide how deep to speculate: it builds paths from exactly the dependency list it is handed (see [Speculation depth](#speculation-depth)). Keeping enumeration tractable for a very wide dependency list is its only real concern.
+
+Scores ride in on the inputs: each dependency is passed as a full `Batch`, which already carries the per-batch success probability (`Batch.Score`) computed by the score stage. The enumerator combines the scores of a path's base batches into the path's score; the combination formula is the enumerator's concern. No separate scoring backend or injected probability source is needed.
```
sbalabanov: I wonder if we ever want to rescore the batch depending on how long it stays in the queue without builds...

doc/rfc/submitqueue/speculation.md
```diff
@@ -0,0 +1,179 @@
...
+
+Why `selected` is distinct from `building`: the selector only *sends* a path to the build controller, which triggers a build **subject to resources** and may not start everything at once. So `Build` moves the path to `selected`; speculate does not assert `building` itself — it learns a build actually started only from a build signal, and only then records `building` and the `BuildID`. Between the two, the path is sent but unconfirmed, and the selector treats `selected` as "already sent — don't re-send, but still cancellable." "Base invalid" is not a status — it is one of the *triggers* that sends a path to `cancelled`.
```
sbalabanov: > which triggers a build **subject to resources** 
shouldn't selector be "resource-aware" too? If not, it may deliver insufficient build signals to the build controller not utilizing CI efficiently.

doc/rfc/submitqueue/speculation.md
```diff
@@ -0,0 +1,179 @@
...
+
+## Speculation depth
+
+A batch can sit deep in a **connected set** of dependencies. Speculating across the entire set is rarely worth it: the deeper a path's base, the less likely the whole prefix holds, and the more build resources each path burns. **Speculation depth** bounds how far back along the dependency chain speculation reaches.
```
sbalabanov: Don't we always want to speculate at least a happy path fully?
May be not, as in case of very long path even applying all the patches serially will take a lot of time

doc/rfc/submitqueue/speculation.md
```diff
@@ -0,0 +1,179 @@
...
+
+Depth is the **controller's** responsibility, not the enumerator's. The controller walks the connected set and trims the dependency list to those within `speculationDepth` *before* handing it to the enumerator. The enumerator then enumerates over exactly that limited set, unaware any trimming happened — which is why enumeration can stay dumb. Depth becomes a single knob, owned in one place, that every enumerator honors for free.
```
sbalabanov: I think we discussed variable speculation depth and not a single knob?

doc/rfc/submitqueue/speculation.md
```diff
@@ -0,0 +1,179 @@
...
+
+## Paths and trees, by example
+
+Consider queue `q` with batches `q/1`, `q/2`, `q/3`, where both `q/2` and `q/3` depend on `q/1` (and not on each other):
```
sbalabanov: nit: let's use dots or nothing and not slashes for names, otherwise diagram complicates because slashes are edges

doc/rfc/submitqueue/speculation.md
```diff
@@ -0,0 +1,179 @@
...
+
+```
+   q/2 tree (q/1 still building)             q/1 FAILS → q/2 tree (after controller reconciles)
+
+   Base   Head  Status      Score           Base   Head  Status      Score
+   [q/1]  q/2   building     0.27            [q/1]  q/2   cancelled   0.27   ← base broke
+   []     q/2   candidate    0.90            []     q/2   candidate   0.90   ← selector now returns Build
+```
+
+- **Bet holds** — `q/1` passes and merges: the build of `[q/1]→q/2` ran against exactly the tree `q/2` will land on, so `q/2` is mergeable and the controller finalizes it (publishes to `merge`) — no selector action involved. `q/1` and `q/2` were validated in parallel — the latency win.
```
sbalabanov: important that [] -> q2 should be dropped and build cancelled

doc/rfc/submitqueue/speculation.md
```diff
@@ -0,0 +1,179 @@
...
+
+**Selector** (`extension/speculation/selector`) — given a speculation tree (with each path's controller-stamped status), returns a per-path decision — the action (`Build` or `Cancel`) for the paths it chooses to act on. It reads status and emits actions only; it never writes status, and paths it leaves alone are simply omitted.
+
+Each package also defines a **`Factory`** following the repo's extension contract (see [extension-contract.md](extension-contract.md)): `For(Config) (Enumerator, error)` / `For(Config) (Selector, error)`, where `Config` carries only the queue's identity (`QueueName`). The system hands a factory nothing but the queue name; everything an implementation needs — including behavioral knobs like speculation depth, top-K, or build budget — is injected at construction by the integrator in the wiring layer, which resolves per-queue settings through `queueconfig`. The factory contract is defined here; concrete factories are deferred (see below).
```
sbalabanov: is this paragraph needed?

doc/rfc/submitqueue/speculation.md
```diff
@@ -0,0 +1,179 @@
...
+
+## Design decisions
+
+**Two seams, not one strategy.** A dumb enumerator lists and scores the possibilities; a separate selector — the policy — chooses what to do with each. *Rejected:* a single interface that both enumerates and picks. It conflates "what is possible" with "what we do this instant," forces the enumeration to re-run every time build reality shifts, and buries the one part worth tuning (the selection policy) inside the mechanical part. Splitting lets enumeration be pure and cacheable while the policy re-runs cheaply on every build signal.
```
sbalabanov: unclear to me how the selector knows if previous run of the selector set the path to "building" so the new run of the selector will set it to "cancel". We store all paths? Only actionable ones?

doc/rfc/submitqueue/speculation.md
```diff
@@ -0,0 +1,179 @@
...
+
+**A path is a Base/Head split, not a flat list.** Base is the assumed-good prefix; Head is the single batch under verification. *Why:* it maps one-to-one to the build stage (apply Base, validate Head), lets a build backend cache or short-circuit a shared base prefix across stacked speculations, and lets failure be attributed to base vs head. *Rejected:* a flat ordered list of batch IDs — the consumer would have to re-derive which portion is assumed-good versus under-test, discarding the one structural fact the orchestrator already knows.
+
+**Grouped, but two packages.** The enumerator and selector live under one `extension/speculation/` namespace because they are one cohesive subsystem that always travels together and shares the path/tree types. They stay separate packages — rather than one shared package — because their implementations are independent (an enumerator impl is not a selector impl), mirroring how `extension/scorer` and `extension/conflict` keep one interface per family.
```
sbalabanov: may be also not needed, implementation detail is not important

doc/rfc/submitqueue/speculation.md
behinddwalls: Good call — made this explicit in the Base/Head model: a path validates only the targets changed by its **Head** on top of the assumed-good **Base**, with each base batch's targets covered by *its own* path (so no path re-validates its base).

doc/rfc/submitqueue/speculation.md
behinddwalls: Clarified: the `Select` contract stays capacity-unaware (tree in → decisions out), but a concrete selector can be made capacity-aware via signals injected at its `Factory` (impl detail). Cross-batch arbitration stays at the build stage (admission control, in Deferred). Also walked back the stale "current capacity folded into status" line.

doc/rfc/submitqueue/speculation.md
behinddwalls: Right — added that the depth bound applies even to the fully-stacked happy path: on a very long chain, applying every predecessor serially is itself slow, so we don't always speculate it in full.

doc/rfc/submitqueue/speculation.md
behinddwalls: Agreed — reworded: depth is owned in one place but is **not** a single fixed constant. It's configured per queue (from `queueconfig`) and may be adaptive (CI pressure / other signals); the adaptive mechanism is left open — a dedicated extension, or possibly folded into enumeration.

doc/rfc/submitqueue/speculation.md
behinddwalls: Done — renamed `q/1`→`q1` etc. throughout the examples and diagrams so node names no longer collide with the DAG edges.

doc/rfc/submitqueue/speculation.md
behinddwalls: Added — on the bet holding and `[q1]→q2` merging, the now-redundant `[]→q2` fallback is dropped: the selector returns `Cancel` and the controller cancels any build still in flight.

doc/rfc/submitqueue/speculation.md
behinddwalls: Removed — the Deferred "Construction & per-queue config" bullet already covers it, and the signatures live in source.

doc/rfc/submitqueue/speculation.md
behinddwalls: Made the persistence model explicit: the controller persists the **entire** tree — every enumerated path with its current status, not just the actionable ones — so each `Select` run reads the up-to-date status of all paths (including `selected`/`building`) and can return `Cancel` for one it earlier asked to build.

doc/rfc/submitqueue/speculation.md
behinddwalls: Removed that bullet — packaging is an impl detail, agreed.

doc/rfc/submitqueue/speculation.md
behinddwalls: possible...but that could be another scoring..I have been wondering wether the current score stage even makes sense or should that be delayed or rescored in speculation again..we can chat more about this.

sbalabanov (changes requested):
sending it back as it is quite old, resubmit if needed

behinddwalls:
Reworked and resubmitting. Reframed the design around **three decisions** (enumeration, selection, prioritization) + **three signal-driven limit policies** (dependency / selection / prioritization limit) that scale with build resources.

Addresses all review comments:
- **Selector resource-awareness** — the new **prioritization** layer is the queue-wide enforcer that rations builds against the queue's budget; the selector stays a pure per-batch policy, with its `SelectionLimit` injected and called internally.
- **Variable depth, not one knob** — the **dependency limit** is a signal-driven policy, not a constant.
- **What depth means** — clarified it's an **eligibility gate** on active in-flight dependencies (a batch waits until predecessors land), *not* a base trim; added a gating-by-example.
- **Store all paths?** — yes, the controller persists the **entire** tree; the selector reads every path's status and can cancel one it earlier asked to build.
- Redundant fallback cancelled on a winning bet; slash-free example names; leaf-only target validation kept prominent; impl-detail bullets trimmed.
- **Open item recorded:** rescoring a batch the longer it waits queued.

PTAL @sbalabanov

doc/rfc/submitqueue/speculation.md
```diff
@@ -0,0 +1,233 @@
...
+
+**Speculation removes the wait by betting on the likely outcome.** Instead of waiting for predecessors to merge, the orchestrator *assumes* they will pass and builds a dependent batch now, on top of an assumed-good prefix of those predecessors. If the bet holds — the predecessors pass and merge — the dependent batch has already been validated against the exact tree it will land on, so it merges immediately. Builds for the whole chain run in parallel instead of in series.
+
+The bet can be wrong. If a predecessor fails, every build stacked on top of it is invalid and the orchestrator must **re-speculate**: discard the broken assumption and fall back to a path that survives (for example, build the dependent batch without the failed predecessor in its base). Because each predecessor is an independent "will it pass?" bet, a batch has *many* possible speculation paths. Enumerating them, choosing among them, and rationing them against finite build capacity — that is what this design is about.
+
+## Vocabulary
```
sbalabanov, line 17: may be vocabulary should not include implementation details given later (i.e. "written only by enumerator")

doc/rfc/submitqueue/speculation.md
```diff
@@ -0,0 +1,233 @@
...
+
+```
+   q2 tree (q1 still building)               q1 FAILS → q2 tree (after controller reconciles)
+
+   Base   Head  Status      Score           Base   Head  Status      Score
+   [q1]   q2    building     0.27            [q1]   q2    cancelled   0.27   ← base broke
+   []     q2    candidate    0.90            []     q2    candidate   0.90   ← selector now returns Build
+```
+
+- **Bet holds** — `q1` passes and merges: the build of `[q1]→q2` ran against exactly the tree `q2` will land on, so `q2` is mergeable and the controller finalizes it (publishes to `merge`) — no selector action involved. `q1` and `q2` were validated in parallel — the latency win. The now-redundant `[]→q2` fallback is dropped: the selector returns `Cancel` for it and the controller cancels any build still in flight.
```
sbalabanov: possibly interesting to indicate that "base -> q2"  spec path would be cancelled if "base -> q1" succeeds on its own

doc/rfc/submitqueue/speculation.md
```diff
@@ -0,0 +1,233 @@
...
+## Open questions
+
+Named here for context; not settled by this design:
+
+- **Rescoring while queued.** A batch's score is set once at the score stage. Should it be recomputed the longer the batch waits without a build — and if so, is that a scoring-stage concern, or something prioritization should weigh when ranking long-waiting builds? Open.
+- **Cross-queue capacity.** Prioritization here is queue-wide. If build capacity is ever shared *across* queues, a further arbitration layer is needed — with its own questions of fairness and starvation across queues, and score comparability between them.
+- **Concrete implementations.** The limit policies (single fixed value, resource-tracking, adaptive), enumerators (single-path, exhaustive, probability-ranked top-K), selectors (top-K, optimistic-first, shadow A/B), and prioritizers (priority-ordered consumption vs explicit admit-top-N, with or without preemption).
+- **Signal sourcing.** Which signals each limit policy weighs, and how they are surfaced (a capacity feed, historical metrics, config) to the wiring layer that constructs the policies.
+
+## Deferred
```
sbalabanov: no need for this section in my opinion as others do not define implementation much either

sbalabanov (approved):

PR #231, opened 2026-06-09
feat(entity,storage): rework speculation tree model and store

behinddwalls:
## Summary
Reshape the speculation tree data model around the Base/Head path that the build stage actually consumes, and align its store with the speculation design.

entity: SpeculationPath{Base, Head} becomes the unit; SpeculationPathInfo carries the path plus its Score, controller-owned Status (candidate/selected/prioritized/building/passed/failed/cancelling/cancelled), and BuildID — Path is immutable once persisted, Score/Status/BuildID are updateable by the controller. Score is scorer-computed and controller-persisted dynamic state — recomputed on every respeculate, not set at enumeration. Selected/Prioritized and Cancelling/Cancelled split desire vs budget-cleared and cancel-intent vs terminal, so the persisted status carries the cross-stage contract. Adds SpeculationPathAction (Promote/Cancel) and SpeculationPathDecision for the selector and prioritizer seams. SpeculationTree.Speculations becomes Paths, and the tree gains a Version for optimistic locking (starts at 1, incremented per change). The Build entity uses the shared SpeculationPath (Head = the batch under verification) and drops its own Score, which was never populated or read.

storage: SpeculationTreeStore.UpdateSpeculations(batchID, []SpeculationInfo) becomes a pure conditional write — Update(ctx, batchID, oldVersion, newVersion, paths) — returning ErrVersionMismatch when the persisted version does not match, per the repo's optimistic-locking pattern (version arithmetic owned by the controller). The MySQL impl treats rowsAffected == 1 as success, 0 as version mismatch, and anything else as a generic error. The MySQL column speculations is renamed paths and the schema gains a version column. MySQL impl, mock, and storage integration coverage (create/get round-trip, duplicate, versioned whole-tree overwrite, stale-write rejection, not-found) added/updated.

## Stack
1. @ #231
1. #315
1. #316
1. #317
1. #320

### DIFF
--- a/submitqueue/entity/speculation_tree.go
+++ b/submitqueue/entity/speculation_tree.go
@@ -14,38 +14,151 @@
 
 package entity
 
-// SpeculationPathAction defines the possible actions for a speculation path.
+// SpeculationPath is a single speculation path: an assumed-good prefix of
+// predecessor batches (Base) on top of which the batch under verification
+// (Head) is built and validated.
+//
+// This is the unit the build stage consumes: Base maps to the build runner's
+// base changes (an assumed-good prefix to apply) and Head maps to the changes
+// being validated.
+type SpeculationPath struct {
+	// Base is the ordered list of predecessor batch IDs assumed to have passed.
+	// Empty means the path builds the head batch directly on the target branch.
+	Base []string
+	// Head is the batch ID being verified by this path.
+	Head string
+}
+
+// SpeculationPathStatus is the observed lifecycle state of a speculation path.
+// It is written only by the orchestrator's speculate controller (into the
+// speculation tree store) and read by the decision seams (selector, prioritizer)
+// as input; the seams never write it.
+type SpeculationPathStatus string
+
+const (
+	// SpeculationPathStatusUnknown is the unreachable zero value, set by default
+	// on init. A persisted path always carries a real status (candidate onward),
+	// so this should never be seen in the store.
+	SpeculationPathStatusUnknown SpeculationPathStatus = ""
+	// SpeculationPathStatusCandidate is a freshly enumerated path the controller
+	// has persisted but not yet acted on.
+	SpeculationPathStatusCandidate SpeculationPathStatus = "candidate"
+	// SpeculationPathStatusSelected is a path the selector has promoted — a
+	// per-batch desire to build it — that the queue-wide prioritizer has not yet
+	// cleared. It is not sent to build while Selected: it waits on the build
+	// budget until it is prioritized (or dropped).
+	SpeculationPathStatusSelected SpeculationPathStatus = "selected"
+	// SpeculationPathStatusPrioritized is a path the prioritizer has admitted
+	// under the queue's build budget; it is cleared to run. The build effector
+	// triggers only Prioritized paths.
+	SpeculationPathStatusPrioritized SpeculationPathStatus = "prioritized"
+	// SpeculationPathStatusBuilding is a path a build signal has confirmed is in
+	// flight; its BuildID is known.
+	SpeculationPathStatusBuilding SpeculationPathStatus = "building"
+	// SpeculationPathStatusPassed is a path whose build succeeded.
+	SpeculationPathStatusPassed SpeculationPathStatus = "passed"
+	// SpeculationPathStatusFailed is a path whose build failed.
+	SpeculationPathStatusFailed SpeculationPathStatus = "failed"
+	// SpeculationPathStatusCancelling is a path whose in-flight build the
+	// controller has asked to stop but whose cancellation is not yet confirmed. It
+	// mirrors the batch-level cancelling intent (BatchStateCancelling): the cancel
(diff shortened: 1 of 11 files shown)


submitqueue/entity/speculation_tree.go
```diff
@@ -14,38 +14,137 @@
...
+// asks the controller to take for a path. It names the decision, not its effect:
+// it is ephemeral (recomputed every time a seam runs) and never persisted. The
+// controller maps it to a guarded SpeculationPathStatus transition and records
+// the result — the seams never write status themselves.
 type SpeculationPathAction string
 
 const (
-	// SpeculationPathActionUnknown is the default zero value for SpeculationPathAction.
+	// SpeculationPathActionUnknown is the unreachable zero value. A real decision
+	// always carries Promote or Cancel; a seam expresses "leave this path as-is"
```
sbalabanov: no need to mention Promote and Cancel in the description of Unknown

submitqueue/entity/speculation_tree.go
```diff
@@ -14,38 +14,137 @@
...
-// SpeculationInfo represents metadata about a single speculation path, including the path through the dependency graph, its current state, and the predicted build score.
-type SpeculationInfo struct {
-	// Path represents the speculation path; which is an ordered list of batches.
-	Path []string
-	// Action is a state that this path is in.
-	Action SpeculationPathAction
-	// Score is score for this speculation path.
+// SpeculationPathInfo is the per-path entry in a speculation tree: a path, its
+// latest predicted-success score, its controller-owned status, and a link to
+// the build dispatched for it (if any).
```
sbalabanov: probably good to indicate what is immutable and what isn't

submitqueue/extension/storage/mysql/speculation_tree_store.go
```diff
@@ -91,31 +91,32 @@ func (s *speculationTreeStore) Create(ctx context.Context, speculationTree entit
...
+		return fmt.Errorf("failed to update speculation tree for batchID=%q: %w", speculationTree.BatchID, err)
 	}
 
 	rowsAffected, err := result.RowsAffected()
 	if err != nil {
-		return fmt.Errorf("failed to get rows affected from update for batchID=%q: %w", batchID, err)
+		return fmt.Errorf("failed to get rows affected from update for batchID=%q: %w", speculationTree.BatchID, err)
 	}
 
 	if rowsAffected != 1 {
```
sbalabanov: we do it all over places, but better design is to rowsAffected == 1 -> success, rowsAffected ==0 -> errNotFound, otherwise generic infra error.

submitqueue/extension/storage/speculation_tree_store.go
```diff
@@ -32,7 +32,7 @@ type SpeculationTreeStore interface {
 	// Returns ErrAlreadyExists if the entry already exists.
 	Create(ctx context.Context, speculationTree entity.SpeculationTree) error
 
-	// UpdateSpeculations updates the speculations of a speculation tree.
-	// Returns ErrNotFound if the speculation tree is not found.
-	UpdateSpeculations(ctx context.Context, batchID string, speculations []entity.SpeculationInfo) error
+	// Update overwrites the paths of an existing speculation tree, identified by
+	// speculationTree.BatchID. Returns ErrNotFound if the speculation tree is not found.
```
sbalabanov: I think it needs concurrency control / optimistic locking as well.

sbalabanov (changes requested):
concurrency comment is critical, others aren't

submitqueue/entity/speculation_tree.go
behinddwalls: Done — the Unknown doc no longer names the verbs; it just says a seam expresses "leave as-is" by omitting the path from its decisions.

submitqueue/entity/speculation_tree.go
behinddwalls: Done — SpeculationPathInfo now marks Path as immutable and Score/Status/BuildID as updateable (controller-written under the tree's version lock), and SpeculationTree marks BatchID immutable / Paths updateable.

submitqueue/extension/storage/mysql/speculation_tree_store.go
behinddwalls: Done — Update now does the three-way split: rowsAffected 1 → success, 0 → ErrVersionMismatch (with the version guard in the WHERE clause, 0 rows means stale version or missing row — matching the batch/request store convention), anything else → generic error.

submitqueue/extension/storage/speculation_tree_store.go
behinddwalls: Done — SpeculationTree now carries a Version (starts at 1), and Update is a pure conditional write per the storage README pattern: Update(ctx, batchID, oldVersion, newVersion, paths), returning ErrVersionMismatch on a stale version. Schema gains a version column; the integration suite covers the versioned overwrite and stale-write rejection against real MySQL.

PR #240, opened 2026-06-12
refactor(entity): promote Change + ChangeID to shared entity/change

behinddwalls:
### Why?

`Change` (a code change identified by provider URIs) and the GitHub/Phabricator `ChangeID` parsers are pure identity-and-parsing types with no domain semantics and no persisted-as-shared schema. They are the one cluster genuinely common to every code-change pipeline — SubmitQueue today, and Stovepipe / Runway / Tango alignment going forward — so they belong in the shared top-level `entity/` tree rather than inside `submitqueue/entity/`.

This is deliberately the *only* promotion: facts a domain persists (the change-provider model embedded in `ChangeRecord`), state machines, and pipeline types stay domain-local, since sharing them would couple every domain's storage schema and bend their state machines toward SubmitQueue's.

### What?

Moves the canonical `Change` struct out of `submitqueue/entity/request.go` into a new `entity/change` package, and relocates the parser packages to `entity/change/github` and `entity/change/phabricator` (preserving git history). No alias shim: every consumer now references `change.Change` and imports the new path directly. JSON tags are unchanged, so queue payloads (`Request`, `LandRequest`, `PushResult`) remain byte-identical across a deploy.

`submitqueue/entity` keeps its full pipeline and facts; `Request`, `LandRequest`, and `ChangeOutcome` now embed `change.Change`.

## Test Plan

- ✅ `make build`
- ✅ `make test` (51/51 pass)
- ✅ `make fmt` / `make gazelle` (converged, no further changes)
- ✅ `make check-tidy` (go.mod / MODULE.bazel unchanged — internal move, no new deps)


entity/change/change.go
```diff
@@ -0,0 +1,32 @@
...
+// limitations under the License.
+
+// Package change holds the shared code-change identity used across SubmitQueue,
+// Stovepipe, and other repo-local domains. A Change names code to act on via
+// provider URIs; the URI parsers live in the github and phabricator subpackages.
+package change
+
+// Change represents a code change identified by URIs from a code change provider (e.g., GitHub Pull Request, Phabricator Diff).
+// The provider is extracted from the URI scheme. The object is immutable after creation.
+type Change struct {
```
mnoah1, line 22: When we say code change here - will we also consider merged changes as a a "Change" under this interface?
- we were proposing to do this under a git:// schema
- definitely makes sense that we could have a git:// uri to a branch that the user wants merged
- but does it also make sense that a git:// uri pointing to a change is also a change - or is that a separate "stovepipe only" concept?

mnoah1 (approved):

sbalabanov (approved):

entity/change/change.go
behinddwalls: Yes, they will be represented by this change entity itself.

PR #241, opened 2026-06-12
Adding Phabricator ChangeProvider

ubettigole:
## Why?
In order to get diff information for Phabricator ChangeIDs, we must implement a Phabricator ChangeProvider.

## What?
This change creates a Phabricator ChangeProvider, similar to the existing GitHub ChangeProvider. Like the GitHub ChangeProvider, it:
- Retrieves diff information in batches (currently set at 10 diffs per batch)
- Converts responses into a `[]ChangeInfo`

## Test Plan
I created a temporary unit test (removed before committing) that pointed the Phabricator ChangeProvider at Uber's internal Phabricator instance, and confirmed that we could retrieve diff information by passing a bearer token, similar to the way we initialize the GitHub ChangeProvider in the example server.

All unit tests pass, with 99% coverage in the new `phabricator` directory.


submitqueue/extension/changeprovider/phabricator/conduit.go
```diff
@@ -0,0 +1,101 @@
...
+
+// buildQueryDiffsRequest builds query parameters for a differential.querydiffs call.
+func buildQueryDiffsRequest(diffIDs []int, apiToken string) url.Values {
+	form := url.Values{}
+	for _, id := range diffIDs {
+		form.Add("ids[]", strconv.Itoa(id))
+	}
+	if apiToken != "" {
+		form.Set("api.token", apiToken)
+	}
```
behinddwalls: wondering if we should always expect it? if so, do we need empty check? or say can we error if empty instead?

submitqueue/extension/changeprovider/phabricator/conduit.go
```diff
@@ -0,0 +1,101 @@
...
+
+	return resp, nil
+}
+
+// parseConduitResponse reads the HTTP response, unwraps the Conduit response,
+// and extracts the diffResults for the requested diff IDs.
+func parseConduitResponse(resp *http.Response, diffIDs []int) (map[int]*diffResult, error) {
+	if resp.StatusCode != http.StatusOK {
+		body, _ := io.ReadAll(resp.Body)
```
behinddwalls: can we handle err instead of swallowing it?

submitqueue/extension/changeprovider/phabricator/conduit.go
ubettigole: Removed this completely, since we handle setting the API token in the internal ChangeProvider

submitqueue/extension/changeprovider/phabricator/conduit.go
ubettigole: Done

submitqueue/extension/changeprovider/phabricator/convert.go
```diff
@@ -0,0 +1,41 @@
...
+func convertFiles(changes []fileChange) []entity.ChangedFile {
+	files := make([]entity.ChangedFile, 0, len(changes))
+	for _, c := range changes {
+		added, _ := strconv.Atoi(c.AddLines)
+		deleted, _ := strconv.Atoi(c.DelLines)
+		files = append(files, entity.ChangedFile{
+			Path:          c.CurrentPath,
+			LinesAdded:    added,
+			LinesDeleted:  deleted,
+			LinesModified: added + deleted,
```
JamyDev: Given what we said about this not being available in Github, should we keep this 0 also?

JamyDev (approved):
minor nit

submitqueue/extension/changeprovider/phabricator/convert.go
ubettigole: Good call - no longer populating LinesModified.

behinddwalls (approved):

PR #242, opened 2026-06-15
feat(entity): add git:// ChangeID parser

mnoah1:
Adds entity/change/git, a git-scheme change URI parser that mirrors the github provider (similar field/structure naming), but is used to refer to a change based on ref + commit instead of by pull request.

The ref segment carries refs/heads/<branch> or refs/tags/<tag>, percent- encoded in case of slashes in a branch name.

Move existing SHA check into a shared entity/change/changeutil.IsFullHex(s, length) helper now used by both the github and git providers, with the expected length passed in per provider.

## Why?
Stovepipe validation will act on changes that are already merged, and referenced by branch + SHA instead of pull request or phabricator revision.

## What?
git:// schema definition

## Test Plan
- Unit tests
- Will use this in subsequent changes to build out stovepipe workflow


entity/change/git/change_id.go
```diff
@@ -0,0 +1,133 @@
...
+// scheme is the canonical URI scheme for git-backed change identifiers.
+const scheme = "git"
+
+// refPrefix is the namespace every fully-qualified git ref lives under
+// (refs/heads/<branch>, refs/tags/<tag>, ...). Disambiguates between branches and tags
+// of the same name.
+const refPrefix = "refs/"
+
+// changeIDFormat is the expected format for change IDs, included in error messages.
+const changeIDFormat = "git://{owner}/{repo}/{ref}/{head_commit_sha}"
```
behinddwalls:
```suggestion
const changeIDFormat = "git://{remote}/{repo}/{refspec}/{commit_sha}"
```

basically something like `git://git.remote.com/<repo>/<refspec>/<commit_sha>`? how do we handle port? ideally all URLs support port usage

entity/change/git/change_id.go
```diff
@@ -0,0 +1,133 @@
...
+// ChangeID represents a parsed git:// change identifier.
+// Format: git://{owner}/{repo}/{ref}/{head_commit_sha}
+//
+// Ref is a fully-qualified, percent-encoded git ref so that branches, tags, and
+// ref names containing slashes all fit a single path segment unambiguously.
+type ChangeID struct {
+	// Scheme captures the URI scheme (always "git" in current implementation).
+	Scheme string
+	// Org is the organization or owner of the repository.
+	Org string
```
behinddwalls: org is github concept and not git, rather we need to know about git remote in this case

entity/change/git/change_id.go
```diff
@@ -0,0 +1,133 @@
...
+	// Repo is the repository name.
+	Repo string
+	// Ref is the fully-qualified git ref the change landed on, decoded from the
+	// URI (e.g. "refs/heads/main", "refs/tags/v1.0").
+	Ref string
+	// HeadCommitSHA is the head commit SHA that pins the landed change.
+	HeadCommitSHA string
```
behinddwalls: doesn't need to be HEAD commit but rather any commit that Ref points at...in history...it's mostly a way to say point in time ref was at this commit

entity/change/git/change_id.go
mnoah1: Originally I was thinking the remote should be configured somewhere else as part of the service config.  Example: changing where we host a repo, but not changing the contents - and that change IDs should be stable across that case.  Updated to include remote here.

entity/change/git/change_id.go
mnoah1: With my latest update, port can be part of the remote URL.

mnoah1: Any reason we would want refspec over ref?  Ref feels less ambiguous.

behinddwalls (approved):

PR #243, opened 2026-06-15
refactor(entity): promote RequestLandStrategy to shared entity/mergestrategy

behinddwalls:
### Why?

The land-merge strategy enum lived in `submitqueue/entity` as `RequestLandStrategy`, but the upcoming runway merge-conflict contract needs the same type, and runway must not import `submitqueue/entity`. Mirroring the `Change` promotion in #240, the strategy belongs in the shared top-level `entity/` so both domains reference one type with no per-domain enum or mapping.

### What?

Moves `RequestLandStrategy` and its constants out of `submitqueue/entity/request.go` into a new top-level `entity/mergestrategy` package as `MergeStrategy` (`MergeStrategyUnknown/Rebase/SquashRebase/Merge`). `Request.LandStrategy` now has type `mergestrategy.MergeStrategy`. All references across the gateway and orchestrator controllers, entities, and tests are updated to the shared type. Pure mechanical refactor — no behavior change.

## Test Plan
✅ `bazel build //...`
✅ `bazel test //... --test_tag_filters=-integration,-e2e` (51 tests pass)
✅ `make gazelle` clean

## Stack
1. @ #243
1. #244
1. #245
1. #247

### DIFF
--- a/entity/mergestrategy/BUILD.bazel
+++ b/entity/mergestrategy/BUILD.bazel
@@ -1,0 +1,8 @@
+load("@rules_go//go:def.bzl", "go_library")
+
+go_library(
+    name = "mergestrategy",
+    srcs = ["mergestrategy.go"],
+    importpath = "github.com/uber/submitqueue/entity/mergestrategy",
+    visibility = ["//visibility:public"],
+)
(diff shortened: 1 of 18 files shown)


sbalabanov (approved):
I still think we need to have a single entity folder for all projects to avoid diffs like this.
At this point it is still possible to untangle dependencies during the move - an still the diff is quite big for what it is.
In the future it will be more complicated.

PR #245, opened 2026-06-15
feat(orchestrator): make merge-conflict check asynchronous via runway

behinddwalls:
### Why?

The merge-conflict check ran synchronously inside the `validate` consumer by calling the `mergechecker` extension inline. A real merge attempt is slow and I/O-heavy, so doing it on the partition lease blocks the pipeline and couples SubmitQueue to the checker's latency. This moves the check to an asynchronous round-trip with runway, modelled on `build`/`buildsignal` but across a service boundary.

### What?

The pipeline gains `validate ⇢ (runway) ⇢ mergeconflictsignal → batch`:

- `validate` drops the inline `mergechecker` call and, after its existing dedup + change-metadata + claim work, publishes the full `MergeRequest` to the runway-owned `merge-conflict-checker` queue, keyed by the request id as the client-owned correlation id. The id round-trips, so the result correlates straight back to the request (unlike `build`, whose server-generated id needs a mapping store). The hand-off to runway is retryable.
- `mergeconflictsignal` (new) consumes runway's `MergeResult` off `merge-conflict-checker-signal`, advances the request to `batch` when mergeable, or fails it when conflicted.
- DLQ reconcilers drive the request to `Error` on dead-letter; the signal DLQ reads the request id straight off the result.

The check is triggered directly from `validate`, not a standalone `mergeconflict` trigger stage: that hop only forwarded the request id before publishing to runway, so folding it into `validate` saves a queue round-trip and removes a stage (its topic key, consumer, and DLQ entry) with no change to the cross-boundary contract.

Crossing the runway boundary is why these payloads carry full data rather than entity IDs; the queue-payload-boundary rule is documented in CLAUDE.md, with the pipeline diagram and stage table updated in workflow.md and the superseded `mergechecker` validate-path row noted in extension-contract.md. The `mergechecker` package is left in-tree (unused on the validate path); removing it is a follow-up. Runway's service implementation is out of scope — only its contract (added in #260) is consumed here.

## Test Plan
- ✅ `bazel build //...`
- ✅ `bazel test //... --test_tag_filters=-integration,-e2e` (56 tests pass)
- ✅ `make gazelle` clean

## Stack
1. #264
1. #259
1. #260
1. @ #245
1. #247


doc/rfc/submitqueue/extension-contract.md
```diff
@@ -43,6 +43,8 @@ Two facts this grounds: `conflict` already resolves nothing (the baseline), and
...
 
+> **Note (superseded for the validate path).** The `mergechecker.MergeChecker` row above describes the *synchronous*, in-process mergeability check `validate` used to run. That check is now **asynchronous and out-of-process**: `validate` hands off to the `mergeconflict` controller, which publishes a full check request to the runway-owned `merge-conflict-checker` queue; runway performs the merge attempt and returns a result that `mergeconflictsignal` consumes (see [workflow.md](workflow.md)). The `mergechecker` extension package is kept in-tree but unused on the validate path; removing it is a follow-up. The row is retained here as the historical contract.
```
behinddwalls: lets update the RFC to reflect the new contract rather than adding a note here

submitqueue/core/topickey/topickey.go
```diff
@@ -27,6 +27,11 @@ const (
 	TopicKeyCancel TopicKey = "cancel"
 	// TopicKeyValidate is the pipeline stage where requests are published for validation.
 	TopicKeyValidate TopicKey = "validate"
+	// TopicKeyMergeConflict is the internal pipeline stage where validated
+	// requests are published to start an asynchronous merge-conflict check. The
+	// mergeconflict controller consumes it, records the check, and publishes the
+	// full request to runway's merge-conflict-checker queue.
+	TopicKeyMergeConflict TopicKey = "mergeconflict"
```
behinddwalls: records the check? we removed storage layer so no recording anymore, request itself captures the state for it.

submitqueue/orchestrator/controller/mergeconflictsignal/mergeconflictsignal.go
```diff
@@ -0,0 +1,178 @@
...
+
+	if !result.Mergeable {
+		metrics.NamedCounter(c.metricsScope, opName, "not_mergeable", 1)
+		c.logger.Infow("request not mergeable",
+			"request_id", request.ID,
+			"reason", result.Reason,
+		)
+		// User error: reject to DLQ, where the request is reconciled to Error.
+		return errs.NewUserError(fmt.Errorf("request %s is not mergeable: %s", request.ID, result.Reason))
+	}
```
behinddwalls: why reject to DLQ when we know the request has failed and we should just mark it failed?

doc/rfc/submitqueue/extension-contract.md
```diff
@@ -35,13 +35,14 @@ Two facts this grounds: `conflict` already resolves nothing (the baseline), and
...
+**Outputs are unchanged except `pusher`.** This RFC moves the *input* toward identity; four of the five return contracts — conflicts, score, change info, build id/status — are exactly what they are today. `pusher` is the lone exception: because its input becomes a *list* of independently-landed batches, its result regroups per batch (`BatchID`-tagged, per-change commit detail kept underneath) so each batch's outcome stays correlatable — the "output mirrors the input unit" principle above. No other output shape changes.
+
+The validate-time mergeability check is no longer in this catalog because it is no longer an in-process extension. It now runs **asynchronously and out-of-process** in runway: `validate` hands off to the `mergeconflict` controller, which publishes a full check request to the runway-owned `merge-conflict-checker` queue, and `mergeconflictsignal` consumes runway's result (see [workflow.md](workflow.md)). The `mergechecker` package stays in-tree but unused on the validate path; removing it is a follow-up.
```
sbalabanov: ask AI to avoid history in documentation i.e. "no longer in this catalog"

doc/rfc/submitqueue/workflow.md
```diff
@@ -1,6 +1,6 @@
...
+The orchestrator processes land requests through a queue-driven pipeline of small, single-purpose controllers. The gateway accepts a request over RPC and hands it off asynchronously; from there each controller consumes one topic, advances the request or batch, and publishes to the next topic. Most hops carry only an ID — the controller fetches the entity from storage — while a few entry points (`start`, `buildsignal`, `log`) carry the full payload because there is no row to fetch yet. The merge-conflict check is the one stage that crosses a service boundary: `mergeconflict` publishes a full payload to the runway-owned `merge-conflict-checker` queue and `mergeconflictsignal` consumes a full payload back off `merge-conflict-checker-signal`, because runway cannot read submitqueue's storage. See the queue-payload-boundary rule in [CLAUDE.md](../../../CLAUDE.md).
```
sbalabanov: merge-conflict is not the only one, so it should not be mentioned explicitly

doc/rfc/submitqueue/workflow.md
```diff
@@ -85,7 +105,7 @@ The pipeline has two cycles: `speculate → build → buildsignal → speculate`
...
+The DLQ controllers do not re-attempt the failed work. They decode the payload to recover the affected `RequestID` (start, validate, mergeconflict, batch, cancel) or `BatchID` (score, speculate, build, buildsignal, merge, conclude) and drive the entity to a terminal failed state — `RequestStateError` for requests, `BatchStateFailed` for batches with fan-out to the member requests. The `mergeconflictsignal` DLQ decodes a runway `MergeResult` whose `id` is the request id echoed back, so it fails that request directly. State writes use the same optimistic-locking CAS as the primary pipeline, so a late primary-pipeline update wins cleanly and a version mismatch is asked back for redelivery.
```
sbalabanov: same, avoid specifics unless only as an example

submitqueue/orchestrator/controller/dlq/mergeconflictsignal.go
```diff
@@ -0,0 +1,113 @@
...
+	defer func() { op.Complete(retErr) }()
+
+	msg := delivery.Message()
+
+	result, err := runwayentity.MergeResultFromBytes(msg.Payload)
+	if err != nil {
+		metrics.NamedCounter(c.metricsScope, opName, "deserialize_errors", 1)
+		return fmt.Errorf("failed to decode merge conflict check result from dlq payload: %w", err)
+	}
+	if result.ID == "" {
```
sbalabanov: do we really need to check for data correctness if we control both producer and consumer?

submitqueue/orchestrator/controller/mergeconflict/mergeconflict.go
```diff
@@ -0,0 +1,193 @@
...
+		)
+		return nil
+	}
+
+	// Build the full payload runway needs to attempt the merge. The request id is
+	// the client-owned correlation id, so a redelivery republishes the same id and
+	// runway dedupes on it; the result is matched straight back to the request. At
+	// validate time the check is a single step (candidate vs target branch); a
+	// future in-flight caller layers earlier steps before the candidate.
+	req := runwayentity.MergeRequest{
```
sbalabanov: Do we even need mergeconflict controller now if the only thing it does is relaying a message? Why not do it in a prev controller in a pipeline?

submitqueue/orchestrator/controller/mergeconflict/mergeconflict.go
```diff
@@ -0,0 +1,193 @@
...
+	"context"
+	"fmt"
+
+	"github.com/uber-go/tally"
+	"github.com/uber/submitqueue/core/consumer"
+	"github.com/uber/submitqueue/core/errs"
+	"github.com/uber/submitqueue/core/metrics"
+	"github.com/uber/submitqueue/entity/change"
+	entityqueue "github.com/uber/submitqueue/entity/messagequeue"
+	runwayentity "github.com/uber/submitqueue/runway/entity"
```
sbalabanov: sq should not reference runway

submitqueue/orchestrator/controller/mergeconflictsignal/mergeconflictsignal.go
```diff
@@ -0,0 +1,219 @@
...
+
+// failRequest drives the request to terminal RequestStateError and records the
+// conflict reason on the request log. A not-mergeable verdict is an expected
+// terminal outcome of the check, so the request is concluded here directly.
+//
+// Idempotent under at-least-once delivery: a redelivery whose request is already
+// in Error skips the state CAS but still publishes the log (so a prior attempt
+// that flipped the state but failed before logging is repaired); a request that
+// reached a different terminal state (e.g. a racing cancel) is left untouched.
+func (c *Controller) failRequest(ctx context.Context, request entity.Request, reason string) error {
```
sbalabanov, line 154: likely a copy from another controller - may be consolidate them into a helper at some point, may not be this diff

sbalabanov (changes requested):

submitqueue/orchestrator/controller/dlq/mergeconflictsignal.go
behinddwalls: likely not, we can get rid of it.

submitqueue/orchestrator/controller/mergeconflict/mergeconflict.go
behinddwalls: my thought was these would be contract entity like APIs, i could move them (in a separate change to the top)... or unless you we should have a mirrored SQ entity for queue payload (which would be same as this entity)

submitqueue/orchestrator/controller/mergeconflict/mergeconflict.go
behinddwalls: i thought about a bit..we could do this in validate controller and get rid of it, but thought it would be cleaner in relaying the message say Validating -> Validated, mergeconflictchecking -> mergeconflictchecked

We are not doing PublishLog yet in all controllers but ideally each controller emit that transition as well to the request logs

submitqueue/orchestrator/controller/mergeconflictsignal/mergeconflictsignal.go
behinddwalls: Yes..this is same as we do in DLQ control flow, ideally needed in multiple places which we don't do yet, so i will make it a utility and move to core later

mnoah1 (approved):

PR #248, opened 2026-06-16
[1/N] doc(runway): add workflow RFC for runway domain

kevinlnew:
## Summary
Runway is a consumer-only landing service that owns VCS operations
(mergeability checking and landing) on behalf of SubmitQueue. This RFC
describes the two independent queue flows (check and land), branch
serialization via partition key, statelessness, and ownership model.

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

PR #244, opened 2026-06-15
feat(runway): add merge wire contract and topic keys

behinddwalls:
### Why?

Both the merge-conflict check and the merge itself are moving out of SubmitQueue into an asynchronous round-trip with a separate service, runway. A merge-conflict check is a dry run of a merge, so **one contract serves both**. Runway owns the cross-service queues, their topic keys, and the wire contract — SubmitQueue cannot read runway's storage and vice versa, so the payloads carry full data, not entity IDs. This change adds only those runway-owned definitions; runway's gateway/orchestrator/controllers are out of scope.

### What?

Adds a new `runway/` domain folder holding contract-only definitions:

`runway/entity` — `MergeRequest` (client-owned `ID`, `QueueName`, ordered `[]MergeStep`), `MergeStep` (`StepID`, `[]change.Change`, `mergestrategy.MergeStrategy`), `StepResult` (`StepID`, VCS-neutral `OutputIDs`, `Reason`), and `MergeResult` (`ID`, `Success`, `Reason`, `[]StepResult`), with JSON `ToBytes`/`FromBytes`. The ordered step list encodes base-layering so one request expresses both "candidate vs target branch" (one step) and "candidate + in-flight vs target" (N steps). It imports only the shared `entity/change` and `entity/mergestrategy` — never `submitqueue/entity`.

`runway/core/topickey` — two queue pairs over the one contract: the dry-run check (`merge-conflict-checker` / `merge-conflict-checker-signal`) and the committing merge (`merger` / `merger-signal`). The queue a request arrives on encodes whether runway commits the result and reports the revisions it produced (`StepResult.OutputIDs` — a git commit SHA, hg changeset, svn revision, …; empty for a dry-run check).

## Test Plan
✅ `bazel test //runway/...` (round-trip serialization tests pass)
✅ `bazel build //...`

## Stack
1. @ #244
1. #245
1. #247

PR #249, opened 2026-06-16
[2/N] feat(runway): add domain entities and topic keys 

kevinlnew:
## Summary
Scaffold the runway domain with entity types for the check and land
queues, plus pipeline topic key constants. Runway reuses the shared
entity/change and entity/mergestrategy packages rather than defining
its own strategy type.

## Stack
1. #248
1. @ #249
1. #250
1. #251
1. #252
1. #254


kevinlnew:
Superseded by #244 which landed the runway entities and topic keys on main.

PR #251, opened 2026-06-16
[3/N] feat(runway): add merge-conflict-check and merge controllers

kevinlnew:
## Summary
Implement the two queue message controllers for the runway orchestrator.
The check controller consumes Check messages, calls VCS.CheckMergeability,
and publishes CheckResult. The land controller consumes Job messages,
calls VCS.Land, and publishes Result — conflicts are acked, infra errors
are nacked. Both controllers are stateless (no storage dependency).

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

## Stack
1. #248
1. #250
1. @ #251
1. #252
1. #254

PR #250, opened 2026-06-16
[2/N] feat(runway): add Merger extension interface, mock, and noop impl

kevinlnew:
## Summary
Define the pluggable Merger interface with CheckMergeability (read-only,
request-level) and Merge (mutating, batch-level) methods. Land returns
[]Outcome so the controller owns Result assembly. Includes generated
mock and a noop implementation for wiring and tests.

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

PR #252, opened 2026-06-16
[4/N] feat(runway): add example orchestrator wiring and Makefile targets

kevinlnew:
## Summary
Wire up the runway orchestrator as a consumer-only binary with noop VCS
factory. Unlike submitqueue, runway has no app database — only the queue
infrastructure. Adds Docker Compose (mysql-queue + orchestrator), build
target, and local-start/stop targets to the Makefile.

## Stack
1. #248
1. #250
1. #251
1. @ #252
1. #254


kevinlnew:
Superseded by #265 which landed the runway orchestrator wiring on main.

PR #254, opened 2026-06-16
[5/N] feat(runway): add proto, ping controller, gRPC server, and client

kevinlnew:
## Summary
Add a RunwayOrchestrator gRPC service with a Ping health-check endpoint,
mirroring the stovepipe/submitqueue pattern. Wire the gRPC server into
the example orchestrator alongside the consumer. This is an optional
accessory — the core service works without it.

Also fixes the workflow.md diagram box-drawing characters that rendered
as strikethroughs in rich diffs.

## Stack
1. #248
1. #250
1. #251
1. #252
1. @ #254


kevinlnew:
Superseded by #265 and #260 which landed the proto, ping, client, and Makefile targets on main.

PR #259, opened 2026-06-17
docs(rfc): add message queue contract RFC

behinddwalls:
### Why?

Queue payloads are Go structs serialized with `encoding/json`, so the wire shape is defined only by Go source. There is no language-neutral contract a non-Go client can compile against, no explicit topic-to-payload binding, and no distinction between a domain's private wiring and a published cross-domain contract.

### What?

Doc-only — the design of record for message queue contracts (`doc/rfc/messagequeue-contract.md`):

- **Contract language: Protobuf**, serialized as **protobuf JSON** (`protojson`) so payloads stay self-describing JSON on the wire. The `.proto` is the authority and the Go binding is generated from it, so it cannot drift (no hand-authored struct, no drift test to keep them in sync).
- **Topic binding:** a custom `topics` proto option (defined in `api/base/messagequeue`) carries the wire topic names on the message itself, read back by reflection — not on the publish/consume hot path, which still resolves topics from a `consumer.TopicKey` via the registry.
- **Location by audience:** external contracts (something outside the owning domain depends on them) live in `api/{domain}/messagequeue/`; internal ones in `{domain}/core/messagequeue/`, with the split enforced by Bazel `visibility`.
- Documents the accepted protojson conventions (snake_case field names, UPPER_SNAKE enums, int64-as-string) and why JSON Schema, binary proto, and Avro were rejected.

## Test Plan
- ✅ `make lint` (doc-only)

## Stack
1. #264
1. @ #259
1. #260
1. #245
1. #247

### DIFF
--- a/doc/rfc/messagequeue-contract.md
+++ b/doc/rfc/messagequeue-contract.md
@@ -1,0 +1,104 @@
+# Message Queue Contract
+
+How queue payloads are defined, located, and bound to topics across domains.
+
+## Problem
+
+Queue payloads are Go structs serialized with `encoding/json` (`submitqueue/entity`, `runway/entity`), so the wire shape is defined only by Go source. Three gaps:
+
+- **No language-neutral contract.** Some payloads cross a domain boundary — a client written in another language has nothing to compile or validate against.
+- **No topic-to-payload binding.** `consumer.TopicRegistry` maps a `TopicKey` to a backend, topic name, and subscription — but not to the payload schema. That knowledge lives implicitly in whichever controller (de)serializes.
+- **No audience distinction.** Nothing separates private wiring between our own services from a published cross-domain contract.
+
+## Decisions
+
+### Contract language: Protobuf
+
+Payloads are defined as **proto3 messages**. The `.proto` is the language-neutral authority, and the Go binding is generated from it. This is the same mechanism the RPC contracts in `api/` already use, so queue payloads and RPC payloads share one toolchain, one set of shared field types, and one mental model. Generation runs through the repo's existing hermetic `protoc` Bazel rule (`tool/proto`); message-only contracts (no service) skip the gRPC/YARPC plugins.
+
+The decisive property is that the binding is *generated*, not hand-authored. There is no separate Go struct that can drift from the contract, and therefore no drift test to keep them in sync — the only binding is the generated one.
+
+### Wire format: protobuf JSON
+
+Payloads stay JSON on the wire; messages are serialized with **protobuf JSON (`protojson`)**, not binary proto. The MySQL-backed queue keeps storing self-describing JSON, exactly as before — only the source of the (de)serialization changes from a hand-written `encoding/json` struct to a generated message.
+
+protojson has its own conventions, which the contract adopts deliberately:
+
+- **Field names are the proto names (snake_case).** Serialized with `UseProtoNames`, so `queue_name` stays `queue_name` rather than protojson's default `queueName`. The wire matches the declared field names.
+- **Enums serialize as their value name in UPPER_SNAKE** (`REBASE`, `SQUASH_REBASE`, `MERGE`) — the proto-conventional spelling.
+- **64-bit integers serialize as strings.** A protojson rule for cross-language safety; relevant for any millisecond timestamp or count a future payload carries.
+- **Unknown fields are ignored on read** and zero-valued fields are omitted on write, which gives additive evolution for free: a consumer skips fields it does not yet know.
+
+### Location: audience decides
+
+A contract is **external** when something outside its owning domain depends on it — another domain's service, or a client written in another language. It is **internal** when only the owning domain's own services use it. The test is concrete: *does anything outside this domain compile or deserialize against it?*
+
+- **External** → `api/{domain}/messagequeue/`. The `api/` prefix is the published surface; outside code is expected to depend on it.
+- **Internal** → `{domain}/core/messagequeue/`.
+
+The `.proto`, its generated `protopb`, and any Go helpers co-locate in each home; only the home differs.
+
+### Visibility
+
+Bazel [`visibility`](https://bazel.build/concepts/visibility) enforces the split: `{domain}/core/messagequeue/` targets are scoped to the owning domain, so depending on one from outside is a build error; `api/` targets are public. No metadata keyword — the directory carries the distinction.
+
+### Topic-key binding: the `topic_keys` option
+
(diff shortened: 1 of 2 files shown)


doc/rfc/messagequeue-contract.md
```diff
@@ -0,0 +1,74 @@
+# Message Queue Contract
+
+How queue payloads are defined, located, and bound to topics across domains. Decisions and rationale; code lands in follow-up PRs.
```
sbalabanov: "code lands in follow-up PRs"
I do not think we want to commit this

doc/rfc/messagequeue-contract.md
```diff
@@ -0,0 +1,74 @@
...
+- **No topic-to-payload binding.** `consumer.TopicRegistry` maps a `TopicKey` to a backend, topic name, and subscription — but not to the payload schema. That knowledge lives implicitly in whichever controller (de)serializes.
+- **No audience distinction.** Nothing separates private wiring between our own services from a published cross-domain contract.
+
+## Decisions
+
+### Contract language: JSON Schema
+
+Payloads stay JSON on the wire; the contract is **JSON Schema (draft 2020-12)**. It describes the JSON we already send, has validators and codegen in every language, and changes nothing about the wire.
+
+- **Not Protobuf** — its payoff assumes binary on the wire; describing JSON with it only buys the proto-JSON mapping's quirks.
```
sbalabanov: protobuf can be easily serialized as json
and data validation and generation tooling is arguably way better than JSON's

doc/rfc/messagequeue-contract.md
```diff
@@ -0,0 +1,74 @@
...
+- **Not Avro** — its value is a schema registry for decoding binary; we have none (MySQL-backed queue) and JSON is self-describing, so none is needed.
+
+### Location: audience decides
+
+`api/` means "depended on from outside this domain" — the test is *does something outside the domain compile against it?*
+
+- **External / cross-domain** → `api/{domain}/messagequeue/` (e.g. runway's merge contract, used by non-Go clients).
```
sbalabanov: mentioning "runway" is confusing here

doc/rfc/messagequeue-contract.md
```diff
@@ -0,0 +1,74 @@
...
+- **Internal** → `{domain}/core/messagequeue/`, beside `core/topickey` and the controllers that use it (e.g. the SubmitQueue pipeline).
```
sbalabanov: both of them have "domain", the difference is really not clear

doc/rfc/messagequeue-contract.md
```diff
@@ -0,0 +1,74 @@
...
+
+Schema and Go binding co-locate in each home; only the home differs.
+
+### Visibility
+
+Bazel `visibility` enforces the split: `{domain}/core/messagequeue/` targets are domain-scoped, so cross-domain use is a build error; `api/` targets are public. No metadata keyword — the directory carries the distinction.
```
sbalabanov: link to bazel visibility attribute doc

doc/rfc/messagequeue-contract.md
```diff
@@ -0,0 +1,74 @@
...
+
+- **No language-neutral contract.** Some payloads cross a domain boundary — a non-Go client (e.g. a service asking runway to merge) has nothing to compile or validate against.
+- **No topic-to-payload binding.** `consumer.TopicRegistry` maps a `TopicKey` to a backend, topic name, and subscription — but not to the payload schema. That knowledge lives implicitly in whichever controller (de)serializes.
+- **No audience distinction.** Nothing separates private wiring between our own services from a published cross-domain contract.
+
+## Decisions
+
+### Contract language: JSON Schema
+
+Payloads stay JSON on the wire; the contract is **JSON Schema (draft 2020-12)**. It describes the JSON we already send, has validators and codegen in every language, and changes nothing about the wire.
```
sbalabanov: what exact codegen to use?
there are annoyingly difficult to implement correctly with Bazel

sbalabanov (changes requested):

doc/rfc/messagequeue-contract.md
behinddwalls: Removed — dropped the "code lands in follow-up PRs" sentence from the intro.

doc/rfc/messagequeue-contract.md
behinddwalls: Agreed, and acted on it: the RFC now adopts **Protobuf serialized as protobuf JSON** as the contract language (the binding is generated, so it can't drift), and JSON Schema is moved to Rejected. This comment drove the flip.

doc/rfc/messagequeue-contract.md
behinddwalls: Removed the runway example here; the Location section no longer names specific domains.

doc/rfc/messagequeue-contract.md
behinddwalls: Rewrote the Location section to make the distinction concrete: a contract is *external* when something **outside its owning domain** depends on it (another domain's service, or a client in another language), *internal* when only the owning domain's services use it.

doc/rfc/messagequeue-contract.md
behinddwalls: Added a link to https://bazel.build/concepts/visibility on the `visibility` mention.

doc/rfc/messagequeue-contract.md
behinddwalls: Clarified: generation runs through the repo's existing hermetic `protoc` Bazel rule (`tool/proto`); message-only contracts skip the gRPC/YARPC plugins. No new codegen toolchain — same one the RPC protos already use.

mnoah1 (approved):

doc/rfc/messagequeue-contract.md
```diff
@@ -0,0 +1,104 @@
...
+protojson has its own conventions, which the contract adopts deliberately:
+
+- **Field names are the proto names (snake_case).** Serialized with `UseProtoNames`, so `queue_name` stays `queue_name` rather than protojson's default `queueName`. The wire matches the declared field names.
+- **Enums serialize as their value name in UPPER_SNAKE** (`REBASE`, `SQUASH_REBASE`, `MERGE`) — the proto-conventional spelling.
+- **64-bit integers serialize as strings.** A protojson rule for cross-language safety; relevant for any millisecond timestamp or count a future payload carries.
+- **Unknown fields are ignored on read** and zero-valued fields are omitted on write, which gives additive evolution for free: a consumer skips fields it does not yet know.
+
+### Location: audience decides
+
+A contract is **external** when something outside its owning domain depends on it — another domain's service, or a client written in another language. It is **internal** when only the owning domain's own services use it. The test is concrete: *does anything outside this domain compile or deserialize against it?*
```
sbalabanov, line 34: might be better give a definition of a "domain"

PR #260, opened 2026-06-17
feat(api/runway): add external merge queue contract

behinddwalls:
### Why?

Runway's merge queues are the first cross-domain message queue contract: a client (potentially non-Go) publishes a merge request and consumes the result without access to Runway's Go types or storage. Per the RFC (#259) this needs a language-neutral, proto-defined contract.

### What?

Establishes the message queue contract pattern using Runway's merge queues as the reference:

- Adds `api/runway/messagequeue/proto/merge.proto` defining `MergeRequest`/`MergeResult`, reusing the shared `Change` and `Strategy` types and the `topics` option from `api/base`; generated into `protopb`.
- Payloads are serialized as **protobuf JSON** via thin `protojson` helpers (`MergeRequestToBytes`/`MergeRequestFromBytes` and the `MergeResult` counterparts); a `Topics()` reflection helper exposes the topic binding. Topic keys are co-located with the contract.
- Moves the merge bindings out of `runway/entity` into `api/runway/messagequeue`.
- A drift test round-trips the payloads and asserts every Runway topic key is bound to exactly one message via the `topics` option.

## Test Plan
- ✅ `./tool/bazel test //api/runway/messagequeue:messagequeue_test`
- ✅ `./tool/bazel build //...`

## Stack
1. #264
1. #259
1. @ #260
1. #245
1. #247


mnoah1 (approved):

api/runway/messagequeue/proto/merge.proto
```diff
@@ -0,0 +1,101 @@
...
+// is attributable to the step that produced (or failed to produce) it.
+message StepResult {
+    // step_id echoes the step_id of the step this result is for.
+    string step_id = 1;
+    // output_ids are the VCS-neutral identifiers of the revisions this step
+    // produced on the target branch -- a git commit SHA, a Mercurial changeset
+    // hash, a Subversion revision number, a Perforce changelist, and so on.
+    // Empty for a dry-run check, for a change already present on the target, or
+    // for a step that failed to apply.
+    repeated string output_ids = 2;
```
sbalabanov: may be it should be an object (message) so it is easier to extend in the future with additional metadata.

sbalabanov: also probably important to contract if it is ordered or not.

api/runway/messagequeue/proto/merge.proto
```diff
@@ -0,0 +1,101 @@
...
+// request completes.
+message MergeResult {
+    option (uber.base.messagequeue.topic_keys) = "merge-conflict-check-signal";
+    option (uber.base.messagequeue.topic_keys) = "merge-signal";
+
+    // id echoes the client-owned correlation id from the request.
+    string id = 1;
+    // success is true if the whole ordered step sequence applied cleanly:
+    // mergeable for a dry-run check, merged for a committing merge.
+    bool success = 2;
```
sbalabanov: better use enum for future extensibility, even if it is binary now

api/runway/messagequeue/merge.go
```diff
@@ -0,0 +1,107 @@
+// Copyright (c) 2025 Uber Technologies, Inc.
```
sbalabanov, line 1: can this be autogenerated?

sbalabanov (approved):

api/runway/messagequeue/merge.go
behinddwalls: porbably but it's marshaling using protojson right now

PR #262, opened 2026-06-17
feat(stovepipe): add Ingest RPC to gateway for trunk commit verification

behinddwalls:
### Why?

Stovepipe verifies commits that are already on trunk (post-merge/post-submit build and test), but its gateway only exposed Ping. It needs an intake RPC, analogous to SubmitQueue's Land, where a client hands in a trunk commit and gets back a tracking id while verification runs asynchronously.

### What?

Adds the `Ingest` RPC to the Stovepipe gateway proto contract (proto-only; controller and server wiring follow separately):

- `IngestRequest{ queue, change }` and `IngestResponse{ spid }`, where `spid` is the "stovepipe id" handle for tracking the request lifecycle (mirroring Land's `sqid`).
- `change` reuses the shared `uber.base.change.Change` wire type from the parent PR; for git, callers pass a `git://<remote>/<repo>/<ref>/<commit_sha>` commit URI. Unlike Land there is no merge strategy, since the commit is already on trunk.

## Test Plan

- ✅ `make proto`
- ✅ `./tool/bazel build //...`
- ✅ `make test`

## Stack
1. #258
1. #261
1. @ #262


api/stovepipe/gateway/proto/gateway.proto
```diff
@@ -39,8 +41,27 @@ message PingResponse {
...
+    string queue = 1;
+    // Change (trunk commit) to ingest, identified by URI. The target trunk and build/test configuration are defined by the queue configuration.
+    // Stovepipe ingests commits that are already on trunk; for git, supply a "git://<remote>/<repo>/<ref>/<commit_sha>" URI.
+    uber.base.change.Change change = 2;
+}
+
+// IngestResponse defines the response to an ingest request.
+message IngestResponse {
+    // Globally unique identifier for the ingested commit's verification request (the "stovepipe id", spid). Used to track the request lifecycle.
+    string spid = 1;
```
gsturges, line 56: Can you give more clarification on what this ID is? Is it for identifying each request? If so should we store each request? Or is the id of the change(s) ingested into the queue. If so, shouldn't we give the id of each uri ingested into the queue?

mnoah1 (approved):

api/stovepipe/gateway/proto/gateway.proto
behinddwalls: yes, so this should be generated by gateway against a queue and the counter, very similar to what is done in SQ

behinddwalls:

**To fix manually:**
```bash
git fetch origin
git checkout preetam/stovepipe-gw-ingest-proto
git rebase --onto origin/main d3e0a48a56f44d928321563515799cde786ca4b5 preetam/stovepipe-gw-ingest-proto
# resolve conflicts, then:
git push --force-with-lease
```

```bash
gh pr edit 262 --base main
```

PR #266, opened 2026-06-22
feat(api/base): add PROMOTE merge strategy

behinddwalls:
### Why?

Stovepipe promotes an already-verified commit onto a verified branch (e.g. `verified/main`) as-is. The existing `REBASE`/`SQUASH_REBASE`/`MERGE` strategies cannot express this — they all transform the change and produce new revisions. We need a shared, VCS-neutral way for a client to ask Runway to advance a branch to an already-existing commit.

### What?

Adds `PROMOTE` (= 4) to the shared `Strategy` enum in `api/base/mergestrategy` and the matching `MergeStrategyPromote` Go entity constant in `platform/base/mergestrategy`. `PROMOTE` means "integrate the exact revision as-is, with no content transform": each backend realizes it natively (git fast-forward, Mercurial bookmark advance, Subversion/Perforce copy), so a step's output revision is the same commit the request named rather than a freshly created one. The Runway merge-queue contract README gains a "Merge strategy" section documenting the semantics.

This is contract + docs only — no producer or consumer wiring is added yet.

## Test Plan

✅ `make proto` (regenerated `api/base/mergestrategy/protopb`)
✅ `bazel build //api/base/mergestrategy/protopb //platform/base/mergestrategy`

Co-authored-by: Cursor <cursoragent@cursor.com>

### DIFF
--- a/api/base/mergestrategy/proto/mergestrategy.proto
+++ b/api/base/mergestrategy/proto/mergestrategy.proto
@@ -36,4 +36,8 @@ enum Strategy {
     SQUASH_REBASE = 2;
     // Merge commits into the target branch by creating a separate merge commit, preserving commit history along with hashes.
     MERGE = 3;
+    // Integrate the exact revision as-is, with no content transform — advance the target branch to an already-existing
+    // commit rather than producing new revisions. The implementer maps it to its backend: git fast-forward, Mercurial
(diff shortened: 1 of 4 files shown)


kevinlnew (approved):
LGTM

PR #267, opened 2026-06-23
feat(stovepipe): add request and request-log entity stores

behinddwalls:
### Why?

Stovepipe needs a persistence layer to track trunk-validation requests, mirroring SubmitQueue's two-store split by service ownership. Until now Stovepipe only had the `IngestRequest` queue payload and no entity stores, so the orchestrator forwarded full payloads and nothing recorded request state or customer-facing status.

### What?

Adds the foundational entities and storage interfaces (plus a MySQL implementation and mocks), mirroring the SubmitQueue pattern. Identity stays the synthetic SPID (`{queue}/{counter}`) and everything is modeled as request + change + URIs — there is no SHA/commit entity.

- `stovepipe/entity`: `Request` (orchestrator-owned state machine with `State`/`Version`), the `RequestState` enum, a thin `RequestID` queue payload, and an append-only `RequestLog` with a customer-facing `RequestStatus`. The existing `IngestRequest` stays as the gateway-to-orchestrator payload.
- `stovepipe/extension/storage`: a `Storage` factory plus `RequestStore` (orchestrator-owned) and `RequestLogStore` (gateway-owned) interfaces, shared error sentinels, and `//go:generate` mockgen directives.
- `stovepipe/extension/storage/mysql`: a MySQL-backed implementation with optimistic-locking conditional writes and `request` / `request_log` schemas.
- `stovepipe/extension/storage/mock`: generated gomock mocks; the Makefile `mocks` target now includes the Stovepipe storage path.

Wiring the stores into controllers and example servers is a deliberate follow-up.

This also includes the regenerated `api/submitqueue/gateway/protopb/gateway.pb.yarpc.go` descriptor closure — a leftover regeneration from the already-merged PROMOTE merge-strategy change, since the gateway yarpc file embeds the transitive mergestrategy descriptor.

## Test Plan

- `bazel test //stovepipe/...` — all targets pass
- `make fmt`, `make lint-license`, `make gazelle` — clean / idempotent
- `make proto` reproduces the committed descriptors


kevinlnew (approved):
LGTM

PR #287, opened 2026-06-30
feat(orchestrator): emit batched request log from batch controller

albertywu:
## Summary
Previously the batch controller was not publishing a request log on completion. This is inconsistent with the other controllers, which publish a terminal request log when its action is complete.

This PR emits a request log after the batch is persisted and before publishing to score.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

## Test Plan
✅ make fmt && make build && make test && make check-mocks && make e2e-test

## Stack
1. @ #287
1. #288
1. #289

### DIFF
--- a/submitqueue/orchestrator/controller/batch/batch.go
+++ b/submitqueue/orchestrator/controller/batch/batch.go
@@ -24,6 +24,7 @@ import (
 	"github.com/uber/submitqueue/platform/consumer"
 	"github.com/uber/submitqueue/platform/extension/counter"
 	"github.com/uber/submitqueue/platform/metrics"
+	corerequest "github.com/uber/submitqueue/submitqueue/core/request"
 	"github.com/uber/submitqueue/submitqueue/core/topickey"
 	"github.com/uber/submitqueue/submitqueue/entity"
 	"github.com/uber/submitqueue/submitqueue/extension/conflict"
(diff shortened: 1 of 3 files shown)


submitqueue/orchestrator/controller/batch/batch.go
```diff
@@ -287,6 +288,20 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) (r
...
 	)
 
+	// Record the "batched" status in the request log. This status corresponds to
+	// the RequestStateBatched transition CAS'd above, so it carries the request
+	// version for reconciliation (unlike the batch-level "scored" status). The
+	// message ID is scoped to (requestID, status), so a redelivery that creates a
+	// fresh batch re-emits "batched" with a different batch_id but is deduped to
+	// the first entry — acceptable, the request is batched either way.
+	logEntry := entity.NewRequestLog(request.ID, entity.RequestStatusBatched, request.Version, "", map[string]string{
+		"batch_id": batch.ID,
```
sbalabanov, line 298: may be also record other request ids in the same batch to make debugging easier

sbalabanov (approved):

submitqueue/orchestrator/controller/batch/batch.go
albertywu: sg

PR #295, opened 2026-07-02
(feat)build: Migrate Bazel target naming to go_default_library

roychying:
Adopt go_default_library naming convention for Bazel targets 

## Why?
The internal go-code monorepo consumes this repo via Bazel with `build_file_generation = "off"` -  it uses the checked-in `BUILD.bazel` files **directly** instead of regenerating them. go-code's gazelle enforces the`go_default_library` naming convention for all external repos globally (`# gazelle:go_naming_convention_external go_default_library`), so every dependency label it writes points at `:go_default_library`. For go-code to reference this repo's targets without per-import fixups, our targets must be reachable as `:go_default_library`.

This complements the proto-disable change (https://github.com/uber/submitqueue/pull/293): together they let go-code build this repo out of the box on Bazel 8.5.1 (which removed `native.java_proto_library`, breaking `proto_library`).

## What?
- Add `# gazelle:go_naming_convention go_default_library` to the root `BUILD.bazel` and regenerate target names via gazelle: libraries/binaries ->`go_default_library`, tests -> `go_default_test`; external refs normalize to `:go_default_library`.
- Manually migrate hand-authored, custom-named test targets that gazelle cannot rename across a convention change:
  - 9 integration/e2e `go_test` targets - `go_default_test` (`data`/`tags` preserved once the name matches the convention),
  - remove the redundant `platform/consumer:consumer_test` (gazelle now emits `go_default_test` for that package),
  - point the `Makefile` integration-test targets at `:go_default_test`.
- Drop redundant `# keep`-pinned short-name `protopb` deps - under the new convention the `:go_default_library` equivalents are emitted beside them, so the short-name `:protopb` labels are dangling.

## Test Plan
- `make gazelle` - clean and idempotent (no duplicate-rule errors).
- `make build` -  223 targets built.
- `make test` -  65/65 unit tests pass.
- Verified downstream consumption from go-code with `build_file_generation="off"`
  by overriding the module at this branch:
  bazel test //src/.../submitqueue/{gateway,orchestrator,stovepipe}/...
    --override_repository=com_github_uber_submitqueue=/path/to/submitqueue
  -  71 targets build, 20/20 consumer tests pass.


gsturges (approved):

api/base/change/protopb/BUILD.bazel
```diff
@@ -1,12 +1,12 @@
...
 go_library(
-    name = "protopb",
+    name = "go_default_library",
     srcs = ["change.pb.go"],
     importpath = "github.com/uber/submitqueue/api/base/change/protopb",
     visibility = ["//visibility:public"],
     deps = [
-        "@org_golang_google_protobuf//reflect/protoreflect",
-        "@org_golang_google_protobuf//runtime/protoimpl",
+        "@org_golang_google_protobuf//reflect/protoreflect:go_default_library",
```
sbalabanov, line 9: how this is supposed to work? it is an external dependency.

sbalabanov (changes requested):
isn't proto lib removal a separate diff?

api/base/change/protopb/BUILD.bazel
roychying: these external dependences have `:go_default_library` alias. Not strictly required, just want to keep consistency across all imported packages
```
bazel query --output=build "@org_golang_google_protobuf//reflect/protoreflect:go_default_library"
INFO: Invocation ID: 6887c543-e59d-40a1-9a29-57ccc312b3aa
# /home/user/.cache/bazel/_bazel_chenghan.ying/9fe95cfc2da7ac7338edf63ca6b444da/external/gazelle++go_deps+org_golang_google_protobuf/reflect/protoreflect/BUILD.bazel:24:6
alias(
  name = "go_default_library",
  visibility = ["//visibility:public"],
  actual = "@org_golang_google_protobuf//reflect/protoreflect:protoreflect",
)
```

roychying: actually changes in external packages are guaranteed. Setting `# gazelle:go_naming_convention go_default_library` for our own local targets unintentionally also renames every external dependency reference to `:go_default_library`, because Gazelle's default `import_alias` convention for external repos falls back to whatever the local naming convention is when no `go_naming_convention_external` directive is explicitly set.
If we manually revert back to the short name, `make gazelle` will just update it again to `go_default_library`

behinddwalls (approved):

PR #299, opened 2026-07-06
chore(bazel): add go_default_library alias for downstream naming compatibility

roychying:
## Why?
Some Bazel-based downstream consumers of this repo enforce their own naming convention where Go targets must be named `go_default_library`/`go_default_test`. When such a consumer uses our checked-in BUILD files directly (rather than regenerating them), our native target names don't satisfy that convention, so their consumer labels fail to resolve.

We need a fix that satisfies that requirement without renaming our own targets or breaking any other Bazel consumer relying on the native names.

This PR is addressing the same issue as #295 but use alias approach which is cheaper and less invasive.

## What?
Adds `# gazelle:go_naming_convention import_alias` to the root `BUILD.bazel` and regenerates all BUILD files via `make gazelle`.

## Test Plan

- make gazelle - clean; re-running make gazelle produces no further diff (idempotent).
- make build - all 337 targets build.
- make test - 65/65 tests pass, with hand-authored test targets keeping their original native names (e.g. buildsignal_test, not renamed to go_default_test).
- Verified consumption from a downstream Bazel workspace using this repo directly (checked-in BUILD files, no regeneration): with a local override pointed at this checkout, successfully built and tested downstream targets against the new go_default_library alias â confirming consumer labels resolve correctly end-to-end.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -2,6 +2,13 @@ load("@gazelle//:def.bzl", "gazelle")
 
 # gazelle:prefix github.com/uber/submitqueue
 
+# Add a go_default_library alias next to every native-named Go target so
+# consumers pinning go-code's naming convention (e.g. go-code's own
+# go_repository consumption with build_file_generation="off") can depend on
+# ":go_default_library" without renaming our own targets or breaking existing
+# consumers that use the native short names.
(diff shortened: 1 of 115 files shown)


roychying:
Close this one. Adopt #295

PR #301, opened 2026-07-06
test(e2e): add stovepipe end-to-end suite driving the ingest→process pipeline

behinddwalls:
## Summary
Intent:
- Stovepipe had a full-stack integration suite but no e2e suite, and nothing proved the asynchronous process stage actually drained a message — the integration suite only asserts that Ingest publishes one.
- Add an e2e suite whose distinguishing charter is proving the ingest→process pipeline runs end-to-end, mirroring the submitqueue test/e2e pattern.

Changes:
- Add test/e2e/stovepipe with a testify suite (suite_test.go) that boots service/stovepipe/docker-compose.yml under the e2e-stovepipe context and a harness (harness_test.go) of intent-level helpers.
- Tests cover Ping, Ingest happy-path (synchronous request/request_uri/queue_messages side effects plus asynchronous consumption), and Ingest idempotency/dedup.
- awaitProcessed observes the durable acked-offset watermark in queue_offsets (which only advances on ack) rather than the message's delivery-state row, because the queue GCs acked messages from queue_messages once the watermark passes them.
- Left the existing integration suite untouched (added alongside, per the submitqueue precedent of keeping both levels).

## Test Plan
- bazel test //test/e2e/stovepipe:stovepipe_test passes (Ping, Ingest happy-path, Ingest idempotency) against the Docker Compose stack. Non-vacuity confirmed: an offset-based signal that is absent makes awaitProcessed time out (observed while iterating), so the async assertion is real.

---

### DIFF
--- a/test/e2e/stovepipe/BUILD.bazel
+++ b/test/e2e/stovepipe/BUILD.bazel
@@ -1,0 +1,30 @@
+load("@rules_go//go:def.bzl", "go_test")
+
+go_test(
+    name = "go_default_test",
+    srcs = [
+        "harness_test.go",
+        "suite_test.go",
+    ],
+    data = [
+        "//:MODULE.bazel",
+        "//:go.mod",
+        "//platform/extension/messagequeue/mysql/schema",
+        "//service/stovepipe:docker-compose.yml",
+        "//stovepipe/extension/storage/mysql/schema",
+    ],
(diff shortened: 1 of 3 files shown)


mnoah1 (approved):
Approved pending the fix to the build failure

PR #302, opened 2026-07-07
[2/N] [List API] feat(context): add immutable request admission store

albertywu:
## Summary

Adds immutable request context when request is admitted.

## Test Plan
✅ `make fmt && make build && make test && make check-mocks && make e2e-test`

## Stack
1. #310
1. @ #302
1. #303
1. #304
1. #305
1. #306

PR #303, opened 2026-07-07
[3/N] [List API] feat(summary): project request logs from admission context

albertywu:
## Summary
Adds the RequestSummary entity, storage contract, generated mocks, and MySQL implementation for the gateway-owned list read model. The store merges immutable request-log events into a queue-scoped summary table with optimistic concurrency.

## Test Plan

## Stack
1. #310
1. #302
1. @ #303
1. #304
1. #305
1. #306


submitqueue/extension/storage/mysql/request_summary_store.go
```diff
@@ -0,0 +1,423 @@
...
+// UpsertFromLog incrementally merges one request-log event into the summary read model.
+//
+// This repo forbids database transactions, so the merge uses optimistic concurrency instead of a
+// SELECT ... FOR UPDATE: read the current summary without a lock, merge the incoming log in memory,
+// then write back with a conditional update guarded by the row version (the same pattern as
+// requestStore.UpdateState). A concurrent writer — another gateway write path or a redelivered log
+// event — invalidates the read, in which case we re-read and re-merge. Merges are monotonic (the
+// winner only advances by request version or timestamp), so the loop converges; re-applying a log
+// that has already been merged is a no-op.
+func (s *requestSummaryStore) UpsertFromLog(ctx context.Context, log entity.RequestLog) (retErr error) {
```
behinddwalls: ideally this is purely a write to DB, like any other store with CAS, any logic to summarize should live in the request logs consumer controller, we can event thing of adding more queues if needed post persisting requestLog to summarize and do any other things as needed. But we can start with requestlog controller itself

submitqueue/extension/storage/request_summary_store.go
```diff
@@ -0,0 +1,65 @@
...
+	Statuses    []entity.RequestStatus
+	Sort        RequestSummarySort
+	Cursor      *RequestSummaryCursor
+	Limit       int
+}
+
+// RequestSummaryCursor is the stable cursor position under the selected admission-time ordering.
+type RequestSummaryCursor struct {
+	StartedAtMs int64
+	RequestID   string
```
behinddwalls, line 56: do we need both or just the last RequestID we returned as cursor?

submitqueue/entity/request_summary.go
```diff
@@ -0,0 +1,29 @@
...
+	RequestID     string
+	Queue         string
+	ChangeURIs    []string
+	Status        RequestStatus
+	LastError     string
+	Metadata      map[string]string
+	StartedAtMs   int64
+	UpdatedAtMs   int64
+	CompletedAtMs int64
+	Terminal      bool
```
behinddwalls: do we need this? should the status be enough to reflect whether it's terminal or not

submitqueue/core/request/store.go
```diff
@@ -0,0 +1,35 @@
...
+
+// PersistLog inserts the immutable request log entry and best-effort updates
+// the request summary read model. The request log is the source of truth; a
+// summary failure is recorded by the summary store's own operation metrics but
+// must not fail request ingestion or cause log-topic redelivery.
+func PersistLog(ctx context.Context, store storage.Storage, log entity.RequestLog) error {
+	if err := store.GetRequestLogStore().Insert(ctx, log); err != nil {
+		return fmt.Errorf("failed to insert request log for request_id=%s: %w", log.RequestID, err)
+	}
+	_ = store.GetRequestSummaryStore().UpsertFromLog(ctx, log)
```
sbalabanov: let's not swallow errors

submitqueue/extension/storage/mysql/schema/request_summary.sql
```diff
@@ -0,0 +1,17 @@
...
+    status_timestamp_ms BIGINT NOT NULL,
+    winner_terminal_version BOOLEAN NOT NULL,
+    last_error TEXT NOT NULL,
+    metadata JSON NOT NULL,
+    started_at_ms BIGINT NOT NULL,
+    updated_at_ms BIGINT NOT NULL,
+    completed_at_ms BIGINT NOT NULL,
+    terminal BOOLEAN NOT NULL,
+    version BIGINT NOT NULL,
+    PRIMARY KEY (queue, request_id)
```
sbalabanov: what queries will be executed on this table?
if we filter by queue and status (i.e. executing or completed) then it will do partial table scan which is inefficient

submitqueue/extension/storage/request_summary_store.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+// RequestSummaryListOptions defines a page-in request for RequestSummaryStore.
+type RequestSummaryListOptions struct {
+	Queue       string
+	StartTimeMs int64
+	EndTimeMs   int64
+	Statuses    []entity.RequestStatus
+	Sort        RequestSummarySort
+	Cursor      *RequestSummaryCursor
```
sbalabanov: why pointer?

submitqueue/extension/storage/request_summary_store.go
```diff
@@ -0,0 +1,65 @@
...
+	Limit       int
+}
```
sbalabanov, line 51: ok now I see a list of criteria. If we intend to keep request log records for each queue indefinitely, the index has to be redesigned. Right now it will do full scan for pretty much all of the query parameters.
If we always sort by the same field, it should be included in the index to at least do paginated index scan.

submitqueue/extension/storage/request_summary_store.go
```diff
@@ -0,0 +1,65 @@
...
+	UpsertFromLog(ctx context.Context, log entity.RequestLog) error
+
+	// List returns a page of request summaries matching the queue, time window, and optional statuses.
+	List(ctx context.Context, opts RequestSummaryListOptions) (RequestSummaryListResult, error)
+}
+
+// RequestSummaryListOptions defines a page-in request for RequestSummaryStore.
+type RequestSummaryListOptions struct {
+	Queue       string
+	StartTimeMs int64
```
sbalabanov, line 44: please comment specification, what values to set to what result to expect. How to specify that filtering should not be done by a particular field, or should it always specify arbitrary big/small values?

sbalabanov (changes requested):

submitqueue/extension/storage/mysql/request_summary_store.go
albertywu: Agreed. Summary reconciliation has moved out of the MySQL store. Application code loads context and the current summary, computes the next projection, and performs a conditional write. The store now exposes only create, get, conditional update, and page query operations.

submitqueue/extension/storage/request_summary_store.go
albertywu: We need both because pagination is ordered by `(started_at_ms, request_id)`. `request_id` alone cannot identify the continuation point in admission-time order, and `started_at_ms` alone is not unique. The cursor therefore carries the complete last-seen ordering key.

submitqueue/entity/request_summary.go
albertywu: Agreed for the API-facing value. The redundant `Terminal` field was removed from the summary response. The remaining internal `WinnerTerminalVersion` field is reconciliation state rather than display state: it distinguishes an authoritative versioned terminal winner from an unversioned status so incremental projection matches full-log reconciliation.

submitqueue/core/request/store.go
albertywu: Fixed. `PersistLog` now returns summary projection errors instead of swallowing them, so the delivery is retried. Reprocessing tolerates the immutable log already being present and retries the projection.

submitqueue/extension/storage/mysql/schema/request_summary.sql
albertywu: The RFC and query shape are now explicit. The first API always requires a bounded admission-time window and orders by admission time. The `(queue, started_at_ms, request_id)` index supports the range predicate, ordering, and keyset continuation in one scan. Optional current-status filtering may inspect additional rows inside that requested window, but it no longer causes a queue-wide historical scan. We are intentionally not adding a speculative status index until representative cardinalities justify it.

submitqueue/extension/storage/request_summary_store.go
albertywu: `nil` represents the first page without inventing sentinel cursor values. A non-nil cursor must contain the complete `(started_at_ms, request_id)` continuation key. The option contract now documents those semantics explicitly.

submitqueue/extension/storage/request_summary_store.go
albertywu: Agreed. We changed the first API semantics to a required admission-time window rather than lifecycle overlap. The query is now bounded by `(queue, started_at_ms)` and uses the same fields for deterministic keyset ordering. The table has `(queue, started_at_ms, request_id)`, so the backend does not scan admissions outside the requested window. The RFC also specifies retention rather than keeping summaries indefinitely. We documented that true 'active at any point in the window' semantics would require an additional purpose-built projection and are deferred.

submitqueue/extension/storage/request_summary_store.go
albertywu: Updated the option contract. Queue, a positive half-open admission window, and a positive limit are required. Empty `Statuses` explicitly disables status filtering, so callers do not use arbitrary sentinel values. Empty `Sort` selects admission ascending, and nil `Cursor` selects the first page.

albertywu:
Closing in favor of new design in https://github.com/uber/submitqueue/pull/314

PR #304, opened 2026-07-07
[4/N] [List API] feat(gateway): expose List API from request summaries

albertywu:
## Summary
Adds the public gateway List RPC and controller for queue-scoped request summaries with queue validation, status filtering, deterministic pagination, and admission-time sorting. Gateway write paths now persist log entries through the summary-aware helper so the read model stays updated.

## Test Plan

## Stack
1. #310
1. #302
1. #303
1. @ #304
1. #305
1. #306


behinddwalls:
Few thoughts on this.

### 1. Group filters vs. pagination

`ListRequest` is a bit hard to read at a glance — filters, ordering, and pagination sit flat in one message, so it's unclear what's required, what's optional, and what's meant to combine. Grouping by role would make it more self-documenting.

On whether filters should be a generic `Filter{key, value}` list for extensibility: I'd lean against it. A proto message with named fields already *is* a typed key/value model — the keys are fields the schema knows. A `Filter{key, value}` list gives that up: keys move into runtime data, values collapse to strings, and we'd reinvent operators and validation. So I'd keep filters as named typed fields, grouped into a sub-message. (If we ever need open-ended user-composed predicates, the usual path is a single `string filter` grammar à la AIP-160 — but not yet.)

### 2. A possible restructure

A starting point:

```proto
message ListRequest {
  // Required: the queue to scan (storage partition key).
  string queue = 1;
  // Optional. Absent = whole queue.
  // Present fields AND together; a repeated field ORs within itself.
  ListFilter filter = 2;
  // Optional ordering; unspecified = admission order (FIFO).
  ListSort sort = 3;
  // Optional pagination, independent of filtering.
  Page page = 4;
}

message ListFilter {
  TimeWindow window = 1;          // either bound 0 = unbounded
  repeated string statuses = 2;   // OR-set; empty = all
}

// Half-open [start, end) in Unix epoch ms. 0 = unbounded.
message TimeWindow {
  int64 start_time_ms = 1;
  int64 end_time_ms = 2;
}

message Page {
  int32 size = 1;   // 0 = server default; server may cap
  string token = 2; // opaque continuation (see below)
}

enum ListSort {
  LIST_SORT_UNSPECIFIED  = 0;  // = admission ascending
  LIST_SORT_ADMITTED_ASC = 1;
  LIST_SORT_ADMITTED_DESC = 2;
}
```

Small things folded in:
- proto3 `optional` on the time bounds would distinguish "no lower bound" from `start_time_ms = 0`.
- A `oneof` could encode any mutually-exclusive filters in the schema rather than in prose.
- I'd suggest dropping the `terminal` filter: it overlaps `statuses`, and `terminal=true` + `statuses=[building]` is a contradictory request the server then has to define. Terminality is derivable from the status set. Worth keeping the derived `terminal` bool on the **response** `RequestSummary` though — as output it saves clients from hardcoding which statuses are terminal.

### 3. In-flight requests in the time window

One edge case worth a look: the contract says "lifecycles overlap a time window," but the storage filter (parent PR) is `started_at_ms < end AND completed_at_ms > start`, and in-flight requests have `completed_at_ms = 0`. So `0 > start` is false whenever `start_time_ms > 0` — still-running requests get filtered out, which seems opposite to what a "queue right now" view wants. Treating 0 as open-ended (`completed_at_ms = 0 OR completed_at_ms > ?`) would fix it, and might argue for allowing an unbounded window. Let me know if I've misread the query.

### 4. Pagination token — a note and an open question

Embedding the query in the token looks safe: values go through bound `?` parameters and sort/order comes from a whitelist switch, so no injection path, and the token carries no entitlements — a tampered one only moves the caller's own cursor within an already-authorized query. Might be worth stating as an invariant: *the token can hold position + a query echo, never anything the server treats as permission.*

Optional for later: if we ever want forged tokens to be impossible rather than just harmless, we could HMAC-sign the token with a server-side secret. The signature also guarantees integrity, so the token could shrink to position + a query hash instead of echoing the whole query. Cost is key management, so probably a follow-up unless the token needs to carry something sensitive. A length cap before decoding is a cheap safeguard regardless.

Open one, no strong opinion — **`token` or `cursor`?**
- `token` matches the AIP convention and the internal `listPageToken`, and signals "opaque, tied to this query, don't introspect" — which fits, since it's position + frozen query and we reject on mismatch. Keeps it distinct from the storage `RequestSummaryCursor` (the actual position).
- `cursor` is more familiar to consumers (GraphQL/Relay), but implies a plain reusable pointer, which is at odds with the mismatch rejection. If we preferred it, we'd probably make it a true position-only cursor.

Curious what you think — feels like a judgment call, and it interacts with the HMAC question.

### 5. Tiny things
- `ListSort` may trip buf/AIP lint (zero-value naming + unprefixed values) — the `LIST_SORT_UNSPECIFIED` + prefixed form above should fix it.
- `statuses` as free strings seems reasonable (mirrors `StatusResponse.status`), just worth documenting intent for unknown values: today a hard error, but we could treat them as ignored so an older client filtering on a renamed status doesn't break.

Let me know your thoughts?

api/submitqueue/gateway/proto/gateway.proto
```diff
@@ -155,4 +214,7 @@ service SubmitQueueGateway {
     // Status returns the current status of a previously submitted request, identified by its sqid.
     // The status is eventually consistent with the request store and reconciled from the append-only request log.
     rpc Status(StatusRequest) returns (StatusResponse) {}
+
+    // List returns request summaries for one queue whose lifecycles overlap a time window.
```
sbalabanov: I can't mentally parse the description :)

submitqueue/gateway/controller/list.go
```diff
@@ -0,0 +1,262 @@
...
+		requestSummaryStore: requestSummaryStore,
+		queueConfigs:        queueConfigs,
+	}
+}
+
+// List returns request summaries for one queue whose lifecycles overlap a time window.
+func (c *ListController) List(ctx context.Context, req *pb.ListRequest) (*pb.ListResponse, error) {
+	start := time.Now()
+	defer func() {
+		c.metricsScope.Timer("list_latency").Record(time.Since(start))
```
sbalabanov, line 62: plz no timers.
use platform wrappers i.e. https://github.com/uber/submitqueue/blob/main/submitqueue/gateway/controller/land.go#L95

submitqueue/gateway/controller/list.go
```diff
@@ -0,0 +1,262 @@
...
+		return nil, fmt.Errorf("ListController invalid status filter: %w", err)
+	}
+	sortMode, err := requestSummarySort(req.Sort)
+	if err != nil {
+		return nil, fmt.Errorf("ListController invalid sort: %w", err)
+	}
+
+	if _, err := c.queueConfigs.Get(ctx, req.Queue); err != nil {
+		if errors.Is(err, queueconfig.ErrNotFound) {
+			return nil, errs.NewUserError(&UnrecognizedQueueError{Queue: req.Queue})
```
sbalabanov, line 87: req.Queue == ""
should also be a user error then? But it is not.
we need to design an error classification propagation for gateways/grpc, similar to orchestrator/queue

submitqueue/gateway/controller/list.go
```diff
@@ -0,0 +1,262 @@
...
+		}
+		return nil, fmt.Errorf("ListController failed to look up queue %q: %w", req.Queue, err)
+	}
+
+	limit := int(req.PageSize)
+	if limit == 0 {
+		limit = defaultListPageSize
+	}
+	if limit > maxListPageSize {
+		limit = maxListPageSize
```
sbalabanov, line 97: I suggest to error instead

submitqueue/gateway/controller/list.go
```diff
@@ -0,0 +1,262 @@
...
+
+func encodeListPageToken(token listPageToken) (string, error) {
+	data, err := json.Marshal(token)
+	if err != nil {
+		return "", err
+	}
+	return base64.RawURLEncoding.EncodeToString(data), nil
+}
+
+func decodeListPageToken(raw string) (*listPageToken, error) {
```
sbalabanov, line 178: no need for a pointer

submitqueue/gateway/controller/list.go
```diff
@@ -0,0 +1,262 @@
...
+	if raw == "" {
+		return nil, nil
```
sbalabanov, line 180: why is this not an error?

submitqueue/gateway/controller/list.go
```diff
@@ -0,0 +1,262 @@
...
+
+func canonicalStatuses(raw []string) ([]entity.RequestStatus, error) {
+	seen := make(map[entity.RequestStatus]struct{}, len(raw))
+	var statuses []entity.RequestStatus
+	for _, statusRaw := range raw {
+		status := entity.RequestStatus(statusRaw)
+		if !entity.IsKnownRequestStatus(status) {
+			return nil, fmt.Errorf("unknown status %q: %w", statusRaw, ErrInvalidRequest)
+		}
+		if _, ok := seen[status]; ok {
```
sbalabanov, line 208: nit: can always assign to a set, then after the iteration completes, transform to a sorted list.

submitqueue/gateway/controller/list.go
```diff
@@ -0,0 +1,262 @@
...
+
+func statusesToStrings(statuses []entity.RequestStatus) []string {
+	out := make([]string, len(statuses))
+	for i, status := range statuses {
+		out[i] = string(status)
+	}
+	return out
+}
+
+func equalStrings(a, b []string) bool {
```
sbalabanov, line 237: could use `slices.Equal`

sbalabanov (changes requested):

sbalabanov:
user inpuit validation could be extracted into a separate function

albertywu:
Closing in favor of new design in https://github.com/uber/submitqueue/pull/314

PR #308, opened 2026-07-07
docs(rfc): add stovepipe process step design 

mnoah1:
Adding a more detailed design/implementation plan for MVP version of the `process` step in the overall stovepipe workflow.


doc/rfc/stovepipe/steps/process.md
```diff
@@ -0,0 +1,225 @@
...
+   - not found yet         -> retryable error (ingest write not visible; redelivery converges)
+2. If R.State is terminal (superseded / recorded-green / recorded-not-green):
+   - ack and return (idempotent no-op).
+3. If R.State is processing (strategy already recorded):
+   - re-publish R to build (the prior publish may have failed), ack, return.
+4. R.State is accepted. Load (or create) the Queue row Q.
+5. Coalesce: if R.Sequence < Q.latest_request_seq:
+   - a newer head exists -> mark R superseded, ack, return. (No slot consumed.)
+6. R is the latest head. Gate: if Q.in_flight_count >= Q.max_concurrent:
+   - park the delivery (extend visibility, no ack/nack, no state change) -> re-check until the slot frees (admit) or a newer head supersedes it. See [Waiting for a slot](#waiting-for-a-slot).
```
behinddwalls: do we have to worry about maxAttempt when we increase the visibility timeout or do we need to just do publishAfter semantics in this case? i think when we admit, are we sending for build step right away or waiting for previous run to complete

doc/rfc/stovepipe/steps/process.md
```diff
@@ -0,0 +1,225 @@
...
+## Entity model
+
+### Queue (new persisted entity)
+
+| Field | Role | Written by |
+|---|---|---|
+| `name` | Stable logical id (`monorepo/main`); the string ingest accepts | ingest (create) |
+| `last_green_uri` | Bookmark; empty until first green | record |
+| `in_flight_count` | Active Phase 1 validations | process (+1), record/DLQ (−1) |
+| `max_concurrent` | Concurrency cap (default 1) | config at create |
```
behinddwalls: do we need this in DB or should this be wire up into the Queue config or some sort of extension?

doc/rfc/stovepipe/steps/process.md
```diff
@@ -0,0 +1,225 @@
...
+| `latest_request_seq` | Highest request sequence ingested | ingest |
+| `version` | Optimistic-locking version | all writers |
+
+### Request (additions to the existing entity)
+
+| Field | Role |
+|---|---|
+| `Sequence` | The per-Queue counter value `n` behind the id; the coalescing order key |
+| `BuildStrategy` | `incremental_since_green` \| `full_monorepo`; immutable once set by `process` |
+| `BaselineURI` | Last-green URI used as the incremental baseline; empty for full builds |
```
behinddwalls: maybe just BaseURI?

doc/rfc/stovepipe/steps/process.md
mnoah1: True, we don't need this here and can have some sort of queue config provider for this type of configuration info.

mnoah1: Updated docs to refer to mirror similar queue config pattern from SubmitQueue instead of including in the DB row.

doc/rfc/stovepipe/steps/process.md
mnoah1: > do we have to worry about maxAttempt when we increase the visibility timeout

When we extend, it extends the exiting attempt and doesn't increment the number of attempts - so would keep the same attempt alive in the parked goroutine.  It would only become a new attempt of the current worker dies, the extended timeout is reached, and the task starts again.

> do we need to just do publishAfter semantics in this case

Yes - this seems like a good alternative for this use case.  I updated the docs to include both the alternative of a the park + extend pattern than I already had, plus this publishAfter pattern.  I would lean toward PublishAfter here because it will leave `process` unblocked and it can continue to supercede out of date messages as they come in, and only needs to re-enqueue the message for the latest head that we expect to build.  Docs include an example walkthrough.

behinddwalls (approved):

PR #309, opened 2026-07-07
feat(stovepipe): add Queue entity

mnoah1:
## Why?
Implementing `process` step.  

This change creates a Queue entity to keep track of queue-level state.  This matches the spec outlined in: https://github.com/uber/submitqueue/blob/main/doc/rfc/stovepipe/steps/process.md#queue-new-persisted-entity

## What?
Add the per-queue Queue entity that backs the process stage's concurrency gate and backlog coalescing.

## Test Plan
-Subsequent PRs will build out functionality that we can test

### DIFF
--- a/stovepipe/entity/queue.go
+++ b/stovepipe/entity/queue.go
@@ -1,0 +1,43 @@
+// Copyright (c) 2025 Uber Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package entity
+
+// Queue holds per-queue pipeline coordination state for a named repo+ref (e.g. "monorepo/main").
+type Queue struct {
+	// ****************
+	// Immutable fields, fixed at queue entity creation
+	// ****************
+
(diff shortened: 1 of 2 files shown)


behinddwalls:

**To fix manually:**
```bash
git fetch origin
git checkout mnoah1/stovepipe-queue-entity
git rebase --onto origin/main c9acfbf57688aa0ed815fdfd64868d58fb64d1aa mnoah1/stovepipe-queue-entity
# resolve conflicts, then:
git push --force-with-lease
```

```bash
gh pr edit 309 --base main
```

stovepipe/entity/queue.go
```diff
@@ -0,0 +1,43 @@
...
+
+	// Name is the stable logical id for the queue. It should be unique within the system.
+	Name string `json:"name"`
+
+	// ****************
+	// Following fields could be changed throughout the lifecycle of the queue
+	// ****************
+
+	// LastGreenURI is the most recent whole-repo-green commit. Empty until the first green.
+	LastGreenURI string `json:"last_green_uri"`
```
behinddwalls, line 32: do we want to call it "green"? or "SuccessfulURI"?

stovepipe/entity/queue.go
```diff
@@ -0,0 +1,43 @@
...
+
+	// InFlightCount is the number of Phase 1 validations admitted but not yet in a terminal state.
```
behinddwalls: do we need to say Phase 1 or should be just document the intent?

mnoah1: Updated

stovepipe/entity/queue.go
mnoah1: I would still lean towards "Green" here as long as we use it consistently - as I feel it conveys a more specific intent than "Successful".

For example, a validation run that catches a breakage is also successful in its own way, even though the repo is not green.

behinddwalls (approved):

PR #311, opened 2026-07-08
feat(orchestrator): Add pluggable validator extension for custom validation checks

ubettigole:
## Why?
Currently, there is not an easy way for integrators to add their own custom validations in the validate step. For example, an integrator may want to check for approvals, call outside services, etc. before allowing the merge to go through.

## What?
Added a validator extension that allows integrators to inject company-specific validation logic into the validate controller. The extension follows the existing Factory/Config pattern and includes a Compose utility for combining multiple validators.

## Test Plan
Unit tested


submitqueue/extension/validator/validator.go
```diff
@@ -0,0 +1,52 @@
...
+
+// Validator runs custom validation checks against a request and its changes.
+// Implementations are provided by integrators for company-specific checks
+// (e.g. org-policy enforcement, repo-specific rules). The built-in validate
+// controller invokes this after fetching change metadata and before claiming
+// URIs in the change store.
+type Validator interface {
+	// Validate runs custom validation on the request and its fetched change metadata.
+	// Returns nil if validation passes, error if the request should be rejected.
+	Validate(ctx context.Context, request entity.Request, changes []entity.ChangeInfo) error
```
behinddwalls: i think we should just limit it to `request` only in the contract and once can pull the changeInfo from change store if needed

submitqueue/extension/validator/validator.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+// Config carries the routing identity handed to a Factory.
+type Config struct {
+	// QueueName identifies the queue this request belongs to.
+	QueueName string
+	// Change is the code change being validated, carrying the full set of URIs.
+	// The URI scheme identifies the change provider (e.g. "github", "phabricator").
+	Change change.Change
```
behinddwalls: i don't think we need it here in Config?

submitqueue/extension/validator/compose.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+// Compose returns a Validator that runs all provided validators and joins their errors.
+func Compose(validators []Validator) Validator {
+	return compositeValidator(validators)
+}
+
+type compositeValidator []Validator
+
+func (c compositeValidator) Validate(ctx context.Context, request entity.Request, changes []entity.ChangeInfo) error {
```
behinddwalls: should we unify to using pointers

submitqueue/extension/validator/compose.go
```diff
@@ -0,0 +1,39 @@
...
+
+	"github.com/uber/submitqueue/submitqueue/entity"
+)
+
+// Compose returns a Validator that runs all provided validators and joins their errors.
+func Compose(validators []Validator) Validator {
+	return compositeValidator(validators)
+}
+
+type compositeValidator []Validator
```
behinddwalls: should it follow the New constructor pattern

service/submitqueue/orchestrator/server/main.go
```diff
@@ -537,6 +538,22 @@ func (f analyzerFactory) For(cfg conflict.Config) (conflict.Analyzer, error) {
 	return f.reg.get(cfg.QueueName).analyzer, nil
 }
 
+// noopValidatorFactory is an example validator.Factory that always returns a no-op validator (validation passes).
+// Replace with a real implementation to enforce company-specific policies (e.g. org-policy, repo-level rules).
+type noopValidatorFactory struct{}
+
+// For returns a no-op validator (validation passes).
+// In a real implementation, it could choose which validator(s) to return, based on the specific queue or change type.
+func (noopValidatorFactory) For(validator.Config) (validator.Validator, error) {
```
behinddwalls: we can probably make noop impl default and add it as an impl

ubettigole: Added a fake impl, following the pattern of the scorerFake

behinddwalls (approved):

PR #313, opened 2026-07-08
feat(stovepipe): add QueueStore interface, mock, and MySQL implementations

mnoah1:
## Why?
Implementing `process` step per the design in https://github.com/uber/submitqueue/blob/main/doc/rfc/stovepipe/steps/process.md 

## What?
Storage setup for Stovepipe's new Queue entity:
- Introduce the per-queue coordination store backing the process stage: GetOrCreate, Get, and Update.
- Wire QueueStore into the Storage aggregator and add the queue table schema.
- Include MySQL integration tests for create, idempotent get-or-create, CAS updates, and version mismatch handling.

## Test Plan
- Integration tests
- Next few PRs will begin to incorporate this logic and we can test it locally


stovepipe/extension/storage/queue_store.go
```diff
@@ -0,0 +1,43 @@
...
+
+	"github.com/uber/submitqueue/stovepipe/entity"
+)
+
+// QueueStore persists per-queue coordination rows, keyed by queue name.
+type QueueStore interface {
+	// GetOrCreate returns the queue row for name, creating it with defaults if absent.
+	// defaults supplies initial field values for a new row (Name is set from name).
+	// On a create race, re-reads and returns the canonical row.
+	GetOrCreate(ctx context.Context, name string, defaults entity.Queue) (entity.Queue, error)
```
behinddwalls: should this just do one thing which is create? and let the logic be in control for get or create part

behinddwalls (changes requested):

stovepipe/extension/storage/queue_store.go
mnoah1: Updated - callers will manage this logic instead.

behinddwalls (approved):

test/integration/stovepipe/extension/storage/suite.go
```diff
@@ -0,0 +1,217 @@
...
+
+	got, err := s.queueStore.Get(s.ctx, name)
+	require.NoError(t, err)
+	assert.Equal(t, entity.Queue{
+		Name:             name,
+		LatestRequestSeq: 0,
+		Version:          1,
+	}, got)
+
+	s.log.Logf("Create passed: created queue %s", name)
```
sbalabanov, line 74: why need such logging in tests?

mnoah1: Looks like it's just carrying over the pattern from existing tests, but I agree logging on a purely "passed" test case case is redundant.  Created https://github.com/uber/submitqueue/pull/355 if we want to clean this up throughout the repo.

PR #314, opened 2026-07-08
[1/N Status & List APIs] docs: RFC for Status and List APIs

albertywu:
## Summary
Add a SubmitQueue RFC defining gateway Status and List API contracts and general design principles.

## Test Plan

## Stack
1. @ #314
1. #321
1. #322
1. #327

### DIFF
--- a/doc/rfc/submitqueue/status-list-api.md
+++ b/doc/rfc/submitqueue/status-list-api.md
@@ -1,0 +1,171 @@
+# Gateway Status and List APIs
+
+## RFC Status
+
+Proposed.
+
+This RFC replaces the unimplemented design in [list-api.md](list-api.md). It defines the gateway-owned request context and materialized status model used by both `Status` and `List`.
+
+## Problem
+
+The gateway currently stores an append-only request log and reconciles it at read time for `Status`. Request-log entries contain status, error, and display metadata, but they do not contain the immutable request context needed by user-facing APIs: queue name, submitted change URIs, and the time the gateway received the request.
+
+The required read patterns are:
+
+1. Look up the current status and request context by sqid.
+2. Look up the current status and request context by change URI.
+3. List requests received for one queue during a bounded time range.
+
+The append-only request log is not shaped for the second or third query. Serving `List` by scanning and reconciling logs at runtime would also make query cost proportional to historical log volume. The gateway therefore needs a write-time materialized read model in addition to the request log.
+
+## Decisions
+
+1. `Status` accepts exactly one selector: sqid or change URI.
+2. A change URI lookup returns all requests containing that exact URI, ordered by receipt time descending.
+3. Change URIs are treated as globally meaningful identifiers. `Status` does not require a queue when selecting by URI.
+4. `List` accepts one queue and a required receipt-time range. It does not accept sqid or URI selectors.
+5. The `List` time range is based only on gateway receipt time, not lifecycle overlap.
+6. `Status` and `List` return the same materialized current state and immutable request context. Neither endpoint returns the request-log timeline.
+7. Materialization happens on write. Reads do not reconcile the append-only log at request time.
+8. The data model uses separate query-shaped stores rather than secondary indexes.
+9. Existing rows are not backfilled. Requests received before rollout may be unavailable through the new read model.
+10. Retention and pruning are outside the scope of this RFC.
+
+## Vocabulary
+
+`sqid` is the public API name for a SubmitQueue request identifier. Storage contracts may call the same value `request_id`.
+
+`received_at_ms` is the immutable gateway timestamp captured after Land passes synchronous request validation and the gateway begins persisting the request. It is not the time at which the orchestrator validates, accepts, or completes the request.
+
+## API Contract
+
+### Status
+
+`Status` requires exactly one selector: sqid or change URI. An unset selector or an explicitly selected empty value is invalid.
+
+An sqid lookup returns exactly one request summary when found. A change URI lookup returns every matching request summary ordered by `(received_at_ms DESC, sqid DESC)`. The sqid tie-breaker makes the result deterministic when requests share a millisecond.
+
+Both selector modes return the same response shape: a list of request summaries. Each summary contains sqid, queue, change URIs, receipt time, current status, last error, and display metadata.
+
+The URI result is intentionally unpaginated and has a hard maximum of 100 requests. A change is expected to have only a small number of SubmitQueue requests, generally fewer than one hundred. Exceeding the maximum returns an error rather than silently truncating the result.
+
+### List
+
+`List` is a queue-scoped request receipt history query. It requires a queue and a bounded half-open receipt-time interval. Both bounds are required, and the start must be less than the end.
+
+Results are ordered by `(received_at_ms DESC, sqid DESC)`, so the newest requests are returned first. Pagination uses an opaque token bound to the original queue and time range plus the last returned ordering tuple. Reusing a token with different query parameters is invalid.
+
+Page size is optional and subject to a server default and maximum.
+
+New requests may appear before the first page while a caller is paging. The immutable tuple cursor prevents duplicates within the caller's traversal.
+
+The first version does not filter by current status. Current status is mutable, while the queue receipt projection is ordered by immutable receipt time. Efficient status filtering requires an additional application-maintained membership projection keyed by queue, status, receipt time, and sqid. That projection is deferred until a concrete server-side filtering use case justifies its write and reconciliation cost. Clients may filter a returned page for presentation, but client-side filtering is not equivalent to a server-side filtered query.
+
+## Errors
+
+Invalid selectors, invalid time ranges, invalid page sizes, invalid page tokens, and URI results above the hard maximum are user errors. A lookup with no matching request returns a not-found user error.
+
+The transport representation of these errors follows the gateway-wide RPC error contract and is outside the scope of this RFC. This RFC does not introduce an endpoint-specific wire error model.
+
+## Gateway-Owned Read Model
+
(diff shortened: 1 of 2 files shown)


doc/rfc/submitqueue/status-list-api.md
```diff
@@ -0,0 +1,285 @@
...
+The interval is half-open: `[admitted_at_or_after_ms, admitted_before_ms)`. Both bounds are required, and the gateway rejects a range whose start is not less than its end.
+
+Results are ordered by `(admitted_at_ms DESC, sqid DESC)`, so the newest admissions are returned first. The first API version does not expose alternate sort modes or status filters. This filtering can easily be done on the consumer end. The gateway applies a default page size when `page_size` is unset or <= 0, and caps `page_size` to a fixed maximum.
+
+The page token is opaque and binds the original queue and time bounds to the last returned `(admitted_at_ms, sqid)` tuple. A token used with different query parameters is invalid. Newly admitted requests may appear before the first page while a caller is paging, but the tuple cursor prevents duplicates within the caller's traversal.
+
+`RequestNotFoundError` adds a `change_uri` field while retaining the existing sqid field number. Exactly one identifier is populated to match the selector used by the caller.
+
+```proto
+message RequestNotFoundError {
```
behinddwalls: we may want to think about errors more holistically for RPCs. Saw this PR: https://github.com/uber/tango/pull/161/changes for tango (haven't reviewed) but something more common for all APIs

doc/rfc/submitqueue/status-list-api.md
```diff
@@ -0,0 +1,285 @@
...
+}
+```
+
+The URI result is intentionally unpaginated. The gateway enforces a hard maximum of 100 matched requests for `change_uri`; if the limit is exceeded, it returns an error.
+
+### List
+
+`List` is a queue-scoped admission history query.
+
+```proto
```
behinddwalls: i think RFC contains very low level detail, maybe we just talk about logical component and integrations and here and leave the exact API spec/schema for the impl PRs

doc/rfc/submitqueue/status-list-api.md
```diff
@@ -0,0 +1,285 @@
...
+```sql
+CREATE TABLE request_summary (
+    request_id VARCHAR(255) NOT NULL,
+    queue VARCHAR(255) NOT NULL,
+    change_uris JSON NOT NULL,
+    admitted_at_ms BIGINT NOT NULL,
+    status VARCHAR(64) NOT NULL,
+    request_version INT NOT NULL,
+    status_timestamp_ms BIGINT NOT NULL,
+    winner_is_versioned_terminal BOOLEAN NOT NULL,
```
behinddwalls: do we need to store it?

doc/rfc/submitqueue/status-list-api.md
```diff
@@ -0,0 +1,285 @@
...
+`request_summary` is the authoritative materialized row for `Status` by sqid and the source used to resolve later status updates.
+
+The API term `sqid` and the storage column name `request_id` refer to the same SubmitQueue request identifier throughout this RFC.
+
+```sql
+CREATE TABLE request_summary (
+    request_id VARCHAR(255) NOT NULL,
+    queue VARCHAR(255) NOT NULL,
+    change_uris JSON NOT NULL,
+    admitted_at_ms BIGINT NOT NULL,
```
behinddwalls: should we say received_at instead? admitted implies we accepted which happens way later in the workflow?

doc/rfc/submitqueue/status-list-api.md
```diff
@@ -0,0 +1,285 @@
...
+The row duplicates response data deliberately. `List` performs one bounded table scan and does not issue one `request_summary` read per result. A status update already knows the queue and admission time from `request_summary`, so it can update the exact queue projection row by full primary key. `projection_version` prevents a delayed writer from replacing a newer queue projection.
+
+### Requests by Change URI
+
+`request_by_change_uri` is an immutable reverse mapping used to resolve a URI to its request IDs in admission order.
+
+```sql
+CREATE TABLE request_by_change_uri (
+    change_uri VARCHAR(255) NOT NULL,
+    admitted_at_ms BIGINT NOT NULL,
```
behinddwalls: do we need admitted here again?

doc/rfc/submitqueue/status-list-api.md
```diff
@@ -0,0 +1,285 @@
...
+```
+
+The immutable context columns are written during Land admission and are never changed. The status columns hold the winning materialized request-log state. `request_version`, `status_timestamp_ms`, and `winner_is_versioned_terminal` retain the comparison state needed to process an incoming log entry without rereading historical logs. `projection_version` is the optimistic-lock version for the materialized row and is independent of the orchestrator request version.
+
+### Request Summary by Queue
+
+`request_summary_by_queue` supports the complete `List` query through a primary-key prefix and range scan.
+
+```sql
+CREATE TABLE request_summary_by_queue (
```
behinddwalls: i guess we won't be able to filter by state?

doc/rfc/submitqueue/status-list-api.md
albertywu: oops, removed this

doc/rfc/submitqueue/status-list-api.md
albertywu: Good point. I removed the endpoint-specific RequestNotFoundError wire change. The RFC now defines the semantic user-error cases only and leaves their transport representation to a gateway-wide RPC error contract. That can build on SubmitQueue’s existing platform/errs classification rather than introducing a separate error model in this API RFC.

doc/rfc/submitqueue/status-list-api.md
albertywu: Agreed. I kept the externally visible query behavior because it determines the required access patterns, but removed the canonical proto, SQL DDL, SQL query, and step-by-step CAS details. The RFC now describes logical API behavior, read-model key shapes, ownership, and consistency guarantees; exact interfaces and schemas can live in the implementation PR.

doc/rfc/submitqueue/status-list-api.md
albertywu: Yes, received_at is clearer. The timestamp is captured by the gateway after synchronous Land validation and before the request enters the orchestrator workflow, so admitted could imply a later lifecycle decision. I renamed the concept consistently to received_at and defined that boundary in the vocabulary section.

doc/rfc/submitqueue/status-list-api.md
albertywu: I think receipt time is needed in the URI lookup key because Status-by-URI promises newest-first results with a hard result bound. Keeping received_at in the immutable mapping allows a bounded ordered scan before resolving summaries. Without it, the controller would have to fetch and sort every mapped request before enforcing the maximum. I added this rationale to the RFC.

doc/rfc/submitqueue/status-list-api.md
```diff
@@ -0,0 +1,165 @@
...
+### Request Summary by Sqid
+
+The authoritative request summary is keyed by sqid. It contains immutable request context plus the current materialized request-log winner and the reconciliation state needed to compare a later log entry without rereading historical logs.
+
+The immutable context is queue, change URIs, and receipt time. The mutable response state is status, last error, and metadata. Optimistic-lock state is internal to the projection and is not part of the API response.
+
+### Request Summaries by Queue
+
+The queue projection is keyed and ordered by `(queue, received_at_ms, sqid)`. This key supports one bounded queue and receipt-time range query with deterministic keyset pagination.
+
```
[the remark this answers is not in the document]
albertywu: Good catch. I clarified the backend contract rather than assuming SQL-style reverse scans: the queue projection must support a bounded descending scan over receipt time and sqid. A backend may implement that with a reverse range scan or descending-encoded key components, while cursors retain the original values.

doc/rfc/submitqueue/status-list-api.md
```diff
@@ -0,0 +1,165 @@
...
+The row duplicates the complete `List` response deliberately. A page is served by one bounded range scan rather than one follow-up authoritative-summary read per result. Status updates propagate from the authoritative sqid summary to this projection.
+
+### Requests by Change URI
+
+The URI reverse mapping is keyed and ordered by `(change_uri, received_at_ms, sqid)`. It contains immutable lookup data and does not duplicate mutable status fields.
```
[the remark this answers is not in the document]
albertywu: Agreed. I added the same explicit descending-scan contract for the URI mapping. Backends may use reverse scans or descending-encoded key components, but the exposed ordering remains received_at_ms DESC, sqid DESC.

doc/rfc/submitqueue/status-list-api.md
```diff
@@ -0,0 +1,165 @@
...
+
+## Read Flows
+
+### Status by Sqid
+
+The gateway reads the authoritative summary by sqid and returns its current materialized state and immutable context. It does not fall back to parsing the sqid, reconciling logs, or reading orchestrator stores.
+
+### Status by Change URI
+
+The gateway performs a bounded newest-first scan of the URI reverse mapping and then reads the authoritative summary for each resolved sqid. Results preserve the mapping order. No mapping is a not-found result; exceeding the 100-request maximum is an error.
```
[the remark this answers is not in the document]
albertywu: Agreed. A URI mapping with no authoritative sqid summary now explicitly fails the entire lookup as an internal consistency error. It is not classified as user-facing not-found and is not silently omitted.

doc/rfc/submitqueue/status-list-api.md
albertywu: Correct; prefer to keep the design simpler here and let clients filter by status

behinddwalls (approved):

PR #315, opened 2026-07-08
feat(speculation): add enumerator + dependency-limit extensions

behinddwalls:
## Summary
Add the first speculation seam and its limit counterpart from the speculation RFC, as vendor-agnostic extension interfaces under submitqueue/extension/speculation/.

enumerator: given a batch and its active dependencies ordered oldest-first (queue arrival order — the ordering that fixes which assumed-good Base prefixes exist), mechanically lists the candidate Base/Head speculation paths — pure, deterministic, and purely structural. It returns only []entity.SpeculationPath: identity, status, and score are controller-owned, and the controller wraps each path into the persisted tree entry (SpeculationPathInfo), keeping a clean ownership boundary between the seam and tree state.

dependencylimit: the "how much" policy bounding how many active (in-flight) dependencies a batch may speculate over. It is the eligibility gate before enumeration; unlike the other speculation limits it is controller-held rather than injected into a seam, keeping the enumerator pure. The value is signal-driven, not a fixed constant.

Each follows the repo extension contract (conflict.Analyzer reference shape): Factory.For(Config) (T, error) with Config carrying only QueueName; behavioral knobs and limit signals are integrator-injected at construction. Includes READMEs, gomock packages, and Makefile mock-gen wiring. Interfaces only; concrete impls and controller wiring are deferred.

## Stack
1. #337
1. @ #315
1. #316
1. #317
1. #320
1. #331
1. #332
1. #333


submitqueue/extension/speculation/enumerator/enumerator.go
```diff
@@ -0,0 +1,63 @@
...
+type Enumerator interface {
+	// Enumerate returns the speculation tree structure for the batch identified
+	// by batchID, given its active dependency batches in arrival order. Each
+	// returned path carries a Base/Head split only: Score and Status are left
+	// unset — the controller stamps Status on persist and calls the scorer to
+	// fill Score.
+	//
+	// Enumeration is pure and deterministic: the same (batchID, deps) always
+	// yields the same tree, so callers may regenerate safely.
+	Enumerate(ctx context.Context, batchID string, deps []entity.Batch) (entity.SpeculationTree, error)
```
albertywu:
```suggestion
	Enumerate(ctx context.Context, batch entity.Batch, deps []entity.Batch) (entity.SpeculationTree, error)
```

Should we pass `entity.Batch` for consistency / symmetry with deps?

Alternatively:

```suggestion
	Enumerate(ctx context.Context, batchID string, dependencyIDs []string) (entity.SpeculationTree, error)
```

if all we care about is IDs

behinddwalls: yes it should take the batch..i guess claude didn't do what i wanted..

behinddwalls: Done — `Enumerate` now takes `entity.Batch` (impls, fake, mock, and the controller call site updated up the stack).

submitqueue/extension/speculation/enumerator/enumerator.go
```diff
@@ -0,0 +1,62 @@
...
+// speculate over.
+type Enumerator interface {
+	// Enumerate returns the speculation tree structure for the batch, given its
+	// active dependency batches in arrival order. Each returned path carries a
+	// Base/Head split only: Score and Status are left unset — the controller
+	// stamps Status on persist and calls the scorer to fill Score.
+	//
+	// Enumeration is pure and deterministic: the same (batch, deps) always
+	// yields the same tree, so callers may regenerate safely.
+	Enumerate(ctx context.Context, batch entity.Batch, deps []entity.Batch) (entity.SpeculationTree, error)
```
albertywu: > Score and Status are left unset

^ entity.SpeculationTree contains controller-owned `Score`, `Status` info. Should we return `[]entity.SpeculationPath` and let the controller construct the persisted tree so there's clear ownership boundaries?

submitqueue/extension/speculation/enumerator/enumerator.go
```diff
@@ -0,0 +1,62 @@
...
+// does not score paths — that is the scorer's job (see
+// extension/speculation/scorer), which the controller re-runs on every
+// respeculate — it does not decide which paths to build — that is the selector's
+// job (see extension/speculation/selector) — it does not set path status, and it
+// does not decide how far back to speculate: the controller gates on the
+// dependency limit and hands Enumerate exactly the active dependencies to
+// speculate over.
+type Enumerator interface {
+	// Enumerate returns the speculation tree structure for the batch, given its
+	// active dependency batches in arrival order. Each returned path carries a
```
albertywu: > in arrival order

^ unclear what arrival order of batches has to do with enumeration?

submitqueue/extension/speculation/enumerator/fake/fake.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+// New returns an empty fake Enumerator. Seed it with Set.
+func New() *Enumerator {
+	return &Enumerator{trees: map[string]entity.SpeculationTree{}}
+}
+
+// Set seeds the tree returned by Enumerate for the given batch ID.
+func (e *Enumerator) Set(batchID string, tree entity.SpeculationTree) *Enumerator {
+	e.trees[batchID] = tree
```
albertywu: `e.trees[batchID] = cloneTree(tree)`

^ prevents callers from mutating the tree stored inside the fake

submitqueue/extension/speculation/enumerator/fake/fake.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+// Enumerate returns the seeded tree for the batch. An unseeded batch returns an
+// empty tree carrying the batch's identity. The deps argument is ignored.
+func (e *Enumerator) Enumerate(_ context.Context, batch entity.Batch, _ []entity.Batch) (entity.SpeculationTree, error) {
+	if e.err != nil {
+		return entity.SpeculationTree{}, e.err
+	}
+	if tree, ok := e.trees[batch.ID]; ok {
+		return tree, nil
```
albertywu: `return cloneTree(tree), nil)`

^ prevents callers from mutating the tree stored inside the fake

submitqueue/extension/speculation/enumerator/enumerator.go
behinddwalls: enumeration is dependent on the arrival order to construct the tree, without it, there is no way to say how they depend on each other

submitqueue/extension/speculation/enumerator/fake/fake.go
behinddwalls: i think we decided early on to not worry about it

submitqueue/extension/speculation/enumerator/fake/fake.go
behinddwalls: same as above

submitqueue/extension/speculation/enumerator/enumerator.go
behinddwalls: we could do that

behinddwalls: Done — `Enumerate` now returns `[]entity.SpeculationPath`: structure only, nothing else. The controller owns the whole persisted entry — it wraps each path into `SpeculationPathInfo`, stamps Status, and assigns the path's ID (introduced down-stack in #337), so the ownership boundary is now exactly the one you described. Impls (chain, fallback), fake, mock, and the createTree call site up-stack are all updated, and createTree now also dedups structural duplicates from an enumerator as a contract violation.

submitqueue/extension/speculation/enumerator/enumerator.go
behinddwalls: Done — reworded the doc to say why order matters rather than just asserting it: deps are handed oldest-first (queue arrival order), and that ordering is load-bearing because a path's Base is an assumed-good prefix of predecessors applied in order — the deps ordering fixes which Base prefixes exist and the order of batch IDs within each. Same clarification in the README.

albertywu (approved):

PR #316, opened 2026-07-08
feat(speculation): add path scorer extension

behinddwalls:
Add the scorer seam from the speculation RFC, as a vendor-agnostic extension interface under submitqueue/extension/speculation/pathscorer/.

The scorer computes each speculation path's predicted-success score from the current state: the per-batch scores of the path's base batches (entity.Batch.Score) and which of those dependencies have resolved (landed or build-passed), plus optionally other signals. It is a prediction over live state, so the controller re-runs it on every respeculate right after reconciling status, and persists the result; the scorer owns only the formula.

The controller hands it the batch's speculation tree directly — the subject it scores. Any richer signal an implementation needs (dependency batch scores, historical pass rates) is injected at its Factory, not put in the signature. It never writes: its only output is per-path scores ([]entity.PathScore — path ID plus fresh score, an entity-level seam-output type alongside the path-decision type), which the controller merges into the tree and persists, staying the single writer of tree state; structure and status never pass through the scorer. Scores are probabilities in [0, 1] — the contract every implementation must satisfy, enforced by the controller on consume.

This is the per-path scorer, distinct from the existing per-batch score stage (`extension/scorer`) that sets entity.Batch.Score — the path scorer consumes those to score whole paths.

Follows the repo extension contract: Factory.For(Config) (Scorer, error) with Config carrying only QueueName. Includes README, gomock package, and a programmable fake. The speculation RFC's seam descriptions are updated to match the identity-keyed minimal-output contracts (and gain a design-decision entry for assigned path identity). Interface only; concrete impls and controller wiring are deferred.

## Stack
1. #337
1. #315
1. @ #316
1. #317
1. #320
1. #331
1. #332
1. #333


submitqueue/extension/speculation/pathscorer/pathscorer.go
```diff
@@ -0,0 +1,73 @@
...
+// pass (resolved assumptions raise confidence), and optionally other signals
+// (how long the batch has waited, historical pass rates).
+//
+// The controller re-runs the scorer on every respeculate, right after it
+// reconciles path status — so when a dependency lands, its build passes, or a
+// sibling path fails, the surviving paths' scores are recomputed against the new
+// reality before anything is selected or prioritized. The controller drives
+// *when* to rescore and persists the result; the scorer owns the *formula*.
+//
+// This is the per-*path* scorer, distinct from the per-*batch* score stage
```
sbalabanov, line 42: may be name it somewhat differently then

sbalabanov (approved):

submitqueue/extension/speculation/pathscorer/pathscorer.go
behinddwalls: Renamed the package to `speculation/pathscorer` (interface stays `Scorer`, read as `pathscorer.Scorer`) — removes the collision with the per-batch `extension/scorer` and the import aliases it forced. Also updated the RFC's reference.

submitqueue/extension/speculation/pathscorer/pathscorer.go
```diff
@@ -0,0 +1,73 @@
...
+
+	"github.com/uber/submitqueue/submitqueue/entity"
+)
+
+// Scorer computes the predicted-success score of every path in a batch's
+// speculation tree.
+//
+// A path's score is a prediction — "how likely is this bet to pay off?" — and
+// predictions must move as evidence arrives. The scorer answers "how good is
+// each path right now" from the current state: the per-batch success
```
albertywu, line 30: Is our default implementation going to try to score each path independently, or assign scores with the intention to stack-rank all paths?

submitqueue/extension/speculation/pathscorer/pathscorer.go
```diff
@@ -0,0 +1,73 @@
...
+// subject it scores. Any richer signal an implementation needs (the dependency
+// batches' scores, historical pass rates) is injected at its Factory, not passed
+// here. It never writes: it returns the scored tree and the controller persists
+// the scores, keeping the controller the single writer of tree state.
+type Scorer interface {
+	// Score returns the given tree with each path's Score set to its freshly
+	// computed predicted-success value. Path structure and controller-stamped
+	// Status are carried through unchanged; only Score is (re)computed. The
+	// combination formula is the implementation's concern.
+	Score(ctx context.Context, tree entity.SpeculationTree) (entity.SpeculationTree, error)
```
albertywu: The controller consumer doesn't need the entire speculation tree in the response right? Could we just return what's needed (path scores) instead of the whole tree:

```
type Scorer interface {
	Score(
		ctx context.Context,
		tree entity.SpeculationTree,
	) ([]PathScore, error)
}
```

```
type PathScore struct {
	PathID string // ID from SpeculationPath
	Score float32
}
```

```
type SpeculationPath struct {
	// ID is deterministically derived from Head and ordered Base.
    // A SpeculationTree can never have duplicate SpeculationPaths (invariant)
	ID string

	Base []string
	Head string
}
```

submitqueue/extension/speculation/pathscorer/pathscorer.go
behinddwalls: yes..path scorer is baysian multiplication on success/failure scores of each batch

submitqueue/extension/speculation/pathscorer/pathscorer.go
behinddwalls: Yes. ID is already part of path in down stacks

behinddwalls: Done — `Score` now returns `[]PathScore` ({PathID, Score}) and nothing else; the controller merges them into the tree by ID (unknown IDs logged+skipped, out-of-[0,1] clamped with a warning) and stays the single writer. Path identity itself moved down the stack per this thread: new bottom PR #337 introduces `SpeculationPathInfo.ID` — assigned by the controller at first persist, immutable, unique within the tree, opaque (never parsed) — so every seam PR above it can build on it. The probability scorer (#333), fake, and mock are updated, and the [0,1] contract is documented on the interface and README.

albertywu (approved):

PR #317, opened 2026-07-08
feat(speculation): add selector + selection-limit extensions

behinddwalls:
Add the selector seam and its limit counterpart from the speculation RFC, as vendor-agnostic extension interfaces under submitqueue/extension/speculation/.

selector: the controller hands it the batch's speculation tree; it returns per-path decisions (Promote/Cancel), each naming a path by its ID (entity.SpeculationPathInfo.ID), for the paths it chooses to act on — at most one decision per path, with conflicting duplicates treated by the controller as a policy bug (applied first, logged and skipped after). It is the policy seam — it reads the tree it is given and emits decisions only, never writing status, and does not decide merging. The controller maps each decision to a guarded status transition (Promote → Selected, Cancel → Cancelling/Cancelled) and persists it, staying the single writer of tree state. entity.PathDecision correspondingly names paths by PathID rather than restating the Base/Head split.

selectionlimit: the "how much" policy bounding how many paths a batch may build in parallel. It is the selector's companion — the selector decides which paths, the limit decides how many. Injected into the selector at construction and called by it, keeping the selector interface limit-free; the value is dynamic, not a fixed constant.

## Stack
1. #337
1. #315
1. #316
1. @ #317
1. #320
1. #331
1. #332
1. #333


submitqueue/extension/speculation/selectionlimit/README.md
```diff
@@ -0,0 +1,17 @@
+# Speculation Selection Limit
+
+Vendor-agnostic "how much" policy that bounds how many paths a batch may build in parallel.
+
+See the [Speculation RFC](../../../../doc/rfc/submitqueue/speculation.md) for the end-to-end design and how limits fit into the two-layer speculation model.
```
sbalabanov: do relative to root path instead

sbalabanov (commented):
can you define sentinel errors as a part of contract - or there aren't planned any?

submitqueue/extension/speculation/selectionlimit/README.md
behinddwalls: Done — switched to a root-relative link (`/doc/rfc/submitqueue/speculation.md`), and swept the same fix through all seven speculation READMEs across the stack (#315/#316/#320 included).

submitqueue/extension/speculation/selector/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+The selector decides only where to spend build resources. It does **not** decide merging: a path becomes mergeable when its build passed and its base matches what actually landed, which is deterministic, not a policy choice — so the controller finalizes it on its own.
+
+The controller hands the selector the batch's **speculation tree** directly — the subject it decides over. The controller is the single writer: it reconciles each path's status (candidate, selected, prioritized, building, passed, failed, cancelling, cancelled) from the latest builds and dependency states, and it maps each of the selector's decisions to a status transition — `Promote` → `Selected`, `Cancel` → `Cancelling` (or `Cancelled` when nothing is building) — applied under the tree's optimistic lock and persisted. The selector's only output is decisions; it **never** writes status. This keeps it a deterministic policy over the tree it is given.
```
albertywu: is `prioritized` for paths that we want to eventually build later but don't have resources at the moment to build? If we had resources would the path be `selected` instead?

submitqueue/extension/speculation/selector/selector.go
```diff
@@ -0,0 +1,64 @@
...
+// of tree state. This keeps the selector a deterministic policy over the tree it
+// is given. Policy knobs such as a top-K limit or budget belong to the
+// implementation's construction, not this method.
+type Selector interface {
+	// Select returns the decisions to take for the given tree. Returning multiple
+	// Promote decisions selects several paths for building in parallel — each
+	// waits for the prioritizer to clear it under the queue's build budget before
+	// a build is dispatched. An empty result means nothing should be done right
+	// now. Paths the selector has no opinion on are simply omitted (leave-as-is).
+	Select(ctx context.Context, tree entity.SpeculationTree) ([]entity.SpeculationPathDecision, error)
```
albertywu: Since the response is a slice, you could potentially have conflicting decisions for the same path returned. Perhaps use a map response shape that would be immune to that error class

```
type SelectorDecisions map[string]entity.SpeculationPathAction // string is an ID for a path

type Selector interface {
	Select(
		ctx context.Context,
		tree entity.SpeculationTree,
	) (SelectorDecisions, error)
}
```

submitqueue/extension/speculation/selector/README.md
behinddwalls: yes... and more over we can only prioritize for the resources we have and that's it.

submitqueue/extension/speculation/selector/selector.go
behinddwalls: possible but i think slice would work as well with ID... map just enforces key to have ID so not any different in that case.

behinddwalls: Kept the slice, but the error class is now closed on both ends: decisions name paths by `PathID` (no more restated Base/Head), the contract says at most one decision per path, and both decision appliers (speculate's applySelection, prioritize's applyDecisions) treat a duplicate for the same path as a policy bug — first one applies, the rest are logged and skipped, same as any other illegal decision. So a conflicting pair can't corrupt tree state; a map would just move the last-wins choice into the seam silently, and the slice keeps output ordering deterministic for tests.

submitqueue/extension/speculation/selector/README.md
behinddwalls: Done — the README now spells this out: `Selected` is per-batch desire (wants a build slot), `Prioritized` is the queue-wide grant (admitted under the shared build budget, cleared to build). A promoted path waits in Selected until the prioritizer clears it — promptly when budget is free, possibly indefinitely (or dropped) under contention.

albertywu (approved):

PR #318, opened 2026-07-08
feat(prioritization): add prioritizer + prioritization-limit extensions

behinddwalls:
Add the prioritizer seam and its limit counterpart from the speculation RFC, as vendor-agnostic extension interfaces under submitqueue/extension/prioritization/.

Prioritization is queue-wide and lives at the build stage, not per batch inside speculate, so it sits in its own prioritization/ family rather than under speculation/.

prioritizer: the queue-wide policy that rations a shared build budget across every in-flight batch. Bound to its queue at construction, it takes no arguments — the store is the source of truth, so it reads the queue's pending builds and each build's path score from storage (read access injected at its Factory), ranks them by score plus any fairness policy, and returns the admitted subset that fits the queue's concurrent-build budget. It never writes; dispatching the admitted builds is the controller's job.

limit: the "how much" policy bounding how many builds the queue runs at once — the queue's concurrent-build budget and the ultimate cap on speculation's demand on CI. Injected into the prioritizer at construction and called by it; the value is dynamic, not a fixed constant.

## Stack
1. #231
1. #315
1. #316
1. #317
1. @ #318


behinddwalls:
Superseded by #320 — the prioritizer moved under submitqueue/extension/speculation/ (branch renamed to preetam/ext/speculation-prioritizer).

PR #319, opened 2026-07-08
feat(stovepipe): extend Request entity to support new process step

mnoah1:
## What?
Updating the Request entity to accommodate upcoming functionality in the process step:
- New `RequestState` values - `processing` once the processing step has started, and `superseded` for requests that will be skipped in favor of building a newer one
- BuildStrategy/BaseURI fields which will get populated by the process process for downstream use during builds
- Since request IDs are minted in order by ingest, I realized it's a bit redundant to store a separate sequence ID in the DB.  Instead, define a `CompareRequestID` function for cases that require ID comparison, and converge to just storing the ID.

## Why?
Implementing behavior in https://github.com/uber/submitqueue/blob/main/doc/rfc/stovepipe/steps/process.md - which introduces two new outcome states and includes logic to determine the build strategy for a request. 

## Test Plan
- Unit tests
- Subsequent changes will pull this logic together into an end to end flow

### DIFF
--- a/stovepipe/entity/request.go
+++ b/stovepipe/entity/request.go
@@ -29,8 +29,25 @@ const (
 	RequestStateUnknown RequestState = ""
 	// RequestStateAccepted is the initial state of a request: a new commit has been observed
 	// for the queue and the request has been admitted into the pipeline, but no validation
-	// strategy has been chosen yet. Later stages (process, build, record) add their own states.
+	// strategy has been chosen yet.
 	RequestStateAccepted RequestState = "accepted"
+	// RequestStateProcessing means process admitted the request, recorded build strategy and
+	// baseline, and published to build.
+	RequestStateProcessing RequestState = "processing"
+	// RequestStateSuperseded means process skipped the request because a newer head exists.
+	RequestStateSuperseded RequestState = "superseded"
+)
+
+// BuildStrategy defines how build validates the request's commit.
+type BuildStrategy string
+
+const (
+	// BuildStrategyUnknown is the zero value before process chooses a strategy.
+	BuildStrategyUnknown BuildStrategy = ""
+	// BuildStrategyIncrementalSinceGreen validates only the delta since the pinned baseline URI.
+	BuildStrategyIncrementalSinceGreen BuildStrategy = "incremental_since_green"
+	// BuildStrategyFull validates the whole repo from scratch.
+	BuildStrategyFull BuildStrategy = "full"
 )
 
 // Request represents a single validation of a queue at a particular commit. The queue reports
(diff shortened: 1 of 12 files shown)


stovepipe/entity/request.go
```diff
@@ -29,8 +29,25 @@ const (
...
+// BuildStrategy defines how build validates the request's commit.
+type BuildStrategy string
+
+const (
+	// BuildStrategyUnknown is the zero value before process chooses a strategy.
+	BuildStrategyUnknown BuildStrategy = ""
+	// BuildStrategyIncrementalSinceGreen validates only the delta since the pinned baseline URI.
+	BuildStrategyIncrementalSinceGreen BuildStrategy = "incremental_since_green"
+	// BuildStrategyFullMonorepo validates the whole monorepo from scratch.
+	BuildStrategyFullMonorepo BuildStrategy = "full_monorepo"
```
behinddwalls: no need to say monorepo, just say full

stovepipe/entity/request.go
```diff
@@ -48,9 +65,19 @@ type Request struct {
...
 
+	// ****************
+	// Set once at process admit — empty until accepted→processing; never overwritten after
+	// ****************
+
+	// BuildStrategy is the validation scope process chose when admitting this request.
+	BuildStrategy BuildStrategy `json:"build_strategy"`
+	// BaseURI is the base URI to be used for this request when the strategy is incremental.
+	// Empty for full-monorepo builds and cold start.
+	BaseURI string `json:"base_uri"`
```
behinddwalls, line 79: do we need this in request itself? would we publish to build only when the last one is done? i guess question i have is where to we decide latest last green?

stovepipe/entity/request_id.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+// requestCounter extracts the per-queue counter suffix from a request id.
+func requestCounter(id, queue string) (int64, error) {
+	prefix := requestIDPrefix(queue)
+	if !strings.HasPrefix(id, prefix) {
+		return 0, fmt.Errorf("request id %q does not match queue %q", id, queue)
+	}
+	return strconv.ParseInt(id[len(prefix):], 10, 64)
+}
```
behinddwalls, line 60: do we need to understand the format the ID? can we treat it string which is separate from seq?

behinddwalls (changes requested):

stovepipe/entity/request_id.go
mnoah1: I was going back and forth a bit on this (initially I was thinking we'll track sequence and ID as two separate fields).  But as I worked on this branch, realized we could avoid a separate sequence field if we require sequence info to be encoded in the ID and use a comparison function.  Do you have any thoughts on one approach vs. the other?  Obviously there is the cost to parse the int on each comparison, but doing that a handful of times per request will be far from a bottleneck.

stovepipe/entity/request.go
mnoah1: > where to we decide latest last green?

We currently propose doing this just before starting builds ([process steps - 7a](https://github.com/uber/submitqueue/blob/main/doc/rfc/stovepipe/steps/process.md#process-algorithm)).

> do we need this in request itself?

Once selected above, we need to store it somewhere for retrieval once the builds start.  I think `Request` is reasonable for the MVP, but alternatively we could put it in a separate table of planned builds.  The latest green value stored in the Queue entity represents more of a moving bookmark of latest green, while the value here serves as the fixed base for that specific request.

> would we publish to build only when the last one is done? 

Yes - when the concurrent builds is 1, that is effectively true. When increased, it would be whenever an available build slot has become free.

stovepipe/entity/request.go
mnoah1: Updated

behinddwalls (approved):

behinddwalls:

The rebase of `mnoah1/stovepipe-request-entity` completed without conflicts, but the resulting code diff does not match the original. This means the rebase silently altered code content. The branch was **not** force-pushed.

PR #320, opened 2026-07-08
feat(speculation): add prioritizer + prioritization-limit extensions

behinddwalls:
Add the queue-wide prioritizer seam and its limit counterpart from the speculation RFC, as vendor-agnostic extension interfaces under submitqueue/extension/speculation/.

prioritizer: selection is per batch and blind to other batches, so it cannot ration a shared build budget. The prioritizer sees every path across the queue's in-flight batches that runs or wants to run (Selected / Prioritized / Building, each carrying its ID and score), ranks them — scores are probabilities in [0, 1] per the path scorer's contract — and returns sparse decisions naming paths by ID: Promote to admit under the budget, Cancel to preempt, at most one decision per path. Whether it preempts at all is implementation policy (sticky-slots vs preemptive). It never writes: the controller maps each decision back to its tree and applies the status transition under that tree's optimistic lock. Prioritized thus means exactly "admitted under the queue's build budget, cleared to build, not yet building".

prioritizationlimit: the "how much" policy bounding the queue's concurrent speculation builds — the budget the prioritizer admits against. Injected into the prioritizer at construction and applied by it, keeping the interface limit-free; the value is signal-driven, not a fixed constant.

## Stack
1. #337
1. #315
1. #316
1. #317
1. @ #320
1. #331
1. #332
1. #333


submitqueue/extension/speculation/prioritizer/README.md
```diff
@@ -0,0 +1,19 @@
+# Speculation Prioritizer
+
+Vendor-agnostic interface for the queue-wide policy that rations a shared build budget across every in-flight batch in a queue.
+
+See the [Speculation RFC](/doc/rfc/submitqueue/speculation.md) for the end-to-end design and how prioritization fits into the orchestrator pipeline.
+
+## Prioritizer
+
+Selection is per batch and blind to other batches, so it cannot ration a shared budget: if every batch promoted generously, their combined demand could swamp CI. The prioritizer closes that gap. It sees every path across all of the queue's in-flight batches that is running or wants to run, ranks them by each path's score (plus any fairness or tie-break policy), and admits only the subset that fits the queue's concurrent-build budget. Selection expresses *desire* per batch; prioritization reconciles that desire against *supply* — it is the queue-wide enforcer.
```
albertywu, line 9: Is there any contract on what score values can be, for example are scores always in [0, 1] range or something else? Or Prioritizer doesn't care as long as there is some ordering it will use relative ordering to decide on a stack-rank of paths using path score?

submitqueue/extension/speculation/prioritizer/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+The controller hands the prioritizer the queue's candidate paths directly — every path that is `Selected` (wants a slot) or `Prioritized`/`Building` (holds a slot), each carrying its score. It returns **sparse decisions**: `Promote` to admit a pending path, `Cancel` to preempt a running one. Paths it omits are left as-is. It never writes: the controller groups the decisions by `Path.Head` (the batch ID) to locate each affected tree, maps each decision to a status transition (`Promote` → `Prioritized`, `Cancel` → `Cancelling`) applied under that tree's optimistic lock, and enacts it, staying the single writer.
```
albertywu: Does `Prioritized` mean "Waiting to Build" or something? Unclear to me what Prioritized means here

Maybe `Admitted` if it means the path is admitted through the build gates but has not actually started building yet?

submitqueue/extension/speculation/prioritizer/README.md
behinddwalls: Yes...they are always between 0-1, generally we would want to enforce that in the controller when consume it

submitqueue/extension/speculation/prioritizer/README.md
behinddwalls: yes..its same...lack of better words i guess in this case but admitted would be too generic as admitted would also admitted to queue and other stages

submitqueue/extension/speculation/prioritizer/README.md
behinddwalls: Done — the contract is now explicit: scores are probabilities in [0, 1], documented on the pathscorer interface (and its README), and the controller enforces it on consume — out-of-range values are clamped into range with a warning (plus a metric) so a scorer bug can't corrupt ranking inputs. The prioritizer's doc now states it ranks on that scale and needs no other assumption about the values.

submitqueue/extension/speculation/prioritizer/README.md
behinddwalls: Kept `Prioritized`, and the README now defines it crisply: admitted under the queue's build budget and cleared to build, but not yet building — it holds a slot while waiting for the build stage to dispatch it and a build signal to confirm (then `Building`). Also added why it isn't "Admitted": admission happens at several points in the queue's lifecycle, while Prioritized names exactly the grant this seam makes, in contrast to `Selected` which is only the batch-level desire.

albertywu (approved):

PR #323, opened 2026-07-09
feat(stovepipe): advance queue latest_request_id on ingest

mnoah1:
## What?
- After ingest inserts a new request row, advance that queue's latest request ID

## Why?
- Implementing logic outlined in https://github.com/uber/submitqueue/blob/mnoah1/process-docs-update/doc/rfc/stovepipe/steps/process.md - the process task will use this as a pointer to determine which requests should get a build, and which ones have already been superseded

## Test Plan
- Unit tests wire QueueStore expectations; integration test asserts latest_request_id on ingest

PR #324, opened 2026-07-09
feat(stovepipe): coalesce backlog in process controller

mnoah1:
## What

This PR begins implementing core logic of the Process controller to supersede older requests as new ones come in.

https://github.com/uber/submitqueue/blob/main/doc/rfc/stovepipe/steps/process.md#process-algorithm

- `accepted` requests get compared against the queue's latest request ID value
  - If a request is outdated compared to the queue's last ID, mark it superceded (this PR)
  - Otherwise, we'll build it now or re-check it later when a build slot is free (logic to come)
- `superseded` / `processing` → ack no-op
- Retryable load for request/queue not-found

## Why
Implementing https://github.com/uber/submitqueue/blob/main/doc/rfc/stovepipe/steps/process.md step.  Building out logic to accept incoming requests and narrow down builds to only the latest head.

## Test plan

- [x] `//stovepipe/controller/process:go_default_test`
- [x] `//stovepipe/controller:go_default_test`
- [x] `//test/integration/stovepipe:go_default_test`

### DIFF
--- a/stovepipe/controller/process/process.go
+++ b/stovepipe/controller/process/process.go
@@ -12,11 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package process holds the process-stage queue controller. It consumes the
-// request ids ingest publishes, reloads the Request from storage, and (in a
-// future change) decides the build strategy by asking SourceControl how the new
-// head relates to the queue's last-green URI. For now it is a thin consumer that
-// reloads and logs the request, establishing the stage and its wiring.
+// Package process holds the process-stage queue controller. It consumes request
+// ids from ingest, reloads the Request from storage, coalesces older heads, and
+// (in later changes) gates concurrency, decides build strategy, and admits
+// winners to build.
 package process
 
 import (
@@ -29,12 +28,13 @@ import (
 	"github.com/uber/submitqueue/platform/errs"
 	"github.com/uber/submitqueue/platform/metrics"
 	stovepipemq "github.com/uber/submitqueue/stovepipe/core/messagequeue"
+	"github.com/uber/submitqueue/stovepipe/entity"
 	"github.com/uber/submitqueue/stovepipe/extension/storage"
 	"go.uber.org/zap"
 )
 
 // Controller consumes ProcessRequest messages from the process stage, reloads the
-// referenced Request from storage, and logs it. Implements consumer.Controller.
+// referenced Request from storage, and coalesces older heads. Implements consumer.Controller.
 type Controller struct {
 	logger        *zap.SugaredLogger
 	metricsScope  tally.Scope
@@ -63,10 +63,8 @@ func NewController(
 	}
 }
 
-// Process reloads the request referenced by the delivery and logs it. Returns nil
-// to ack (success) or an error to nack (retry). A not-yet-visible request is
-// retryable: ingest persists and publishes, but a stale read may not see the row
-// yet, so redelivery converges.
+// Process reloads the request referenced by the delivery and coalesces older heads.
+// Returns nil to ack (success) or an error to nack (retry).
 func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) (retErr error) {
 	const opName = "process"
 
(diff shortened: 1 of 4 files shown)


behinddwalls (approved):

stovepipe/controller/process/process.go
```diff
@@ -82,29 +80,119 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) (r
...
 	if err != nil {
 		metrics.NamedCounter(c.metricsScope, opName, "storage_errors", 1)
-		if errors.Is(err, storage.ErrNotFound) {
-			// Retryable: the request row may not be visible yet; redelivery converges.
-			return errs.NewRetryableError(fmt.Errorf("request %s not found yet: %w", pr.Id, err))
+		return err
+	}
+
+	switch request.State {
+	case entity.RequestStateSuperseded, entity.RequestStateProcessing:
```
behinddwalls: The RFC's step 3 says `processing` → re-publish to `build` (the prior publish may have failed). Acking as a no-op is right for now since the build stage doesn't exist yet, but only a test name records that intent. Could you add a `// TODO: re-publish to build once the build stage lands (RFC process algorithm, step 3)` on the `RequestStateProcessing` case so it isn't forgotten?

stovepipe/controller/process/process.go
```diff
@@ -82,29 +80,119 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) (r
...
+		return nil
+	case entity.RequestStateAccepted:
+		return c.processAccepted(ctx, request)
+	default:
+		c.logger.Infow("ignored request in unexpected state",
```
behinddwalls: An unknown state reaching `process` means a state machine bug or a bad row, not routine traffic — consider `Warnw` here so it surfaces. Also worth a table case in the test to pin the ack behavior.

stovepipe/controller/process/process.go
```diff
@@ -82,29 +80,119 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) (r
...
+			"request_id", request.ID,
+			"queue", request.Queue,
+			"state", string(request.State),
+		)
+		return nil
+	}
+}
+
+// processAccepted coalesces older heads against queue.latest_request_id. The latest
+// head is left in accepted until admit and the concurrency gate land in later PRs.
```
behinddwalls, line 109: Acking the latest head leaves it `accepted` with no pending delivery — with only this PR, nothing ever moves it forward. The implicit recovery is that the poller re-reports the head and ingest re-publishes while `State == Accepted`. That invariant is load-bearing; can we say so in this comment? e.g. "Acking here is safe because ingest re-publishes accepted heads on every re-report, so the head is re-delivered until admit logic lands."

stovepipe/controller/process/process.go
```diff
@@ -82,29 +80,119 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) (r
...
+func (c *Controller) processAccepted(ctx context.Context, request entity.Request) error {
+	queueRow, err := c.loadQueue(ctx, request.Queue)
+	if err != nil {
+		return err
+	}
+
+	if queueRow.LatestRequestID == "" {
```
behinddwalls, line 119: Ingest stamps `latest_request_id` before publishing and fails the call otherwise, so an empty pointer at delivery time can only be a stale read (or a pre-#323 row). Acking is the safe choice, but logging it as `"latest head awaiting admit"` hides the anomaly — a distinct message (and probably `Warnw`) would make a stuck replica visible. Optionally dedupe with the identical log below by branching only on the comparison.

stovepipe/controller/process/process.go
```diff
@@ -82,29 +80,119 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) (r
...
+			"state", string(request.State),
+		)
+		return nil
+	}
+}
+
+// processAccepted coalesces older heads against queue.latest_request_id. The latest
+// head is left in accepted until admit and the concurrency gate land in later PRs.
+func (c *Controller) processAccepted(ctx context.Context, request entity.Request) error {
+	queueRow, err := c.loadQueue(ctx, request.Queue)
```
behinddwalls, line 111: `storage_errors` now only counts the request load; queue-load and supersede CAS failures skip it. Suggest counting those too — and adding a `superseded` counter in the `cmp < 0` branch, since coalescing rate is the number we'll want on a dashboard for this feature.

stovepipe/controller/process/process_test.go
```diff
@@ -61,39 +76,154 @@ func processPayload(t *testing.T, id string) []byte {
...
-	err := c.Process(context.Background(), delivery(t, ctrl, processPayload(t, testID)))
-	require.Error(t, err)
-	assert.False(t, errs.IsRetryable(err))
-}
+			id := testID
+			if tt.id != "" {
+				id = tt.id
+			}
+			payload := processPayload(t, id)
+			if tt.name == "malformed payload is not retryable" {
```
behinddwalls: Keying the malformed payload off `tt.name ==` will silently stop working if the case is renamed. A `malformed bool` (or `payload []byte`) field in the table is sturdier.

stovepipe/controller/process/process_test.go
```diff
@@ -61,39 +76,154 @@ func processPayload(t *testing.T, id string) []byte {
...
+					Name:    testQueue,
+					Version: 1,
+				}, nil)
+			},
+		},
+		{
+			name: "older accepted head is superseded",
+			id:   "request/monorepo/main/3",
+			setup: func(m processMocks) {
+				olderID := "request/monorepo/main/3"
```
behinddwalls: `olderID` is duplicated in both supersede cases and also mirrored in `tt.id` — hoist to a const next to `testID`. While here: a case where `CompareRequestID` fails (e.g. `LatestRequestID` from a different queue) asserting non-retryable would pin the DLQ behavior for malformed ids.

behinddwalls (commented):
Reviewed the coalesce logic against the ingest-side invariants (pointer stamped before publish, queue row ensured before publish) — the supersede path and retryability calls all check out. Left a few minor inline suggestions, nothing blocking.

One non-inline note: unit coverage for the new transition is good; `test/integration/stovepipe` still only exercises ingest. A follow-up that ingests two successive heads through the real stack and asserts the older ends `superseded` would cover the CAS path against real MySQL — fine as a separate PR once the fake SourceControl can report successive heads.

stovepipe/controller/process/process.go
behinddwalls: Resolved by the stack: #335 admits the latest head (`admitRequestToBuild`), so the acked-into-limbo state no longer exists — the only remaining ack-and-wait path is gate-closed, which carries the re-enqueue TODO.

stovepipe/controller/process/process.go
behinddwalls: Resolved by the stack: #335 renames this log to "latest head awaiting queue.latest_request_id stamp from ingest", which makes the stale-read case distinguishable.

PR #325, opened 2026-07-09
docs: History API RFC

albertywu:
## Summary
Adds an RFC for the History API in Gateway


doc/rfc/submitqueue/history-api.md
```diff
@@ -0,0 +1,141 @@
...
+    // Free-form display or debugging context associated with the event.
+    map<string, string> metadata = 4;
+}
+
+message RequestHistory {
+    // Globally unique identifier for the request associated with these events.
+    string sqid = 1;
+    // Retained request-log events ordered by timestamp_ms ascending.
+    repeated HistoryEvent events = 2;
+}
```
behinddwalls: do we need this container or can we have sqid in HistoryResponse itself? also with HistoryRequest (API) and RequestHistory, having both seems a bit confusing

doc/rfc/submitqueue/history-api.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Status Contract
+
+`HistoryEvent.status` is a string, not a protobuf enum. Its value is populated from `entity.RequestStatus`, the same customer-facing status type stored in `RequestLog` and returned by `Status`.
+
+Keeping the wire field as a string allows SubmitQueue to add request statuses without requiring clients to adopt a new generated enum before they can read the response. Clients must tolerate status strings they do not recognize.
+
+## Event Projection
+
+Each stored `RequestLog` row produces exactly one `HistoryEvent`:
```
behinddwalls: do we need this at all? can we not use the RequestLog table to get this data needed for this API directly

doc/rfc/submitqueue/history-api.md
```diff
@@ -0,0 +1,141 @@
...
+
+These repeated entries are part of the stored audit trail and must remain visible. Hiding them would require an arbitrary definition of event equivalence and could conceal useful retry behavior during debugging.
+
+## Pagination
+
+The initial API returns all currently retained histories and events matching the selected identifier in one response and does not support pagination.
+
+A SubmitQueue request has a bounded lifecycle under normal operation, and a change is expected to have a modest number of retained submissions, so the additional cursor and query complexity is not justified initially. Pagination can be added later if production histories demonstrate that response size needs an explicit bound.
+
+## Consistency
```
behinddwalls, line 109: i think it will depend if we can use requestLog as is or not

doc/rfc/submitqueue/history-api.md
albertywu: container is needed here to support requests with change_id

Since a change_id maps to potentially multiple sqids, we embed the sqid in the response so the caller gets the history for multiple sqids and can disambiguate which request history corresponds to which sqid.

We could have sqid in the HistoryResponse itself, but it would duplicate the information. I could go either way, lmk if you feel strongly about flattening things a bit and including sqid directly inside HistoryEvent and returning `repeated HistoryEvent` without the extra `RequestHistory` layer

doc/rfc/submitqueue/history-api.md
```diff
@@ -0,0 +1,141 @@
...
+
+`RequestID` is excluded from each event because it is represented once by `RequestHistory.sqid`.
+
+## Ordering
+
+Events are returned in `timestamp_ms` ascending order so the response reads as a lifecycle from oldest to newest. This matches the ordering contract of `RequestLogStore.List`.
+
+Request-log timestamps are generated by callers, not by the storage backend. The API therefore defines chronological order by the recorded `timestamp_ms`, not database insertion order.
+
+Multiple events may have the same timestamp. Their relative order is unspecified to API clients because the storage tie-breaker is an internal implementation detail with no lifecycle meaning.
```
sbalabanov: it is better be deterministic though as a part of the contract. For that reason, mysql implementation has salt
https://github.com/uber/submitqueue/blob/main/submitqueue/extension/storage/mysql/schema/request_log.sql#L4

doc/rfc/submitqueue/history-api.md
albertywu: Ah the wording here is unclear, clarified that we will not be projecting onto another table and just using RequestLog directly

doc/rfc/submitqueue/history-api.md
albertywu: The plan was to use directly RequestLog as implementation

doc/rfc/submitqueue/history-api.md
```diff
@@ -0,0 +1,141 @@
...
+
+When a change URI matches multiple requests, `histories` is ordered by each request's first retained event timestamp ascending, with `sqid` ascending as a stable tie-breaker.
```
sbalabanov: if we allow same change to be in multiple requests (we probably do - the user can resubmit the same change), then the whole response API likely needs to be different and history records organized by request ids

sbalabanov: may be we need two RPC functions
1) HistoryBySQID -> list[Event]
2) HistoryByChangeID -> map[sqid]list[Event]

sbalabanov (changes requested):

doc/rfc/submitqueue/history-api.md
albertywu: updated RFC to use separate APIs HistoryByChangeID and HistoryBySQID with corresponding shapes as discussed over slack

doc/rfc/submitqueue/history-api.md
albertywu: updated language here

doc/rfc/submitqueue/history-api.md
albertywu: updated shapes as discussed over slack

doc/rfc/submitqueue/history-api.md
```diff
@@ -0,0 +1,160 @@
...
+Users need to inspect how a request progressed through SubmitQueue, not only its current reconciled state. They may start with the `sqid` returned by `Land` or with a provider-specific change ID represented by a change URI supplied to `Land`. The existing `Status` API collapses the append-only request log into one current status, which is appropriate for polling but hides the sequence of events needed for debugging and lifecycle displays.
+
+The gateway owns the request log and is the only service that reads it. The history APIs preserve that ownership boundary by serving gateway-owned `RequestLog` records for the request or requests selected by the caller.
+
+## API Shape
+
+The gateway exposes two read-only RPCs because an `sqid` selects one event list while a change ID may select multiple requests:
+
+```proto
+message HistoryBySQIDRequest {
```
behinddwalls, line 18: we could probably drop SQ? `HistoryByIDRequest`?

doc/rfc/submitqueue/history-api.md
```diff
@@ -0,0 +1,160 @@
...
+    int64 timestamp_ms = 1;
+    // Customer-friendly request status recorded by the event.
+    string status = 2;
+    // Error message associated with the event. Empty if none.
+    string last_error = 3;
+    // Free-form display or debugging context associated with the event.
+    map<string, string> metadata = 4;
+}
+
+message HistoryBySQIDResponse {
```
behinddwalls, line 34: same, we could drop SQ

doc/rfc/submitqueue/history-api.md
```diff
@@ -0,0 +1,160 @@
...
+}
+
+message HistoryEventList {
+    // Retained request-log events ordered by timestamp_ms ascending with a stable tie-breaker.
+    repeated HistoryEvent events = 1;
+}
+
+message HistoryByChangeIDResponse {
+    // Retained event lists keyed by the sqid of each request that included the change.
+    map<string, HistoryEventList> histories_by_sqid = 1;
```
behinddwalls: do we need this to be ordered? like in desc order of request timestamps? we may want to change this to be a list of containers to support that. map would be just random

sbalabanov: I think list is better than map
each list item is an object itself, sqid + list[event]
sorted by sqid

doc/rfc/submitqueue/history-api.md
sbalabanov: I think SQID is the settled term already

doc/rfc/submitqueue/history-api.md
```diff
@@ -0,0 +1,160 @@
...
+    string status = 2;
+    // Error message associated with the event. Empty if none.
+    string last_error = 3;
+    // Free-form display or debugging context associated with the event.
+    map<string, string> metadata = 4;
+}
+
+message HistoryBySQIDResponse {
+    // Retained request-log events ordered by timestamp_ms ascending with a stable tie-breaker.
+    repeated HistoryEvent events = 1;
```
sbalabanov, line 36: could also be HistoryEventList for consistency

doc/rfc/submitqueue/history-api.md
behinddwalls: fair :)

doc/rfc/submitqueue/history-api.md
albertywu: See updated shape, lmk if any concerns

doc/rfc/submitqueue/history-api.md
albertywu: updated shape as discussed

behinddwalls (approved):

sbalabanov (approved):

PR #328, opened 2026-07-09
docs(rfc): add change URI design

behinddwalls:
### Why?

Change URIs are the system-wide identity of a code change — stored as a primary-key column, correlated by strict string equality, and shared across SubmitQueue, Stovepipe, and Runway via the `api/base/change` contract. Today only `git://` URIs name the provider instance; GitHub and Phabricator URIs carry no host, so the same URI can name two different changes on two provider instances, and the host lives out-of-band in queue config — coupling change identity to queue topology.

### What?

Adds `doc/rfc/change-uri.md`, a spec for change URIs: every URI is `scheme://{host[:port]}/{path}` where the scheme names the provider model, the mandatory authority names the instance, and the path pins the change to an exact code state. Defines the per-provider formats (GitHub PR, Phabricator Diff, git ref/commit), canonical-form rules (validated, never normalized: lowercase host, verbatim optional port, full 40-hex SHAs, byte-for-byte round-trip), and rejected alternatives (out-of-band host, per-flavor `ghe`/`ghes` schemes, web URLs as identity).

Also links the RFC from `doc/rfc/index.md` under Shared, and drops a pre-existing duplicated "Build Runner" entry from the index.

## Stack
1. @ #328
1. #329


doc/rfc/change-uri.md
```diff
@@ -0,0 +1,42 @@
...
+- **authority** — the provider *instance*: the `host[:port]` the change lives on. Mandatory.
+- **path** — the change within that instance, pinned to an exact code state (head SHA or diff ID), so staleness is detectable by comparing the pin against the provider's current state.
+
+## Formats
+
+| Provider | Format | Example |
+|---|---|---|
+| GitHub PR | `github://{host[:port]}/{org}/{repo}/pull/{pr}/{head_sha}` | `github://github.uberinternal.com/uber/submitqueue/pull/123/c3a4…89ab` |
+| Phabricator Diff | `phab://{host[:port]}/D{revision}/{diff}` | `phab://phabricator.example.com/D12345/67890` |
+| git ref/commit | `git://{host[:port]}/{repo}/{ref}/{sha}` | `git://git.example.com:9418/uber/mono/refs%2Fheads%2Fmain/c3a4…89ab` |
```
JamyDev, line 19: Thought we had decided we don't need host in git://? though I do prefer it..

doc/rfc/change-uri.md
```diff
@@ -0,0 +1,42 @@
...
+- **Port** — optional, digits only, verbatim when present. Custom schemes have no registered default port, so there is nothing to strip; omit it unless the backend listens on a non-standard one.
+- **Commit SHAs** — the full 40-character lowercase hex form. Abbreviated or uppercase SHAs are rejected, not expanded or folded.
+- **Round-trip** — parsing a valid URI and re-serializing the parsed form yields the input byte-for-byte.
+
+Parsing is delegated to `net/url`, which handles `host:port` splitting, bracketed IPv6 hosts, and percent-encoding correctly.
+
+## Rejected alternatives
+
+- **Host out-of-band in queue config.** Conflates identity with routing: the meaning of a stored primary-key value must not depend on deployment wiring, and the shared contract must be interpretable by every domain that imports it.
+- **Per-flavor schemes** (`ghe://`, `ghes://`). Redundant with the authority, and an open-ended enum baked into parsers and routing — a new instance should be configuration, not code.
```
JamyDev: ghe/ghex doesn't have a different API? If so then I guess single github authority is sufficient.

JamyDev (approved):

doc/rfc/change-uri.md
behinddwalls: git:// has carried the host since that parser landed — `Remote` is required and an empty host is rejected — so for git the RFC documents the status quo rather than adding anything. Keeping it is also what makes the shape uniform across providers (scheme = model, authority = instance, path = change): without it, `git://uber/monorepo/…` cannot say which remote `uber` lives on, and the same repo path on two remotes would collide byte-for-byte. Left mandatory.

doc/rfc/change-uri.md
behinddwalls: Right — github.com / GHE / GHES share one PR model and one API surface. What differs per instance is the API base path (`api.github.com` vs `https://<host>/api/v3` / `/api/graphql`) plus some version skew, and that is already per-instance wiring config on the HTTP client (`BaseURLTransport`), not something the identifier needs to encode. So a single `github` scheme + authority is sufficient. Added a sentence to the per-flavor-schemes bullet capturing this.

PR #329, opened 2026-07-09
feat(change): require host[:port] authority in change URIs

behinddwalls:
### Why?

The Change URIs RFC (doc/rfc/change-uri.md, previous PR in this stack) makes every change URI carry the provider instance's `host[:port]` as its RFC 3986 authority, so a URI is globally unambiguous on its own: GitHub and Phabricator URIs previously carried no host (the same URI could name two different changes on two provider instances), and the `ghe`/`ghes` schemes stood in for an instance discriminator without actually being one.

### What?

- `platform/base/change/github`: parser rewritten on `net/url`; format is now `github://{host[:port]}/{org}/{repo}/pull/{pr}/{sha}`; single `github` scheme — `ghe`/`ghes` are rejected; parsed form gains a `Host` field.
- `platform/base/change/phabricator`: same treatment; format is now `phab://{host[:port]}/D{revision}/{diff}`; the host moves in-band from queue config into the URI authority.
- `platform/base/change/git`: unchanged format (it already carried the authority); gains only the new lowercase-host validation.
- `platform/base/change/changeutil`: new `IsLowercaseASCII` predicate — canonical form is validated, never normalized (uppercase hosts are rejected, not folded), matching the existing full-40-hex SHA rule.
- Contract docs updated in `api/base/change` proto (comment-only, regenerated via `make proto`), `platform/base/change`, and entity comments; all test fixtures across gateway, orchestrator, extensions, runway, integration, and e2e suites moved to authority-bearing URIs (`github.example.com` / `phab.example.com`).

No wire change: `Change.uris` stays a repeated string; only the accepted grammar of its contents changes. Pre-production, so this is a hard cutover with no dual-format window (per the RFC's rejected alternatives).

## Test Plan

✅ `make test` (65/65), ✅ `make build`, ✅ `make proto` / `make fmt` / `make gazelle` idempotent. Parser suites cover the new grammar: host and host:port round-trips, nested orgs, and rejections for missing host, uppercase host, host-less legacy URIs, and `ghe://`/`ghes://` schemes.

## Stack
1. #328
1. @ #329


platform/base/change/git/change_id.go
```diff
@@ -72,6 +72,9 @@ func ParseChangeID(raw string) (ChangeID, error) {
 	if u.Host == "" {
 		return ChangeID{}, fmt.Errorf("invalid change ID %q: missing remote (expected format: %s)", raw, changeIDFormat)
 	}
+	if !changeutil.IsLowercaseASCII(u.Hostname()) {
```
JamyDev, line 75: This just looks at the host? I could have an uppercase monorepo name and potentially cause uncaught issues. Why don't we normalize the entire URL to lowercase and get rid of this error mode?

behinddwalls: i asked claude earlier about it and it seems rest of the URL can case sensitive...so it decided not to account for them

behinddwalls: Lowercasing the whole URI would corrupt identities rather than canonicalize them, because the components have different case semantics:

- **git refs and repo paths on a remote are case-sensitive** — `refs/heads/Foo` and `refs/heads/foo` are different refs, so folding would silently repoint the URI at a different resource.
- **GitHub org/repo is a middle case**: lookups are case-insensitive, but each repo has one canonical casing and uppercase is legal (`uber/RIBs`). The parser cannot know the canonical form, so it must take these verbatim — folding would mangle legitimate names.
- **The host is the one component with a universally known canonical form** (DNS is case-insensitive, lowercase is canonical), so it is the one place we can validate without ever being wrong. Same for SHAs (already enforced full lowercase hex).

We validate rather than normalize because URIs are compared byte-for-byte as primary keys and claim-lookup keys: a normalizer has to run identically at every entry point forever or the split identity comes back silently, and it breaks the parse→`String()` byte-for-byte round-trip.

You are right that verbatim org/repo leaves a residual aliasing window on GitHub specifically: the same repo submitted as `Uber/Repo` and `uber/repo` would be two identities (both still operate on the correct PR, since the API is case-insensitive). The right place to close that is where the provider is consulted — the change provider can compare the URI's org/repo against the canonical `nameWithOwner` GitHub returns and reject mismatches. Can add that as a follow-up PR on this stack if we want it. The RFC (#328) now has an "all other path segments — verbatim" bullet capturing the per-component case rules.

JamyDev: Good enough justification for me

JamyDev (approved):

PR #331, opened 2026-07-10
feat(storage,entity): path identity + speculation path→build mapping store

behinddwalls:
### Why?

Integrating speculation needs the controllers to answer "which build belongs to this path" in both directions, without ever looking a row up by non-key attribute — the storage contract stays get/put-by-key so any KV backend can satisfy it. The build system mints its own build identifiers, and those stay the build store's primary key: the runner's ID is the natural name for the build row. What's missing is a durable link between a tree path and its build that exists independent of any in-flight message, hung on the path identity (SpeculationPathInfo.ID) introduced at the bottom of this stack.

### What?

`entity.SpeculationPathBuild` is the new mapping entity ({PathID, BuildID, BatchID, CreatedAt, Version}, named after its `speculation_path_build` table) with a `SpeculationPathBuildStore` (Create/Get keyed by PathID): the forward path→build lookup, written only by the build controller, at most one build per path with ErrAlreadyExists making the existing row the truth on races. PathID is globally unique — the path ID contract on SpeculationPathInfo.ID is strengthened accordingly, since this table keys rows by it alone. BatchID makes the row self-describing without parsing PathID's format; Version follows the repo's optimistic-locking convention (write-once today, reserved for future conditional re-pointing flows). `entity.Build` keeps the runner-minted `ID` as its primary key and gains `SpeculationPathID` as a plain column — the reverse build→path lookup; the previously embedded `SpeculationPath` value is dropped as a redundant denormalized copy of what the tree already stores. `SpeculationPath.Equal` (order-sensitive Base + Head) provides structural path identity for the few controller spots where only structure can identify a path (deduplicating enumerator output, carrying entries over across re-enumeration).

MySQL gains the `speculation_path_build` table and the build table swaps `speculation_path`/`runner_id` for `speculation_path_id`; schema files are picked up automatically (the schema dir is globbed by both Bazel and the testutil ApplySchema helper).

## Test Plan
✅ `bazel test //submitqueue/entity/... //submitqueue/extension/storage/... //submitqueue/orchestrator/...` and the storage integration suite exercises Create/Get round-trips and the duplicate-create race.

## Stack
1. @ #331
1. #332
1. #333


behinddwalls:

The rebase of `preetam/int/storage-build-lookup` completed without conflicts, but the resulting code diff does not match the original. This means the rebase silently altered code content. The branch was **not** force-pushed.

behinddwalls:

The rebase of `preetam/int/storage-build-lookup` completed without conflicts, but the resulting code diff does not match the original. This means the rebase silently altered code content. The branch was **not** force-pushed.

behinddwalls:

The rebase of `preetam/int/storage-build-lookup` completed without conflicts, but the resulting code diff does not match the original. This means the rebase silently altered code content. The branch was **not** force-pushed.

behinddwalls:

The rebase of `preetam/int/storage-build-lookup` completed without conflicts, but the resulting code diff does not match the original. This means the rebase silently altered code content. The branch was **not** force-pushed.

behinddwalls:

The rebase of `preetam/int/storage-build-lookup` completed without conflicts, but the resulting code diff does not match the original. This means the rebase silently altered code content. The branch was **not** force-pushed.

submitqueue/entity/queue.go
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) 2025 Uber Technologies, Inc.
```
sbalabanov: why is this file in the diff? does not seem to be used for the purpose

submitqueue/entity/speculation_path_build.go
```diff
@@ -0,0 +1,43 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package entity
+
+// SpeculationPathBuild is the path->build mapping: given a speculation path's
+// ID, it records the build that path resolved to. It is written by the build
+// controller when it triggers a build for a speculation path; this is the
+// forward lookup (path->build), while the reverse lookup (build->path) is
+// Build.SpeculationPathID. Today a row is write-once — only Create and Get
```
sbalabanov: leave logic implementation details to the function, not an entity object

submitqueue/entity/speculation_path_build.go
```diff
@@ -0,0 +1,43 @@
...
+// exist, no Update — so Version is always 1 on every row; the field is
+// reserved for a future repair/re-trigger flow that conditionally re-points a
+// path to a newer build without a schema migration (not used yet).
+type SpeculationPathBuild struct {
+	// PathID is the speculation path's ID (SpeculationPathInfo.ID). It is the
+	// primary key of this mapping.
```
sbalabanov: indicate if it is globally unique, or within queue etc

submitqueue/entity/speculation_path_build.go
```diff
@@ -0,0 +1,43 @@
...
+	BuildID string
+	// BatchID is the batch whose speculation tree contains this path. It
+	// makes the row self-describing without parsing PathID's internal format.
+	BatchID string
+	// Version is the version of the object, used for optimistic locking:
+	// updates are conditional on the persisted version matching the caller's
+	// expected version. Versioning starts at 1; version arithmetic is owned
+	// by the controller, the store performs a pure conditional write. Not
+	// used yet — reserved for a future conditional-update flow.
+	Version int32
```
sbalabanov: I think we put a version last in the struct... along with other mutable fields

submitqueue/entity/speculation_tree.go
```diff
@@ -29,6 +29,27 @@ type SpeculationPath struct {
...
+// carrying entries over across re-enumeration); everything else references a
+// persisted path by its assigned ID (SpeculationPathInfo.ID).
+func (p SpeculationPath) Equal(other SpeculationPath) bool {
+	if p.Head != other.Head {
+		return false
+	}
+	if len(p.Base) != len(other.Base) {
+		return false
+	}
+	for i := range p.Base {
```
sbalabanov: slices.Equal() ?

sbalabanov (changes requested):

submitqueue/entity/queue.go
behinddwalls: You're right — nothing in this PR uses `QueueID`. Removed it here; it now rides with the prioritize-stage change that introduces its first consumers.

submitqueue/entity/speculation_path_build.go
behinddwalls: Done — trimmed the type doc to what the mapping *is* (the two lookup directions); the write/read semantics (written by the build controller, write-once, ErrAlreadyExists on races) live on `SpeculationPathBuildStore`'s doc in the storage package.

submitqueue/entity/speculation_path_build.go
behinddwalls: Done — documented `PathID` as globally unique, and strengthened the `SpeculationPathInfo.ID` contract accordingly (it was documented as unique-within-tree, but this table keys rows by it alone, so the ID contract must be global).

submitqueue/entity/speculation_path_build.go
behinddwalls: Done — `Version` moved to the end of the struct, matching `Batch` / `SpeculationTree`.

submitqueue/entity/speculation_tree.go
behinddwalls: Done — `slices.Equal`.

submitqueue/entity/build.go
```diff
@@ -56,16 +56,18 @@ func (s BuildStatus) IsTerminal() bool {
 // Build represents a build scheduled for a batch along a specific speculation path.
 // All fields except the Status are immutable after creation.
 type Build struct {
-	// ID represents the build ID. It is the responsibility of a build management system to ensure
-	// that this is unique.
+	// ID is the identifier minted by the queue's build runner when the build
```
albertywu, line 59: > ID is the identifier minted by the queue's build runner

^ does our system enforce single build runner? If multiple build runners, then IDs could potentially collide if created by build runners that don't coordinate with each other?

albertywu (approved):

submitqueue/entity/build.go
behinddwalls: expectation is on buildrunner to mint unique IDs however they want for the whole system i guess...

PR #333, opened 2026-07-10
feat(speculation): add probability path scorer

behinddwalls:
### Why?

The pathscorer seam has no real implementation — only mocks and a programmable fake — yet its output is the ranking currency the selector and prioritizer act on. Speculation needs a default scorer whose scores mean something from the start: inert while trees hold a single path, but ranking chain-vs-fallback paths correctly the moment an enumerator produces alternatives. It is split out from the parity-stub impls because it is the one seam implementation with an actual model, worth reviewing on its own.

### What?

`pathscorer/probability` — scores each path as the probability that exactly that path's assumption holds: `Score = p(head) × Π p(d) for base deps × Π (1−p(d)) for non-base head deps`, returned as path-ID-keyed `PathScore`s per the seam contract (probabilities in [0, 1] by construction).

Each batch's probability resolves from its state, so resolved outcomes override predictions: `succeeded` → 1, `failed`/`cancelled` → 0, otherwise the score stage's predicted `Batch.Score` (the score stage sets it before a batch reaches speculation; best-effort `cancelling` keeps its prediction). A dead dependency therefore zeroes every path built on it and boosts every path that excluded it by the full `(1−p)=1` factor — score mass shifts to the paths consistent with reality, with no cross-path coupling. Folding outcomes into path *status* stays the controller's reconcile job; the scorer only recomputes scores, over every path regardless of status.

Dependencies are read one at a time through the injected `storage.BatchStore` per the store's key/value contract, each batch loaded at most once per call; store errors return wrapped and unclassified. The README walks a three-path worked example (chain / drop-B / alone) before and after a dependency failure.

## Test Plan
✅ `make gazelle && make fmt && bazel test //submitqueue/extension/speculation/...`. Unit coverage: probability-formula table including all state-resolution cases (succeeded/failed/cancelled in and out of base, cancelling keeps prediction), a three-path dependency-failure scenario pinning the score-mass shift, per-dep single store read, error propagation, empty-tree short-circuit.

## Stack
1. #331
1. #332
1. @ #333

### DIFF
--- a/submitqueue/extension/speculation/pathscorer/probability/README.md
+++ b/submitqueue/extension/speculation/pathscorer/probability/README.md
@@ -1,0 +1,11 @@
+# Probability Path Scorer
+
+The probability `pathscorer.Scorer` scores every path in a batch's speculation tree as the probability that exactly that path's assumption holds: its base dependencies all land, and every other active dependency of the head fails to land. Concretely, a path's score is the head batch's probability, multiplied by the probability of each base dependency, multiplied by one minus the probability of each of the head's dependencies not in the base.
+
+Each batch's probability is resolved from its state, so a resolved outcome overrides the prediction. A batch that has succeeded contributes certainty 1; a batch that has failed or been cancelled contributes 0; a batch still in flight contributes `entity.Batch.Score`, the per-batch success probability the score stage sets before a batch ever reaches speculation. A batch in the best-effort `cancelling` state keeps its prediction — it may still succeed. This state resolution is what redistributes score when the world changes: a dead dependency zeroes every path that built on it and boosts every path that excluded it by the full `(1 − p) = 1` factor, with no cross-path coupling needed. A worked example of this redistribution lives in the package documentation.
+
+## Scope
+
+The model is deliberately simple: it treats dependency outcomes as independent, and does not weigh how long a batch has waited, its historical pass rate, or any correlation between siblings. It scores every path in the tree regardless of status, returning one path-ID-keyed score per path — the controller merges them into the tree and overwrites `Score` wholesale on every respeculate; nothing else about a path passes through the scorer. Folding resolved outcomes into path *status* (dead-pathing, cancellation) is the controller's reconcile job, not the scorer's; the scorer only makes the scores reflect them.
+
+Batches are read one at a time through the injected `storage.BatchStore`, matching the store's key/value contract, and each batch referenced by the tree is loaded at most once per `Score` call. A store error, including a missing batch, is returned wrapped and unclassified — scorer implementations, like all extensions, leave user/infra classification to the controller's error classifier.
--- a/submitqueue/extension/speculation/pathscorer/probability/probability.go
+++ b/submitqueue/extension/speculation/pathscorer/probability/probability.go
@@ -1,0 +1,158 @@
+// Copyright (c) 2025 Uber Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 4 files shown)


submitqueue/extension/speculation/pathscorer/probability/probability.go
```diff
@@ -0,0 +1,165 @@
...
+	"fmt"
+
+	"github.com/uber/submitqueue/submitqueue/entity"
+	"github.com/uber/submitqueue/submitqueue/extension/speculation/pathscorer"
+	"github.com/uber/submitqueue/submitqueue/extension/storage"
+)
+
+// neutralScore is the probability assigned to a non-terminal batch that has
+// not yet been scored (entity.Batch.Score's zero value), so an unscored
+// dependency neither helps nor hurts a path's score.
```
sbalabanov: how can it even happen?

submitqueue/extension/speculation/pathscorer/probability/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+## Scope
+
+The model is deliberately simple: it treats dependency outcomes as independent, and does not weigh how long a batch has waited, its historical pass rate, or any correlation between siblings. It scores every path in the tree regardless of status, returning one path-ID-keyed score per path — the controller merges them into the tree and overwrites `Score` wholesale on every respeculate; nothing else about a path passes through the scorer. Folding resolved outcomes into path *status* (dead-pathing, cancellation) is the controller's reconcile job, not the scorer's; the scorer only makes the scores reflect them.
+
+Batches are read one at a time through the injected `storage.BatchStore`, matching the store's key/value contract, and each batch referenced by the tree is loaded at most once per `Score` call. A store error, including a missing batch, is returned wrapped and unclassified — scorer implementations, like all extensions, leave user/infra classification to the controller's error classifier.
```
sbalabanov, line 11: the scorer will work on multiple speculation paths as it needs to respeculate everything, and it will load same batches from the database
this does not look efficient

sbalabanov (approved):

submitqueue/extension/speculation/pathscorer/probability/probability.go
behinddwalls: It can't — by the time a batch reaches speculation the score stage has already set `Batch.Score`, so the unscored case is unreachable. Removed the fallback and the `neutralScore` constant; `batchProbability` now returns `b.Score` directly for non-terminal batches.

submitqueue/extension/speculation/pathscorer/probability/README.md
behinddwalls: Fair point, and agreed there's a gap here: within a single `Score` call each batch is loaded at most once, but a respeculate wave runs the scorer once per tree, so shared dependency batches do get re-read once per tree. Context on why the reads exist at all: the tree input only carries path structure, while the formula needs each dependency batch's current `State`/`Score`, which live on the batch rows — so the scorer resolves them through the injected `BatchStore` per the extension contract (identity in, resolve internally). These are primary-key point reads bounded by the head's dependency count, so I'd like to keep the contract as-is for now — and if this shows up as real load, we have options that don't touch the seam: e.g. a read-through caching `BatchStore` decorator scoped to a respeculate cycle in wiring, or batching the scorer's input at queue granularity. Happy to explore those in a follow-up.

submitqueue/extension/speculation/pathscorer/probability/probability.go
behinddwalls: removed

PR #336, opened 2026-07-10
doc(stovepipe): build and buildsignal pipeline steps

roychying:
## Why?
Stovepipe's pipeline had design docs for process but nothing for the two stages downstream of it - build (triggers the build-runner) and buildsignal (polls it to completion). Both stages needed a documented contract before implementation could start, in particular because stovepipe can't reuse SubmitQueue's BuildRunner interface as-is: SubmitQueue validates stacks of batches, stovepipe validates a single URI against a baseline, and that mismatch has to be resolved deliberately rather than discovered mid-implementation.

## What?
- doc/rfc/stovepipe/steps/build.md (new): documents the build stage end to end - the trigger-only algorithm, idempotency posture under at-least-once redelivery, error classification, edge cases (head == last-green, Phase 2 re-triggers), and the new Build entity/BuildStore/queue contract it needs. Defines the stovepipe BuildRunner contract (design sketch) with URI-based Trigger(headURI, baseURI, metadata) instead of SubmitQueue's batch-based trigger, explains why the contracts are kept separate (materialize-vs-checkout mismatch, baseline semantics batches can't express, URI ownership), and lays out - as open discussion, not settled decisions - alternative shapes for sharing the contract (single platform-level interface, metadata-smuggled scope, shared-backend-thin-contracts) and alternatives for how the build id could be keyed/deduped.
- doc/rfc/stovepipe/steps/buildsignal.md (new): documents the buildsignal poll loop - reusing SubmitQueue's buildsignal shape (poll cadence, PublishAfter-not-Nack primitive, error classification) pipe diverges: it only forwards to record once terminal

## Test Plan
N/A


doc/rfc/stovepipe/steps/build.md
```diff
@@ -0,0 +1,351 @@
+# Build stage
+
+`build` triggers the build-runner for the scope `process` already decided and hands the resulting build id to `buildsignal`. See [workflow.md](../workflow.md) for where it sits in the pipeline, and [process.md](process.md) for how the scope it reads (`BuildStrategy`, `BaseURI`) is chosen.
+
+It handles only the trigger: it does not poll for completion, record greenness, or decide incremental-vs-full — those are `buildsignal`'s, `record`'s, and `process`'s jobs respectively.
+
+`build` is structurally the same controller as `submitqueue/orchestrator/controller/build/build.go`, and [doc/rfc/submitqueue/build-runner.md](../../submitqueue/build-runner.md) is the reference rationale for the trigger-then-poll shape this stage reuses. The `BuildRunner` contract's `Status`/`Cancel` half now carries over as literally shared code with SubmitQueue, via `platform/extension/buildrunner`; what does *not* carry over is `Trigger` — the two domains model different problems, so each keeps its own, following the [extension-contract.md](../../submitqueue/extension-contract.md) "identity in, resolve internally" principle.
```
behinddwalls: can we use relative to root paths than relative to current

behinddwalls: do we need to have it as buildrunner extension at the platform vs per technologu say `buildkite`, i don't think the idea of breaking the contract in places so defining somewhat similar thing in 2 places for status and cancel is fine..and we abstract away by the underlying impl like buildkite to so ? thoughts

doc/rfc/stovepipe/steps/build.md
```diff
@@ -0,0 +1,351 @@
...
+
+## Error classification
+
+Per `platform/errs`' non-retryable-by-default rule (see [platform/errs/README.md](../../../../platform/errs/README.md)), the controller returns raw errors and lets the consumer's classifier decide; it opts into retryability explicitly only where it knows something the classifier cannot:
+
+| Failure | Disposition | Why |
+|---|---|---|
+| `Request` not found (`storage.ErrNotFound`) | retryable (`errs.NewRetryableError`) | Producing stage's write not visible yet; redelivery converges — same as `process.go`. |
```
behinddwalls: why it wouldn't be visible?

doc/rfc/stovepipe/steps/build.md
```diff
@@ -0,0 +1,351 @@
...
+| Factory lookup | non-retryable | A queue with no registered builder is a config error, not transient. |
+| Malformed message | non-retryable | The payload is broken; retrying can't fix it. |
+| `Trigger` | raw error; classifier decides | Runner timeout/connection is transient (may be classified retryable); a bad URI is permanent. |
+| Build store, other than `ErrAlreadyExists` | non-retryable | Storage down; DLQ reconciliation recovers. |
+| Publish to buildsignal | non-retryable | Not wrapped retryable, per the errs rule against retryable publishes. The message dead-letters, the fail-closed reconciler forces a conservative not-green, and an operator republish can still resume polling later — a late green wins over the conservative verdict via CAS ([workflow.md](../workflow.md#fail-closed-on-unprocessable-work)). |
+
+Everything non-retryable rejects to DLQ, where the fail-closed path forces a conservative not-green so a Request never wedges its Queue's slot ([workflow.md](../workflow.md#fail-closed-on-unprocessable-work)).
```
behinddwalls: by default everything/unclassified in non-retryable so we can document the ones we want to retry

doc/rfc/stovepipe/steps/build.md
```diff
@@ -0,0 +1,351 @@
...
+    // Trigger starts a new build every call and mints the build's identity —
+    // there is no caller-supplied dedup input, matching SubmitQueue's contract
+    // exactly (see "Alternatives considered for the build identity" below
+    // for other shapes this doc considered). headURI is the commit
+    // under validation; baseURI is the incremental baseline (empty for a full
+    // build). metadata is caller annotations the runner may echo but must not
+    // depend on. Runner-side work is async; callers learn progress via Status.
+    // Returns the runner-assigned build id, which the caller adopts as Build.ID.
+    Trigger(ctx context.Context, headURI, baseURI string, metadata base.BuildMetadata) (base.BuildID, error)
+}
```
behinddwalls: don't this is giving us too much so i would prefer it separate for SQ and Stopvepipe and try to re-use underlying buildkite contracts as much as possible..

doc/rfc/stovepipe/steps/build.md
```diff
@@ -0,0 +1,351 @@
...
+
+The shape isn't decided here because project semantics belong to `analyze`, not `build`: how a project maps to a buildable scope (a Bazel target pattern, a directory, a service name) is implementer-specific per [workflow.md](../workflow.md#project---greenness-at-a-finer-grain). The expectation is that this stays an opaque token — following the same "identity in, resolve internally" shape already used for `headURI`/`baseURI` (owned and interpreted by `SourceControl`) — that `build` reads off the `Request`/message and hands to the runner uninterpreted, rather than a structured type `build` would have to understand:
+
+```go
+Trigger(ctx context.Context, headURI, baseURI string, projectScope entity.ProjectScope, metadata base.BuildMetadata) (base.BuildID, error)
```
behinddwalls: should be added anyway in both SQ and Stovepipe... currently there no metadata as such from SQ to be passed but i can think of things which might be valuable to pass like conflict graph info or any other decisions which maybe relevant to build.

cc: @JamyDev

doc/rfc/stovepipe/steps/build.md
```diff
@@ -0,0 +1,351 @@
...
+
+### Target graph: not part of `Status`, resolved separately
+
+`Status` does not return a `TargetGraph` alongside `BuildStatus`. `buildsignal`'s poll loop runs for every build in the pipeline, including every Phase-2 project build, but only `analyze` ever needs a target graph, and only once, right after a Phase-1 build succeeds. Routing it through `Status`/`buildsignal` would mean every poll tick — and every Phase-2 build, which has no further target-to-project mapping to do — computes or shuttles a graph almost nothing reads. Per the extension-contract's "identity in, resolve internally" rule, `analyze` resolves the target graph itself from the identity it already has (`Build.ID`, arriving via `record`'s fan-out), rather than receiving it pre-resolved through a chain of controllers that don't need it. See [Alternatives considered for the target graph](#alternatives-considered-for-the-target-graph) for the shapes weighed for how `analyze` does that, and for the open question of what `TargetGraph` itself contains.
+
+### Alternatives considered for the target graph
```
behinddwalls: we can skip from this rfc, we should focus on it separately

behinddwalls (changes requested):

doc/rfc/stovepipe/steps/build.md
roychying: Probably the commit-then-publish order cannot guarantee read-your-writes for a different reader on every backend? that is, a replica or an eventually-consistent store can still serve build's Get a stale miss. Similar design is also in `process.md` (L20)

behinddwalls: i think that's fine..it's more of constraint/contract for the build. in think case, we may want to use the primary system for all the work we need to do here vs a replica

doc/rfc/stovepipe/steps/build.md
roychying: 1. paths fixed.
2. yes and this is one of the alternatives I'd already recorded. The shared-interface route does give a compile-time guarantee that both domains' Status/Cancel stay identical (I believe they should be?), but it costs fragmenting one conceptual contract across two packages, plus migrating SubmitQueue's already-shipped code onto promoted platform/base types for a half only SubmitQueue uses today. Your approach is what the doc now adopts.

doc/rfc/stovepipe/steps/build.md
roychying: removed

doc/rfc/stovepipe/steps/build.md
roychying: i see. removed the `not visible` statement but keep it as a retryable failure since it's cheap insurance

doc/rfc/stovepipe/steps/build.md
roychying: Shall we decided the concrete shape here? my preference is leaving it deferred until there's a real consumer in either domain..

doc/rfc/stovepipe/steps/build.md
roychying: removed redundant content

doc/rfc/stovepipe/steps/build.md
roychying: same above as the 1st thread

behinddwalls (approved):

PR #340, opened 2026-07-13
[1/N] feat(storage): add request read-model stores

albertywu:
## Summary
Add storage contracts, MySQL implementations, generated mocks, and integration coverage for request summaries, queue projections, and change URI mappings.

## Test Plan
✅ `make fmt && make build && make test && make e2e-test`

## Stack
1. @ #340
1. #341
1. #342
1. #343
1. #344
1. #345
1. #346

PR #342, opened 2026-07-13
[3/N] feat(request): materialize request log status

albertywu:
## Summary
Add projection winner selection, optimistic-concurrency convergence, and queue-summary repair without activating production call sites yet.

## Test Plan

## Stack
1. #340
1. #341
1. @ #342
1. #343
1. #344
1. #345
1. #346

### DIFF
--- a/submitqueue/core/request/materializer.go
+++ b/submitqueue/core/request/materializer.go
@@ -1,0 +1,184 @@
+// Copyright (c) 2025 Uber Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package request
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"maps"
+	"slices"
+
+	"github.com/uber/submitqueue/submitqueue/entity"
+	"github.com/uber/submitqueue/submitqueue/extension/storage"
+)
+
+// Materializer appends request logs and projects the winning public request state.
+// It owns winner selection, optimistic concurrency, and public projection repair.
+type Materializer struct {
+	store storage.Storage
+}
+
+// NewMaterializer creates a request read-model materializer.
+func NewMaterializer(store storage.Storage) *Materializer {
+	return &Materializer{store: store}
+}
+
+// PersistLog appends one audit log and materializes its winning state.
+// Projection errors are returned so queue deliveries are retried rather than silently dropping the side write.
+// Because the append happens first, retrying after a projection failure may retain another copy of the event in History.
+func (m *Materializer) PersistLog(ctx context.Context, log entity.RequestLog) error {
+	if err := m.store.GetRequestLogStore().Insert(ctx, log); err != nil {
+		return fmt.Errorf("failed to insert request log request_id=%s: %w", log.RequestID, err)
+	}
+
+	for {
+		summary, err := m.store.GetRequestSummaryStore().Get(ctx, log.RequestID)
+		if err != nil {
+			return fmt.Errorf("failed to get request summary request_id=%s: %w", log.RequestID, err)
+		}
+
+		if logWins(log, summary) {
+			oldVersion := summary.Version
+			newVersion := oldVersion + 1
+			updated := summary
+			updated.Status = log.Status
+			updated.RequestVersion = log.RequestVersion
+			updated.StatusTimestampMs = log.TimestampMs
+			updated.LastError = log.LastError
+			updated.Metadata = cloneMetadata(log.Metadata)
+
+			if err := m.store.GetRequestSummaryStore().Update(ctx, updated, oldVersion, newVersion); err != nil {
+				if errors.Is(err, storage.ErrVersionMismatch) {
+					continue
+				}
+				return fmt.Errorf("failed to update request summary request_id=%s: %w", log.RequestID, err)
+			}
+			updated.Version = newVersion
+			summary = updated
+		}
+
+		if err := m.repairPublicProjections(ctx, summary); err != nil {
+			return err
+		}
+		return nil
+	}
+}
+
+// repairPublicProjections activates and repairs the public query projections.
+// URI mappings are created before the queue summary, which acts as the marker that activation completed.
+func (m *Materializer) repairPublicProjections(ctx context.Context, authoritative entity.RequestSummary) error {
+	desired := queueSummaryFromSummary(authoritative)
+	for {
+		current, err := m.store.GetRequestQueueSummaryStore().Get(ctx, desired.Queue, desired.ReceivedAtMs, desired.RequestID)
+		if errors.Is(err, storage.ErrNotFound) {
+			if err := m.createURIMappings(ctx, authoritative); err != nil {
+				return err
+			}
+			if err := m.store.GetRequestQueueSummaryStore().Create(ctx, desired); err != nil {
+				if errors.Is(err, storage.ErrAlreadyExists) {
+					continue
+				}
+				return fmt.Errorf("failed to recreate queue summary request_id=%s: %w", desired.RequestID, err)
+			}
+			return nil
+		}
+		if err != nil {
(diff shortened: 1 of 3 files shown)


submitqueue/core/request/materializer.go
```diff
@@ -0,0 +1,130 @@
...
+			if errors.Is(err, storage.ErrVersionMismatch) {
+				continue
+			}
+			return fmt.Errorf("failed to update queue summary request_id=%s: %w", desired.RequestID, err)
+		}
+		return nil
+	}
+}
+
+func logWins(log entity.RequestLog, summary entity.RequestSummary) bool {
```
behinddwalls, line 134: probably deserves some comments

submitqueue/core/request/materializer.go
```diff
@@ -0,0 +1,130 @@
...
+	"errors"
+	"fmt"
+
+	"github.com/uber/submitqueue/submitqueue/entity"
+	"github.com/uber/submitqueue/submitqueue/extension/storage"
+)
+
+// Materializer appends request logs and projects the winning current status.
+// Storage implementations remain mechanical; this type owns winner selection, optimistic concurrency, and projection repair.
+type Materializer struct {
```
behinddwalls, line 30: thinking if there is better name for this like RequestReconciler or something

submitqueue/core/request/materializer.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+}
+
+func logWins(log entity.RequestLog, summary entity.RequestSummary) bool {
+	incomingTerminal := isVersionedTerminal(log)
+	currentTerminal := isVersionedTerminalSummary(summary)
+	if incomingTerminal != currentTerminal {
+		return incomingTerminal
+	}
+	if incomingTerminal {
```
behinddwalls, line 150: When both the incoming log and the summary are versioned-terminal with **equal** `RequestVersion`, the decision falls through to the wall-clock tie-break at line 121 — so a *different* terminal status can displace an existing one purely by having a later timestamp. This is exactly the path the DLQ reconciler hits in the next PR of this stack: `failRequest` writes an unconditional `error` log at the request's current version, conclude's `landed` log carries that same version, and the DLQ log is always stamped later — so the projection flips a landed request to `error` (details on #343). Suggest: at equal RequestVersion, a log whose terminal status *differs* from the summary's terminal status should not win on timestamp alone — same-status redelivery stays idempotent, and genuinely newer terminal states still win via the version comparison at 117-119.

submitqueue/core/request/materializer.go
```diff
@@ -0,0 +1,130 @@
...
+
+// PersistLog appends one audit log and materializes its winning state.
+// Projection errors are returned so queue deliveries are retried rather than silently dropping the side write.
+func (m *Materializer) PersistLog(ctx context.Context, log entity.RequestLog) error {
+	if err := m.store.GetRequestLogStore().Insert(ctx, log); err != nil {
+		return fmt.Errorf("failed to insert request log request_id=%s: %w", log.RequestID, err)
+	}
+
+	for {
+		summary, err := m.store.GetRequestSummaryStore().Get(ctx, log.RequestID)
```
behinddwalls, line 48: `PersistLog` hard-fails when the summary row is absent — including `ErrNotFound`. Summary rows only exist for requests admitted after #341, so any request in flight across the deploy that activates this (#343) has logs but no summary: the gateway log consumer will classify `ErrNotFound` non-retryable and dead-letter into `log_dlq` (which has no registered consumer — status silently frozen), and orchestrator DLQ consumers run `AlwaysRetryableProcessor`, so `failRequest` → `PersistLog` loops forever, appending a fresh salted `request_log` row on every redelivery. I couldn't find a backfill anywhere in the stack. Either create-on-missing here (upsert a summary from the log — it carries enough to synthesize one), or gate #343 on a backfill migration.

Separately: `log_dlq` having no registered consumer deserves its own ticket independent of this rollout concern — anything non-retryable from the log consumer currently parks there invisibly, with no reconciliation and no alerting.

submitqueue/core/request/materializer.go
```diff
@@ -0,0 +1,130 @@
...
+			}
+			return nil
+		}
+		if err != nil {
+			return fmt.Errorf("failed to get queue summary request_id=%s: %w", desired.RequestID, err)
+		}
+		if current.Version == desired.Version {
+			return nil
+		}
+		if current.Version > desired.Version {
```
behinddwalls, line 104: `current.Version > desired.Version` isn't a corruption signal — it means a concurrent materializer already repaired the queue projection from a **newer** authoritative snapshot, i.e. convergence already succeeded. Returning an error here turns a benign race into: (a) a failed Cancel RPC *after* the cancelling intent was recorded but *before* the cancel message is published (once #343 wires it), and (b) a dead-lettered log message for a fully-converged write. The projection version is monotone and strictly follows the authoritative row, so ahead-of-desired is safe to treat as success. The `queue projection ahead fails fast` test at `materializer_test.go:139` currently enshrines the error behavior and would flip to asserting convergence.

submitqueue/core/request/materializer.go
```diff
@@ -0,0 +1,130 @@
...
+
+// NewMaterializer creates a request read-model materializer.
+func NewMaterializer(store storage.Storage) *Materializer {
+	return &Materializer{store: store}
+}
+
+// PersistLog appends one audit log and materializes its winning state.
+// Projection errors are returned so queue deliveries are retried rather than silently dropping the side write.
+func (m *Materializer) PersistLog(ctx context.Context, log entity.RequestLog) error {
+	if err := m.store.GetRequestLogStore().Insert(ctx, log); err != nil {
```
behinddwalls, line 43: The salted log insert runs before up to five more fallible round trips, and every nack→redelivery re-inserts a new row (the salt guarantees no dedup). The test at `materializer_test.go:113` documents this as an accepted trade-off, but two things change the calculus in this stack: the duplicates now surface in the user-facing History API (#346), and combined with the missing-summary loop above, the duplication becomes unbounded rather than per-crash.

The duplication is avoidable for retries of the *same* logical event: derive the row identity from the event instead of a random salt — e.g. key on `(request_id, status, request_version, timestamp_ms)` or thread the queue delivery's message ID through as the idempotency key — so a redelivered insert hits the primary key and becomes a no-op (`ErrAlreadyExists` → continue), matching how the rest of the pipeline treats redelivery. A cheaper variant is an existence check before insert on retry attempts, though the PK approach makes dedup exact rather than best-effort. Fine as a follow-up ticket if not in this PR, but worth stating the chosen bound explicitly in the doc comment.

submitqueue/core/request/materializer.go
albertywu: Added comments covering Materializer responsibility, retained duplicate history on retry, public projection activation and repair, winner selection, and the monotone ahead-projection case.

submitqueue/core/request/materializer.go
albertywu: Kept `Materializer` because this component appends retained request logs and materializes their public read projections. `RequestReconciler` would imply broader operational reconciliation than this type owns.

submitqueue/core/request/materializer.go
albertywu: Fixed the contradictory event at its source in PR #343. `failRequest` now skips error materialization for `Landed` and `Cancelled`, and `failBatch` skips error fan-out for `Succeeded` and `Cancelled`. Existing `Error` and `Failed` states still replay for repair. This prevents the divergent equal-version terminal event while preserving the established winner rules documented in [Request-Log Materialization](https://github.com/uber/submitqueue/blob/wua/request-read-model-storage/doc/rfc/submitqueue/status-list-api.md#request-log-materialization).

submitqueue/core/request/materializer.go
albertywu: No backfill is required for this rollout because SubmitQueue is pre-production and we are starting fresh. PR #341 creates the authoritative receipt before any public event can be materialized. The missing `log_dlq` consumer is a separate operational follow-up rather than part of this read-API stack.

submitqueue/core/request/materializer.go
albertywu: Changed this case to return success when `current.Version > desired.Version`. It now treats the newer projection as successful concurrent convergence and does not overwrite it. The test was updated accordingly.

submitqueue/core/request/materializer.go
albertywu: Retained duplicate rows are an intentional receipt-history tradeoff for this pre-production design. `PersistLog` now documents that a retry after projection failure may retain another copy of the event. The missing-summary retry loop does not apply to the fresh rollout because admission creates the authoritative receipt first. Exact retry deduplication can be added later if production requirements call for it.

behinddwalls (approved):

PR #343, opened 2026-07-13
[4/N] feat(submitqueue): activate request status materialization

albertywu:
## Summary
Route synchronous, queued, and DLQ request-log writes through the shared materializer so authoritative and queue projections converge with retained history.

Preserve the original DLQ failure reason as the terminal request last error.

## Test Plan

## Stack
1. @ #343
1. #344
1. #345
1. #346


submitqueue/orchestrator/controller/dlq/dlq.go
```diff
@@ -116,9 +118,9 @@ func failRequest(ctx context.Context, store storage.Storage, logger *zap.Sugared
 	// A duplicate log entry from such a retry is the accepted trade-off; a
 	// missing one would leave the gateway-visible status divergent from the
 	// entity state.
-	logEntry := entity.NewRequestLog(requestID, entity.RequestStatusError, logVersion, "", nil)
-	if err := store.GetRequestLogStore().Insert(ctx, logEntry); err != nil {
-		return fmt.Errorf("failed to insert request log for %s: %w", requestID, err)
+	logEntry := entity.NewRequestLog(requestID, entity.RequestStatusError, logVersion, lastError, nil)
+	if err := requestcore.NewMaterializer(store).PersistLog(ctx, logEntry); err != nil {
```
behinddwalls: This unconditional `PersistLog(error)` is now authoritative-status-bearing, and it runs even when the request is already in a *different* terminal state — with `logVersion` left at the request's current version on that branch. Conclude's `landed` log carries that same post-CAS version, so in `logWins` the two are versioned-terminal at equal `RequestVersion` and the decision falls to the timestamp tie-break — which this reconcile-time log always wins. Net effect: a request that landed (or cancelled) gets its user-visible Status/List state flipped to `error` whenever a stale message for it dead-letters afterward. `failBatch` amplifies it: an already-**Succeeded** batch still "fans out only" and flips every landed member. Before this PR the spurious error log was cosmetic history noise; now it overwrites the answer users see.

Fix shape: mirror conclude's divergent-terminal skip (`conclude.go:128-140`) — only write the error log when the terminal state is actually `Error` (i.e. this reconcile owns it, including the repair-on-redelivery case the comment above describes); a genuinely landed/cancelled request needs no error entry. The `logWins` tie-break on #342 is the belt to this suspender.

albertywu: Fixed as suggested. An existing `Error` request repeats materialization for repair, while `Landed` and `Cancelled` return success without an error log. An existing `Failed` batch repeats fan-out for repair, while `Succeeded` and `Cancelled` skip error fan-out. Focused tests cover each terminal path.

submitqueue/orchestrator/controller/dlq/dlq.go
```diff
@@ -101,24 +108,19 @@ func failRequest(ctx context.Context, store storage.Storage, logger *zap.Sugared
...
-	// the state but then failed to insert the log is repaired on redelivery.
-	// A duplicate log entry from such a retry is the accepted trade-off; a
-	// missing one would leave the gateway-visible status divergent from the
-	// entity state.
-	logEntry := entity.NewRequestLog(requestID, entity.RequestStatusError, logVersion, "", nil)
-	if err := store.GetRequestLogStore().Insert(ctx, logEntry); err != nil {
-		return fmt.Errorf("failed to insert request log for %s: %w", requestID, err)
+	// An existing Error state still writes the log so a previous attempt that
+	// changed the entity but failed materialization can be repaired.
+	logEntry := entity.NewRequestLog(requestID, entity.RequestStatusError, logVersion, lastError, nil)
```
behinddwalls, line 115: should this be publishing to log topic than writing directly?

albertywu: I kept direct persistence here intentionally. Publishing would make DLQ reconciliation depend on another asynchronous consumer and its DLQ behavior. Direct materialization keeps this delivery unacknowledged until the terminal log and public projections are repaired. I also corrected the ownership comments so they no longer imply the log consumer is the only writer.

behinddwalls: i think the concern here is that Orchestrator has no access to request log directly so only way to make it there is through `log` queue which is consumed and written by gateway... gateway is sole write to the request_log so this will likely not work when deployed end to end and fail on write due to missing table

albertywu: You are right. The local stack masked this because Gateway and Orchestrator shared the same MySQL root DSN and schemas. I changed DLQ reconciliation to publish the terminal RequestLog to the log topic instead. Gateway remains the sole writer of request logs and public projections. Publish failures are returned so the always-retryable DLQ consumer redelivers, and an already-Error request republishes the same logical event to repair an earlier failed publish. I added focused publisher tests and reran `make fmt && make build && make test && make e2e-test` on PR #343 and each descendant.

behinddwalls (approved):

behinddwalls:

The rebase of `wua/request-projection-activation` completed without conflicts, but the resulting code diff does not match the original. This means the rebase silently altered code content. The branch was **not** force-pushed.

behinddwalls:

The rebase of `wua/request-projection-activation` completed without conflicts, but the resulting code diff does not match the original. This means the rebase silently altered code content. The branch was **not** force-pushed.

behinddwalls:

The rebase of `wua/request-projection-activation` completed without conflicts, but the resulting code diff does not match the original. This means the rebase silently altered code content. The branch was **not** force-pushed.

PR #347, opened 2026-07-13
feat(stovepipe): add DLQ reconciliation for the process stage

gsturges:
## Summary

A request whose `process`-stage message exhausts retries would previously sit in a non-terminal state forever — indistinguishable from "not yet validated" to callers gating deployments on greenness. Per the fail-closed policy in [doc/rfc/stovepipe/workflow.md](doc/rfc/stovepipe/workflow.md), such requests must be driven to a conservative not-green terminal state instead.

- **Entity**: add `RequestStateFailed` terminal state and `IsRequestStateTerminal` helper.
- **New `stovepipe/controller/dlq` package**: a `consumer.Controller` that drains the `process_dlq` topic and CAS-transitions the affected request to `failed`. If the request had already been admitted (`processing`), the queue's `in_flight_count` slot is released first, per the integrity rule in [doc/rfc/stovepipe/steps/process.md](doc/rfc/stovepipe/steps/process.md).
- **Server wiring**: second consumer using `errs.AlwaysRetryableProcessor` (a reconciliation failure retries in place — the DLQ subscription is a final destination with `DLQ.Enabled = false`, so nothing can cascade to a `_dlq_dlq`); `process_dlq` registered in the topic registry with raised retry attempts; DLQ consumer starts first and stops last.

Reconciliation is idempotent: already-terminal requests are skipped, and optimistic locking lets a late legitimate pipeline transition win. The design mirrors SubmitQueue's proven `submitqueue/orchestrator/controller/dlq` pattern, simplified for stovepipe's single payload shape.

Only `process` is a queue-consuming step today (`ingest` is synchronous RPC); future stages (`build`, `buildsignal`, `record`) can reuse the same pattern.

## Test plan

- [x] New table-driven unit tests in `stovepipe/controller/dlq/dlq_test.go`: accepted→failed, processing→failed with slot release, already-terminal no-ops, not-found no-op, CAS version-mismatch paths, malformed/empty payloads
- [x] `go build ./...` and `go test ./stovepipe/... ./service/stovepipe/...` pass
- [x] `make gazelle`, `make fmt`, `make lint-license` clean

### DIFF
--- a/platform/extension/messagequeue/subscription_config.go
+++ b/platform/extension/messagequeue/subscription_config.go
@@ -78,6 +78,23 @@ type DLQConfig struct {
 	TopicSuffix string
 }
 
+// DLQSubscriptionConfig returns a SubscriptionConfig for consuming a dead-letter
+// topic (DLQ reconciliation). It starts from DefaultSubscriptionConfig and applies
+// the two overrides every DLQ consumer needs:
+//
+//   - DLQ.Enabled is false, so a reconciliation failure retries in place instead of
+//     cascading to a second-level "_dlq_dlq" topic that nobody consumes.
+//   - Retry.MaxAttempts is a very high backstop so the per-message retry budget
+//     effectively never runs out. This pairs with errs.AlwaysRetryableProcessor
+//     wired into the DLQ consumer: reconciliation converges eventually instead of
+//     being silently dropped after the default retry count.
+func DLQSubscriptionConfig(subscriberName, consumerGroup string) SubscriptionConfig {
+	config := DefaultSubscriptionConfig(subscriberName, consumerGroup)
+	config.DLQ.Enabled = false
+	config.Retry.MaxAttempts = 1000
+	return config
+}
+
 // DefaultSubscriptionConfig returns a SubscriptionConfig with sensible defaults.
 func DefaultSubscriptionConfig(subscriberName, consumerGroup string) SubscriptionConfig {
 	return SubscriptionConfig{
--- a/service/stovepipe/server/main.go
+++ b/service/stovepipe/server/main.go
@@ -341,14 +341,8 @@ func run() error {
 // newTopicRegistry builds the TopicRegistry for Stovepipe's internal pipeline queues. ingest
 // publishes to process; process publishes admitted requests to the publish-only build topic.
 // The process_dlq topic is the dead-letter destination the queue backend routes to (per
-// DefaultSubscriptionConfig's DLQ.TopicSuffix) when the process controller exhausts retries;
-// DLQ.Enabled is false on its own subscription so a reconciliation failure retries in place
-// rather than cascading to a further DLQ.
+// DefaultSubscriptionConfig's DLQ.TopicSuffix) when the process controller exhausts retries.
 func newTopicRegistry(q extqueue.Queue, subscriberName string) (consumer.TopicRegistry, error) {
-	dlqSub := extqueue.DefaultSubscriptionConfig(subscriberName, "stovepipe-process-dlq")
-	dlqSub.DLQ.Enabled = false
-	dlqSub.Retry.MaxAttempts = 1000
-
 	return consumer.NewTopicRegistry([]consumer.TopicConfig{
 		{
 			Key:   stovepipemq.TopicKeyProcess,
@@ -367,7 +361,7 @@ func newTopicRegistry(q extqueue.Queue, subscriberName string) (consumer.TopicRe
 			Key:          dlq.TopicKey(stovepipemq.TopicKeyProcess),
 			Name:         "process_dlq",
 			Queue:        q,
-			Subscription: dlqSub,
+			Subscription: extqueue.DLQSubscriptionConfig(subscriberName, "stovepipe-process-dlq"),
 		},
 	})
 }
--- a/service/stovepipe/server/main.go
+++ b/service/stovepipe/server/main.go
@@ -37,6 +37,7 @@ import (
 	queueMySQL "github.com/uber/submitqueue/platform/extension/messagequeue/mysql"
 	"github.com/uber/submitqueue/service/stovepipe/server/mapper"
 	"github.com/uber/submitqueue/stovepipe/controller"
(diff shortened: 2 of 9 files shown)


stovepipe/controller/dlq/request.go
```diff
@@ -0,0 +1,118 @@
...
+	defer func() { op.Complete(retErr) }()
+
+	msg := delivery.Message()
+
+	pr := &stovepipemq.ProcessRequest{}
+	if err := stovepipemq.Unmarshal(msg.Payload, pr); err != nil {
+		metrics.NamedCounter(c.metricsScope, _opName, "deserialize_errors", 1)
+		// Malformed DLQ payload is non-retryable: a re-delivery will decode the same
+		// bytes and fail the same way.
+		return fmt.Errorf("failed to decode dlq payload: %w", err)
```
mnoah1, line 87: This says non-retryable, but paired with the `AlwaysRetryableProcessor` will keep retrying when this error is returned.  Maybe it's still reasonable to retry this case (e.g. if a deployment changes the payload shape and requests will eventually get handled when an upgrade completes on the host), if so update comments to clarify one way or the other.

service/stovepipe/server/main.go
```diff
@@ -296,8 +324,15 @@ func run() error {
 }
 
 // newTopicRegistry builds the TopicRegistry for Stovepipe's internal pipeline queues. ingest
-// publishes to the process topic and the process consumer subscribes to it.
+// publishes to the process topic and the process consumer subscribes to it. The process_dlq
+// topic is the dead-letter destination the queue backend routes to (per DefaultSubscriptionConfig's
+// DLQ.TopicSuffix) when the process controller exhausts retries; DLQ.Enabled is false on its own
+// subscription so a reconciliation failure retries in place rather than cascading to a further DLQ.
 func newTopicRegistry(q extqueue.Queue, subscriberName string) (consumer.TopicRegistry, error) {
+	dlqSub := extqueue.DefaultSubscriptionConfig(subscriberName, "stovepipe-process-dlq")
```
mnoah1: add an `extqueue.DLQSubscriptionConfig` helper that sets these values then you can just call it in the `Subscription` key similar to the DefaultSubscriptionConfig

stovepipe/controller/dlq/dlq.go
```diff
@@ -0,0 +1,143 @@
...
+// the suffix literal.
+func TopicKey(main consumer.TopicKey) consumer.TopicKey {
+	return consumer.TopicKey(string(main) + topicSuffix)
+}
+
+// failRequest transitions request to RequestStateFailed if it is not already in a
+// terminal state. If the request had reached RequestStateProcessing — meaning process's
+// admit step already CAS-incremented the queue's in_flight_count for it — the queue's
+// slot is released first, so a crash between the two writes leaves the count still
+// bound to a non-terminal request rather than double-released.
```
mnoah1: Is this true?  Current ordering is: 1) release slot 2) potential crash 3) request state never gets updated out of processing, and could cause this whole logic to run again.  In that cause it could get double released.

Either way I think it's ok for now, I don't foresee this being a typical issue and better to leave the queue unblocked than potentially deadlock it.

service/stovepipe/server/main.go
gsturges: updated

stovepipe/controller/dlq/dlq.go
gsturges: You are correct, I updated comment

stovepipe/controller/dlq/request.go
gsturges: correct again. Updated

gsturges:
Rebased onto main. Besides the review feedback already discussed (the `DLQSubscriptionConfig` helper and the corrected `failRequest` / decode-error comments), the rebase required one substantive adaptation worth calling out:

**The reconciler now forces `recorded_not_green` instead of introducing a separate `failed` state.** While this PR was in review, main landed the terminal Request states (#362): `recorded_green` / `recorded_not_green` with an `IsTerminal()` method, and the `RequestStateRecordedNotGreen` doc comment (plus workflow.md's fail-closed section) already designates it as the conservative verdict the DLQ reconciler writes. Keeping this PR's `RequestStateFailed` would have meant two overlapping terminal-not-green states, so this branch now:

- no longer touches `stovepipe/entity/request.go` (no `RequestStateFailed`, no `IsRequestStateTerminal` free function),
- transitions reconciled requests to `entity.RequestStateRecordedNotGreen` and checks `request.State.IsTerminal()`.

Reconciliation behavior is otherwise unchanged (idempotent terminal-skip, slot release before the terminal CAS, always-retryable pairing). `newTopicRegistry` keeps main's publish-only `build` topic alongside the new `process_dlq` subscription.

mnoah1 (approved):

platform/extension/messagequeue/subscription_config.go
```diff
@@ -78,6 +78,23 @@ type DLQConfig struct {
...
+//   - DLQ.Enabled is false, so a reconciliation failure retries in place instead of
+//     cascading to a second-level "_dlq_dlq" topic that nobody consumes.
+//   - Retry.MaxAttempts is a very high backstop so the per-message retry budget
+//     effectively never runs out. This pairs with errs.AlwaysRetryableProcessor
+//     wired into the DLQ consumer: reconciliation converges eventually instead of
+//     being silently dropped after the default retry count.
+func DLQSubscriptionConfig(subscriberName, consumerGroup string) SubscriptionConfig {
+	config := DefaultSubscriptionConfig(subscriberName, consumerGroup)
+	config.DLQ.Enabled = false
+	config.Retry.MaxAttempts = 1000
```
behinddwalls, line 94: should we just add -1? thoughts? i don't recall if messaging queue support that symantix but i think it might

behinddwalls (approved):

platform/extension/messagequeue/subscription_config.go
gsturges: I had claude go through the logic to double check me. Did not see logic for -1 to be used

PR #348, opened 2026-07-14
fix(submitqueue): keep internal merge stage topic distinct from runway merge queue

sbalabanov:
## Summary

The orchestrator registers its internal merge pipeline stage and the runway-owned merge queue in one topic registry over one shared queue backend, and both used the topic key `"merge"` (`topickey.TopicKeyMerge == runwaymq.TopicKeyMerge`). The publish-only runway entry silently shadowed the internal stage entry (last write wins on the registry maps), collapsing the two topics into one:

- runway consumed the internal batch-ID messages as **zero-step MergeRequests** and reported `SUCCEEDED` without ever seeing the real steps;
- the orchestrator merge controller's real `MergeRequest` publish was then **silently dropped** by the queue's `(topic, partition_key, id)` dedupe, because it reuses the batch ID already published to the same topic.

The pipeline still reached `landed` only because the zero-step result correlates by batch ID — with a real merger, nothing would ever have been merged. Found by the new e2e harness (stacked PR) asserting that runway's merge signal carries the request's step IDs.

## Changes

- Namespace the colliding merge topic keys by owner so they can never collapse into each other again:
  - `api/runway/messagequeue`: the runway-owned merge queue key/wire name becomes `runway-merge` (proto `topic_keys` option + generated descriptor regenerated).
  - `submitqueue/core/topickey`: the internal orchestrator merge stage key/wire name becomes `submitqueue-merge`.
- `platform/consumer`: `NewTopicRegistry` now rejects duplicate topic keys, so a future collision fails at startup instead of silently misrouting messages.
- `service/submitqueue/orchestrator` + `service/runway`: register both stages under their new, distinct wire names.

## Test plan

- `bazel test //... --test_tag_filters=-manual,-integration` — 75/75 pass (includes the updated dlq controller test).
- `bazel test //test/integration/...` — 8/8 pass.
- e2e suite (with the stacked harness PR): runway's merge signal now carries the request step IDs and the full land pipeline completes in ~2s.


submitqueue/core/topickey/topickey.go
```diff
@@ -42,7 +42,11 @@ const (
 	// PublishAfter when the build has not yet reached a terminal state.
 	TopicKeyBuildSignal TopicKey = "buildsignal"
 	// TopicKeyMerge is the pipeline stage where speculated batches are published for merging.
-	TopicKeyMerge TopicKey = "merge"
```
behinddwalls, line 53: should we namespace these to say `runway-merge` vs just `merge` and same from SQ stuff to be `submitqueue-merge` in keys?

behinddwalls (approved):

PR #353, opened 2026-07-14
feat(speculate): write the speculation tree in shadow mode

behinddwalls:
### Why?

This is the first of four slices reproducing the tree-driven speculate/build/buildsignal rework as an incremental, always-shippable stack instead of one large commit. Each slice keeps the orchestrator working end to end so the e2e suite stays green throughout, and reviewers can follow the pipeline shift (direct publish -> tree -> prioritize -> build) one deliberate step at a time.

This slice only adds the speculation tree as bookkeeping: nothing downstream reads it yet, so the batch's forward step (when it builds, when it merges) is untouched — with one deliberate exception called out below: a liveness fix to the dependent wake-up that review of this slice surfaced.

### What?

speculate's Controller now takes an enumerator, path scorer, selector, and dependency-limit factory, and every Created/Scored/Speculating pass loads or creates the batch's entity.SpeculationTree, applies the scorer's and selector's outputs, and persists it only if something changed (the apply steps report whether they mutated anything, gating a version+1 conditional Update). The controller stays pure mechanics: which paths exist, how they score, and which are promoted or cancelled are the seams' decisions alone — the controller validates and records them. Tree creation is gated by the queue's dependency limit; the controller wraps the enumerator's structure-only paths into persisted entries itself — stamping Candidate, minting each path's immutable ID as `{batchID}/path/{i}`, and skipping structural duplicates as a contract violation — and a concurrent create race re-reads the winner's tree instead of erroring. Seam outputs are consumed by path ID: scores merged by ID (unknown IDs skipped, out-of-[0,1] values clamped, both logged), selector decisions resolved through the ID-keyed PathIndex with duplicates logged and skipped.

The pre-existing forward step is otherwise unchanged: Created/Scored batches CAS to Speculating and publish straight to build, and Speculating batches run the original tryFinalize (merge once every dependency has landed, cascade-fail on a failed dependency). Cancelling is the unmodified pre-existing flow.

Liveness fix in the forward step: a batch waiting on dependencies (in tryFinalize, or now in the dependency gate) was only ever woken when a dependency was cancelled — a dependency reaching Succeeded or Failed never re-published its dependents, because mergesignal routes every terminal transition back through speculate under the batch's own ID and the terminal branch only re-published conclude. The terminal branch now fans out to dependents for every terminal state, and failOnDependency wakes its own dependents right after the terminal CAS so failures cascade downstream. The fan-out publishes a wake for every listed dependent; a dangling reverse-index entry (left by an abandoned batch creation) surfaces as a not-found speculate message that dead-letters and is skipped by the DLQ reconciler — eliminating that class at the source, by creating the batch before the reverse-index update, is tracked in #354.

Cancel decisions on tree paths are recorded as status only (Cancelling on an in-flight build) — nothing in this controller calls a build runner, which is why it takes no buildrunner.Factory.

main.go wires the new seams with parity defaults: chain enumerates a single path per batch, the probability path scorer, the all-selector promotes every candidate, and a static dependency limit that is effectively ungated.

## Test Plan
`bazel test //submitqueue/... //service/...` — 56/56 targets pass.

New unit coverage for the wake-up fix: dependent fan-out on every terminal state (with publish-order assertions), missing BatchDependent row surfaced as an invariant error, and dependent-publish failure after the terminal CAS nacking for redelivery convergence.

`make gazelle && make fmt` — no diffs beyond the intended BUILD file updates.

## Issue

Part of the speculation rework. Interim fan-out caveats tracked in #354.

## Stack
1. @ #353
1. #365
1. #371
1. #379
1. #380
1. #381
1. #382


submitqueue/orchestrator/controller/speculate/speculate.go
```diff
@@ -114,17 +145,17 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) (r
...
+	// dependency gate / tryFinalize against the new dependency state. On
+	// redelivery the same branch doubles as the crash self-heal: both the
+	// dependent fan-out and the conclude publish are idempotent re-sends.
 	if batch.State.IsTerminal() {
 		metrics.NamedCounter(c.metricsScope, opName, "self_heal_terminal", 1)
-		if batch.State == entity.BatchStateCancelled {
-			if err := c.respeculateDependents(ctx, batch); err != nil {
-				return err
-			}
+		if err := c.respeculateDependents(ctx, batch); err != nil {
```
albertywu, line 157: **High risk: this terminal wake can be silently deduplicated.** `respeculateDependents` ultimately publishes each dependent with `depID` as the queue message ID. The MySQL queue deduplicates `(topic, partition_key, id)` against consumed rows until garbage collection. For example: an older speculate row for B is consumed but not yet collected; A becomes terminal; this fan-out publishes `wake(B)` with ID B; the insert succeeds as a duplicate no-op. B never re-evaluates A's terminal state and can remain stuck. Keep B in the payload, but derive the event ID from the source batch, source version or terminal transition, and B. This remains present through PR #380.

submitqueue/orchestrator/controller/speculate/speculate.go
```diff
@@ -135,19 +166,344 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) (r
...
+	if err != nil {
+		metrics.NamedCounter(c.metricsScope, opName, "selector_errors", 1)
+		return fmt.Errorf("failed to select speculation decisions for batch %s: %w", batch.ID, err)
+	}
+
+	tree, selectionChanged := c.applySelection(batch, tree, decisions)
+
+	if scoresChanged || selectionChanged {
+		newVersion := tree.Version + 1
+		if err := c.store.GetSpeculationTreeStore().Update(ctx, batch.ID, tree.Version, newVersion, tree.Paths); err != nil {
```
albertywu, line 245: **High risk: expected tree CAS conflicts are treated as poison messages.** Two speculate deliveries can read the same tree version and both compute valid updates. The loser returns wrapped `storage.ErrVersionMismatch` here, but the production error processor does not classify that sentinel as retryable. The message can therefore enter DLQ, where reconciliation may change a healthy non-terminal batch to `Failed`. An optimistic conflict should reload, recompute, and retry rather than trigger terminal recovery. This remains present through PR #380 and affects additional CAS paths added later in the stack.

behinddwalls:

PR #355, opened 2026-07-14
chore(tests): clean up "passed" logging in integration tests

mnoah1:
## Why?
Bazel already reports pass/fail outcome on test cases - logging that a test case passed is redundant

## What?
Remove logging at the end of tests that just log "passed"

## Test Plan
- CI run


sbalabanov (approved):
e2e/integration test logger also prints time elapsed since the previous run of the logger, so sometimes could be useful for debugging
probably not in these cases though

PR #358, opened 2026-07-14
feat(stovepipe): derive build strategy in process step

mnoah1:
## What?
- Add source-control-backed build-strategy selection to Stovepipe’s process stage.
- Pin admitted requests to an incremental last-green baseline when ancestry is valid; otherwise use a full build.
- Fall back safely to full builds when the baseline is unavailable, and return unexpected source-control errors for normal consumer handling.
- Add coverage for cold starts, rewrites, unavailable ancestry, failures, and queue CAS reloads.

## Why?
Implementing proposed strategy selection from https://github.com/uber/submitqueue/blob/main/doc/rfc/stovepipe/steps/process.md

## Test Plan

Incremental:
- Seed an existing queue value in the DB matching the fake source control head
- call uber.submitqueue.stovepipe.Stovepipe/Ingest
- Check request in DB:
```
state:          processing
build_strategy: incremental_since_green
base_uri:       git://manual/incremental/HEAD
```

New Request:
- Empty/fresh DB
- call uber.submitqueue.stovepipe.Stovepipe/Ingest
- Check request in DB
```
state:          processing
build_strategy: full
base_uri:      <blank>
```

Not an Ancestor
- Seed an existing queue value in the DB that does not match the fake source control (so ancestor check returns false)
- Empty/fresh DB
- call uber.submitqueue.stovepipe.Stovepipe/Ingest
- Check request in DB
```
state:          processing
build_strategy: full
base_uri:      <blank>
```
- Planning some more integration tests around this flow


behinddwalls (commented):
A few non-blocking nits from review — the core derive→claim→persist ordering looks right (strategy is derived against the same queue-row version the claim CAS validates, so the persisted baseline can't be stale), and threading strategy/baseURI through `markProcessing` also fixes the latent reload-clobber on main. Nits inline.

stovepipe/controller/process/process.go
```diff
@@ -186,7 +195,21 @@ func (c *Controller) admitLatestHead(ctx context.Context, request entity.Request
 			return nil
 		}
 
-		err := c.claimBuildSlot(ctx, &queueRow)
+		if queueRow.LastGreenURI != "" && sc == nil {
+			var err error
```
behinddwalls: **Nit:** this inner `var err error` shadows the function-scoped `err` from line 184 for no behavioral difference — the outer `err` is dead here and reassigned at the `deriveBuildStrategy` call either way. Dropping the declaration removes the shadow and the risk of a future edit in this block reading the wrong `err`.

stovepipe/controller/process/process.go
```diff
@@ -220,15 +239,49 @@ func (c *Controller) admitLatestHead(ctx context.Context, request entity.Request
...
+				metrics.NewTag("reason", "unknown_ancestry"),
+			)
+			c.logger.Warnw("last-green URI is not in request history; using full build",
+				"queue", request.Queue,
+				"last_green_uri", queueRow.LastGreenURI,
+				"request_uri", request.URI,
+			)
+			return entity.BuildStrategyFull, "", nil
+		}
+		metrics.NamedCounter(c.metricsScope, _opName, "source_control_errors", 1)
```
behinddwalls: **Nit (observability):** `source_control_errors` is incremented both here (ancestry RPC failed) and at the factory-resolution failure above (line 202). Those are different pages — backend flaking vs. wiring/config broken — but the counter can't distinguish them, while the sibling `strategy_fallbacks` does carry a `reason` tag. Consider a `stage=resolve|ancestry` tag (same pattern as buildsignal's tagged counters) so the two failure modes alert separately.

stovepipe/controller/process/process_test.go
```diff
@@ -103,6 +121,195 @@ func expectAdmit(m processMocks, id string) {
...
+	expectAdmit(m, testID)
+
+	require.NoError(t, c.Process(context.Background(), delivery(t, ctrl, processPayload(t, testID))))
+
+	counter, ok := scope.Snapshot().Counters()["test.process_controller.process.admitted+strategy=full"]
+	require.True(t, ok)
+	assert.Equal(t, int64(1), counter.Value())
+}
+
+func TestProcessRederivesStrategyAfterQueueReload(t *testing.T) {
```
behinddwalls: **Nit (coverage):** this covers set→set `LastGreenURI` across the CAS reload with the factory memoized, but not the empty→set transition — iteration 1 skips factory resolution (`LastGreenURI == ""`), the reload sets `LastGreenURI`, and iteration 2 must resolve the factory late. That's exactly the branch the `sc == nil` guard exists for; a second scenario (initial queue without `LastGreenURI`, reloaded queue with it, expecting one `For` + one `IsAncestor` on the reloaded value) would pin it.

Relatedly (free assertion): the existing gate-closed fixtures use queues with empty `LastGreenURI`, so nothing demonstrates that a deferred delivery makes no source-control calls. Setting `LastGreenURI` on one of those fixtures gets that guarantee for free from strict gomock — any unexpected `For`/`IsAncestor` call fails the test — pinning the gate-before-derive ordering.

stovepipe/controller/process/process_test.go
```diff
@@ -230,6 +453,32 @@ func TestProcess(t *testing.T) {
 				m.reqStore.EXPECT().Update(gomock.Any(), superseded, int32(1), int32(2)).Return(nil)
 			},
 		},
+		{
+			name: "mark processing retry preserves derived strategy after accepted reload",
```
behinddwalls, line 503: **Nit (coverage):** this proves `BuildStrategy` survives the reload-clobber (reloaded row has `""`, retry expects `"full"`), but `BaseURI` is empty throughout, so its reapplication after reload is only exercised with the zero value. Using an incremental strategy + non-empty base URI here would prove both fields are reapplied.

behinddwalls (approved):

PR #365, opened 2026-07-14
feat(speculation): route builds through the tree via prioritize

behinddwalls:
### Why?

The speculate controller previously drove a batch's own build directly (publish to build, CAS to Speculating) while the speculation tree it maintained was pure shadow-mode bookkeeping nobody read. To make the tree the actual source of truth for what gets built, the pipeline needs to flip: speculate hands off to prioritize, and prioritize's admission decisions (Prioritized paths) are what the build stage acts on.

This also completes the ownership rule this stack is building toward: the build stage becomes the sole owner of runner interaction for path-level work. Other stages (prioritize's preemptive eviction, speculate's reconcile for dead paths, both landing in a later commit) only ever record a cancel decision as intent (SpeculationPathStatusCancelling) on the tree; nothing but the build stage calls BuildRunner.Cancel.

### What?

speculate.go: speculateBatch no longer publishes to build. It CASes Created/Scored batches to Speculating, then publishes the queue to prioritize on every pass (after the CAS, so a concurrent prioritize round never observes a batch as not-yet-Speculating with no later wake-up). The legacy tryFinalize step is otherwise unchanged and still only runs once a batch has reached Speculating on this or a prior pass, so merge timing is unchanged for now — tightening it to gate on the tree's own path outcome is deferred to a later commit in this stack.

build.go is replaced with its tree-driven form: it loads the batch's speculation tree and, per path, either triggers a build for a Prioritized path with no build yet, or enacts a persisted Cancelling intent. Trigger dedup is on the path->build mapping (SpeculationPathBuildStore), checked before Trigger runs since it is readable up front unlike a Build row. The write order is Trigger -> Build.Create -> mapping.Create -> publish to buildsignal; a crash between Trigger and the mapping write is recovered by redelivery re-running the same path, and a lost mapping-create race defers to the winner (republishing buildsignal for its build) rather than erroring. A Prioritized path whose mapping already resolves to a non-terminal build republishes buildsignal to close the crash window between the original Create and its original publish; a terminal build is a no-op, since reconcile is what will fold that outcome into the path's status.

The new Cancelling arm resolves the path's build via the same mapping, calls runner.Cancel if the build is not already terminal, and republishes buildsignal so the poll loop reacts promptly. Every lookup miss (no mapping yet, or a mapping pointing at a missing build row) is tolerated as a no-op rather than an error, leaving the intent for a later reconcile pass to settle. A missing speculation tree, by contrast, is an invariant violation — the tree is created before the batch is ever published to prioritize and is never deleted — so it surfaces as a plain error and dead-letters, failing the batch loudly instead of retrying against broken state.

buildsignal.go picks up its RFC-final wording (build's ID doubling as the runner handle) plus a corrected error-classification doc paragraph — classification is per error cause via the wired classifier walk, not the per-call-site policy the old text described — with no behavior change; the test file is byte-identical to before.

## Test Plan
✅ `go build ./submitqueue/... ./service/...`
✅ `go vet ./submitqueue/... ./service/...`
✅ `bazel test //submitqueue/... //service/...` — 56/56 targets pass, including new build.go tests for the Cancelling arm (non-terminal enacts + republishes, terminal no-ops, no-mapping/dangling-mapping skip, runner error surfaces with no silent ack).
✅ `make gazelle` / `make fmt` — no diffs beyond the intended source changes.
✅ `make e2e-test` — 2/2 targets pass (stovepipe, submitqueue), confirming parity: prioritize still admits the single chain path and exactly one build runs per batch.

## Stack
1. #353
1. @ #365
1. #371
1. #379
1. #380
1. #381
1. #382

### DIFF
--- a/submitqueue/orchestrator/controller/build/build.go
+++ b/submitqueue/orchestrator/controller/build/build.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"time"
 
 	"github.com/uber-go/tally"
 	entityqueue "github.com/uber/submitqueue/platform/base/messagequeue"
@@ -31,7 +32,27 @@ import (
 )
 
 // Controller handles build queue messages.
-// It consumes batches, triggers builds, and publishes scheduled builds to the build signal stage (which processes build results).
+//
+// It loads the batch's speculation tree and, for each path, acts on the
+// path's status: a Prioritized path that has no build yet gets one triggered
+// through the queue's build runner, and a Cancelling path has its in-flight
+// build cancelled. Every other status is left untouched, and the tree is
+// never written here — path statuses are read-only inputs to this
+// controller. It is the sole caller of the build runner for path-level work,
+// which is why persisted Cancelling intents are enacted here rather than
+// where they were decided.
+//
+// Halted (terminal or Cancelling) batches are skipped before any path work:
+// batch-level cancellation has its own owner, and the skip guarantees this
+// controller never starts CI for a batch that is being torn down.
+//
+// Dedup for triggering is on the path->build mapping (PathBuildStore), not
+// on a Build row keyed by a derived key: the mapping is readable before
+// Trigger ever runs, so it is checked first. A crash between Trigger
+// succeeding and the mapping Create means redelivery re-triggers a fresh
+// build for the same path; the new mapping Create then races the
+// (never-persisted) old one and simply wins, orphaning the earlier CI
+// build. See the per-path loop below for the full crash-safety ordering.
 // Implements consumer.Controller interface for integration with the consumer.
 type Controller struct {
 	logger        *zap.SugaredLogger
@@ -45,6 +66,9 @@ type Controller struct {
 
 // Verify Controller implements consumer.Controller interface at compile time.
 var _ consumer.Controller = (*Controller)(nil)
+
+// opName is the metric operation name shared by every emit in this file.
+const opName = "process"
 
 // NewController creates a new build controller for the orchestrator.
 func NewController(
@@ -68,11 +92,12 @@ func NewController(
 }
 
 // Process processes a build delivery from the queue.
-// Deserializes the batch, triggers a build, and publishes a build entity to the build signal topic.
-// Returns nil to ack (success), or error to nack (retry).
+// Deserializes the batch, loads its speculation tree, and for every path
+// either triggers a build (Prioritized, no build yet) or enacts a persisted
+// cancel intent (Cancelling), publishing to the build signal topic as
+// appropriate.
+// Returns nil to ack (success), or error to nack/reject.
 func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) (retErr error) {
-	const opName = "process"
-
 	op := metrics.Begin(c.metricsScope, opName)
 	defer func() { op.Complete(retErr) }()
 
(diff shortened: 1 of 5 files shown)


submitqueue/orchestrator/controller/speculate/speculate.go
```diff
@@ -249,16 +250,38 @@ func (c *Controller) speculateBatch(ctx context.Context, batch entity.Batch) err
...
+	if err := c.publishQueue(ctx, topickey.TopicKeyPrioritize, batch.Queue); err != nil {
+		metrics.NamedCounter(c.metricsScope, opName, "publish_errors", 1)
+		return fmt.Errorf("failed to publish queue %s to prioritize: %w", batch.Queue, err)
+	}
+
+	// Legacy forward path: tryFinalize only runs once a batch has already
+	// reached Speculating on a prior pass — merge timing is unchanged from
+	// before prioritize/build existed. This tightens in a later commit to
+	// gate on the tree's own path outcome instead of dependency states alone.
+	if originalState == entity.BatchStateSpeculating {
```
albertywu, line 281: **High risk in this PR slice: redelivery can finalize before any build starts.** Suppose the first pass persists `Speculating` and then crashes or errors while publishing prioritize. Redelivery starts with `originalState == Speculating`, republishes prioritize, and immediately enters `tryFinalize`. A dependency-free batch can move to `Merging` before prioritize starts a build; prioritize then skips it because it is no longer `Speculating`. This is fixed by PR #371, which requires a passed speculation path before finalization.

submitqueue/orchestrator/controller/speculate/speculate.go
```diff
@@ -808,6 +806,35 @@ func (c *Controller) publish(ctx context.Context, key consumer.TopicKey, batchID
...
+// queue-scoped stages (prioritize) whose message payload carries no batch
+// identity at all — just the queue to re-evaluate.
+func (c *Controller) publishQueue(ctx context.Context, key consumer.TopicKey, queue string) error {
+	qid := entity.QueueID{Name: queue}
+	payload, err := qid.ToBytes()
+	if err != nil {
+		return fmt.Errorf("failed to serialize queue ID: %w", err)
+	}
+
+	msg := entityqueue.NewMessage(queue, payload, queue, nil)
```
albertywu, line 819: **High risk: distinct prioritize rounds share one persistent dedup key.** Every wake for queue `q` uses `(topic=prioritize, partition=q, id=q)`. If an older consumed row has not been garbage-collected, a later batch can publish successfully here without creating a delivery. Its selected paths may then never be reconsidered for capacity. Use a logical event ID that is stable for retrying one round but distinct for later rounds, such as source batch plus version. This remains present through PR #380.

submitqueue/orchestrator/controller/build/build.go
```diff
@@ -115,58 +140,316 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) (r
...
+		// Persist the path->build mapping. This is the dedup marker for the
+		// path, checked at the top of this loop before Trigger runs.
+		mapping := entity.SpeculationPathBuild{
+			PathID:    p.ID,
+			BuildID:   build.ID,
+			BatchID:   batch.ID,
+			Version:   1,
+			CreatedAt: time.Now().UnixMilli(),
+		}
+		if err := c.store.GetSpeculationPathBuildStore().Create(ctx, mapping); err != nil {
```
albertywu, line 304: **High risk: the mapping race deliberately abandons a live external build.** Two deliveries can both observe no mapping and both call `runner.Trigger` before reaching this create. Only one mapping wins. The loser has already started an external build, but that build is not mapped to the path, polled, cancelled, or counted by the queue-wide budget. This can leak CI work and exceed configured capacity. Persist a dispatch reservation before the external call and pass a stable idempotency key to the runner; cancelling a known race loser is useful defense in depth. This remains present through PR #380.

submitqueue/orchestrator/controller/build/build.go
```diff
@@ -115,58 +140,316 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) (r
...
+						"path_id", p.ID,
+						"build_id", pb.BuildID,
+					)
+					continue
+				}
+				metrics.NamedCounter(c.metricsScope, opName, "storage_errors", 1)
+				return fmt.Errorf("failed to get build %s for path %s of batch %s: %w", pb.BuildID, p.ID, batch.ID, err)
+			}
+
+			if b.Status.IsTerminal() {
```
albertywu, line 220: **High risk if PR #365 is merged independently: completed builds retain capacity.** This branch assumes a later reconcile pass will fold the terminal build into `SpeculationPathInfo`, but this PR has no production writer that performs that reconciliation. The path can remain `Prioritized` or `Building`, so sticky continues counting its slot indefinitely and new work can starve. This structural omission is fixed by PR #371. The later stack still depends on buildsignal observing and persisting the terminal runner status, so the inherited polling dedup issue can still delay that observation.

behinddwalls:

PR #370, opened 2026-07-14
feat(stovepipe): add BuildStore extension with MySQL implementation

roychying:
## Why?
The stovepipe build pipeline step needs a persistence layer for the Build entity added in #362. Following the immutable-entity + optimistic-locking storage pattern used elsewhere in the codebase (RequestStore, QueueStore), builds need their own store: build creates a row once (at Trigger), and buildsignal is the sole writer of Status/Version afterward. No reverse index from Request to its builds is needed since buildsignal and record always reach a build by the ID carried on their queue messages.

## What?
- Adds the storage.BuildStore interface (stovepipe/extension/storage/build_store.go) with Create, Get, and a version-guarded Update (caller owns version arithmetic; the store performs a pure conditional write on oldVersion/newVersion).
- Adds the MySQL implementation (stovepipe/extension/storage/mysql/build_store.go) plus the build table schema (id, request_id, uri, base_uri, status, version), keyed by the runner-assigned build ID.
- Wires BuildStore into storage.Storage / mysqlStorage via a new GetBuildStore() accessor.
- Regenerates mocks for the new BuildStore interface (mockgen-generated, stovepipe/extension/storage/mock/).
- Adds a shared BuildStoreContractSuite (test/integration/stovepipe/extension/storage/suite.go) covering create/get, duplicate-create (ErrAlreadyExists), not-found (ErrNotFound), and CAS update (ErrVersionMismatch) semantics, and runs it against the MySQL backend (test/integration/.../mysql/storage_test.go).

## Test Plan
 - make test — unit tests pass
 - make integration-test — attempted; currently blocked in this dev environment by a pre-existing, unrelated infra issue (local docker-compose is v1.29.2, which silently no-ops on the --wait flag our test harness passes, so no containers ever start). Confirmed this is environmental, not a regression from this change: all 8 Docker-based integration suites fail identically the same way, including ones untouched by this diff. The new TestMySQLBuildStore suite builds and its logic was verified by reading through the assertions against the BuildStoreContractSuite (create/get, duplicate-create, not-found, CAS update).


stovepipe/extension/storage/mysql/schema/build.sql
```diff
@@ -0,0 +1,11 @@
+-- build holds one CI build triggered for a request's commit. id is the runner-assigned build id
+-- minted at Trigger (e.g. a Buildkite build number), opaque and never parsed or derived.
+CREATE TABLE IF NOT EXISTS build (
+    id          VARCHAR(255) NOT NULL,
+    request_id  VARCHAR(255) NOT NULL,
+    uri         VARCHAR(255) NOT NULL,
+    base_uri    VARCHAR(255) NOT NULL DEFAULT '',
```
behinddwalls: why do we need base and uri in build store?

stovepipe/extension/storage/mysql/build_store.go
```diff
@@ -0,0 +1,134 @@
+// Copyright (c) 2025 Uber Technologies, Inc.
```
behinddwalls, line 1: don't see tests for these?

behinddwalls (changes requested):

stovepipe/extension/storage/mysql/build_store.go
roychying: Not specific to `build_store.go`. No store file under `stovepipe/extension/storage/mysql/` or `submitqueue/extension/storage/mysql/` has unit tests. These are thin SQL wrappers; real behavior is already covered by the `BuildStoreContractSuite` integration tests against real MySQL. Happy to add the sqlmock unit tests but I would prefer to raise another pr for it, since it's not specific to `stovepipe/build`

stovepipe/extension/storage/mysql/schema/build.sql
roychying: Intent was for build to be a self-contained audit record (head/baseline visible without joining back to request), but it's true that request_id already gives us that via a lookup. Going to follow SubmitQueue's build table pattern here (no URI copy, resolve through the parent reference) and drop both columns and struct fields.

stovepipe/extension/storage/mysql/build_store.go
behinddwalls: sure.. okay we should have tests for them..feel free to create a separate PR

behinddwalls (approved):

Issue #369, opened 2026-07-14
pathDead: treat a Cancelled base dependency as dead once multi-path enumeration lands


behinddwalls:
## Context

`speculate.reconcile` marks a path dead (`pathDead` in `submitqueue/orchestrator/controller/speculate/speculate.go`) when its merge assumptions can no longer hold: a base dependency reached **Failed**, or a dependency outside the base **Succeeded** (the path bet it would not land first).

A **Cancelled base dependency deliberately does NOT dead the path** today (Phase-A leniency). This mirrors the pre-tree chain semantics: the cancelled batch will never land, so it can no longer conflict with the path's changes, and we accept that the path's build ran against a stale assumption set (it built on top of a base member that will now never merge ahead of it).

## What should change

Once multi-path enumeration lands (a batch's tree carries more than the single chain path), a Cancelled base dependency should mark the path dead the same way a Failed one does: a sibling path that never assumed the cancelled batch picks up the slack, and the batch merges on an accurate assumption set instead of a stale build.

This is gated on multi-path enumeration because today there is exactly one path per batch (chain enumerator) — deading the only path on a benign cancel would fail batches that currently survive via the chain semantics.

## Where

- `pathDead` in `submitqueue/orchestrator/controller/speculate/speculate.go` — treat a Cancelled base dependency as dead.
- Revisit the reconcile tests' Phase-A expectations at the same time.

PR #371, opened 2026-07-14
feat(speculation): reconcile dead paths and finalize from the tree

behinddwalls:
### Why?

Two speculation.go behaviors were still driven off the pre-tree dependency-only model from earlier commits in this stack: dependency outcomes were folded into the batch via a bespoke tryFinalize/failOnDependency pair that only looked at batch dependency state and never at the tree, and per-path build outcomes were never reconciled back into the tree at all. A build finishing, or a dependency dying mid-flight, left stale Candidate/Selected/Building statuses that scoring and selection had to route around indirectly. This split made the tree an incomplete picture of a batch's speculation state and kept the merge decision in logic with no visibility into per-path build results.

### What?

speculateBatch now reconciles the tree against reality on every pass, before scoring. reconcile() folds each path's build outcome (via the path->build mapping and build store) into its status through reconcileStatus(), then walks the tree a second time to dead any path whose dependency chain can no longer land. Paths already at a terminal status (Passed/Failed/Cancelled) are settled and skipped without touching storage, and statuses are folded into the tree's own path slice in place (mirroring applyScores/applySelection — no per-pass copy), so a pass costs at most two point reads per unresolved path and stays O(paths) in reads and CPU while allocating nothing. The scaling rules for paths and trees — the enumerator implementation decides which and how many paths a batch gets, bounded frontier, never the 2^N power set of dependency subsets, O(paths) per pass — are captured in a new "speculation paths and trees" section in CLAUDE.md so future changes account for them.

reconcileStatus() is pure status arithmetic with two rules its doc now spells out: a terminal path status is never downgraded, and Cancelling records a decision about the path (deselected, or assumptions dead), not a prediction about the build — so a Cancelling path holds until its build reaches any terminal state and then settles to Cancelled, including a build that raced to Succeeded before the intent was enacted (its result is moot for a path that will never merge; the build stage never cancels an already-terminal build).

pathDead() treats a Failed base dependency, or a Succeeded non-base dependency, as fatal to a path; a Cancelled base dependency is deliberately tolerated (Phase-A leniency — a cancelled batch never lands and so cannot conflict, and with exactly one chain path per batch today, deading it would fail batches that currently survive). Tightening that case once multi-path enumeration lands is tracked in https://github.com/uber/submitqueue/issues/369 and linked from the code.

A dead path that is still Building is captured as a cancel intent (Cancelling status) with no runner call. Actually cancelling the in-flight build is the build stage's job (build.go's enactCancel, added earlier in this stack), reached one hop later through the prioritize round speculate already publishes on every pass — prioritize's republishBuilds re-triggers the build stage for any tree with a Cancelling path. A dead path with no build yet drops straight to Cancelled.

finalize()/mergeableNow()/viable() replace tryFinalize()/failOnDependency(). finalize settles the batch's forward step in exactly one of three outcomes per pass, which its doc now enumerates: merge now (some path is mergeableNow — publish to merge, CAS to Merging), wait (no path mergeable but at least one still viable — no-op until the next event), or fail (no viable path remains — CAS to Failed, fan out to dependents, publish to conclude). mergeableNow checks the path's own status (must be Passed) in addition to its dependencies; viable distinguishes "still might merge" from "definitely won't" (Failed, Cancelled, and Cancelling paths are not viable). finalize runs unconditionally on every pass — it no longer needs the batch's original-state gate the previous commit introduced as a stopgap, since mergeableNow correctly requires a path to have actually passed before merging. The dependent fan-out on the fail outcome is publish-only — one speculate wake-up per dependent, each processed asynchronously on its own delivery — now stated explicitly at the call site.

Process's terminal self-heal now re-fans dependents for Failed batches too, not just Cancelled ones: a batch failed by finalize (no viable path left) never passes through mergesignal, so speculate's redelivery self-heal is the only stage that can re-notify its dependents if the fan-out publish was ever lost.

The buildRunners dependency and the tree-driven cancelBatch/cancelBuilds/cancelTree replacement are deliberately left out of this commit — reconcile() here never talks to a build runner by design. That lands in the next commit in the stack, which is also the one place a direct runner cancel remains justified (batch-level cancel, not path-level).

## Test Plan
✅ `go build ./...` and `go vet ./...` — clean (pre-existing vet warnings in mergeconflictsignal/mergesignal test files are unrelated to this change).
✅ `bazel test //submitqueue/... //service/...` — 56/56 targets pass; new reconcile/dead-path/finalize/pure-function tests pass alongside the preserved legacy cancel tests, and the merge-gate tests pin that terminal paths cost no path->build reads.
✅ `make gazelle && make fmt` — no changes.
✅ `make e2e-test` — stovepipe and submitqueue e2e suites both pass.

## Issue

Follow-up tracked in https://github.com/uber/submitqueue/issues/369 (tighten pathDead's Cancelled-base-dependency leniency once multi-path enumeration lands).

## Stack
1. #353
1. #365
1. @ #371
1. #379
1. #380
1. #381
1. #382

### DIFF
--- a/submitqueue/orchestrator/controller/speculate/speculate.go
+++ b/submitqueue/orchestrator/controller/speculate/speculate.go
@@ -35,7 +35,8 @@ import (
 
 // Controller handles speculate queue messages.
 //
-// Each invocation reconciles the batch's entity.SpeculationTree, publishes
+// Each invocation reconciles the batch's entity.SpeculationTree — against
+// the build stage's outcomes and the current dependency states — publishes
 // the batch's queue to prioritize, and advances the batch one step in the
 // state machine. Tree reconciliation is pure mechanics: the controller runs
 // the queue's configured seams — enumerator (path structure), path scorer
@@ -43,17 +44,18 @@ import (
 // applies them to the persisted tree, and writes it back under optimistic
 // concurrency. Which paths exist, how they score, and which are promoted or
 // cancelled are the seams' decisions alone; the controller never originates
-// one. The batch's own merge timing is still driven by dependency states
-// (tryFinalize), pending a later change that gates it on the tree's path
-// outcomes.
+// one. Finalization is driven by the tree itself: publish to merge as soon
+// as any path is actually mergeable, or fail the batch once no path can
+// ever merge again.
 //
 // Per invocation, the controller advances the batch one step in the state
 // machine:
 //
 //   - Created, Scored, or Speculating → speculateBatch: reconcile the tree
-//     (created on the first pass the dependency gate admits), CAS to
-//     Speculating for Created/Scored, publish to prioritize every pass, and
-//     run tryFinalize once the batch has reached Speculating.
+//     (created on the first pass the dependency gate admits) against build
+//     outcomes and dead dependencies, CAS to Speculating for Created/Scored,
+//     publish to prioritize every pass, and finalize — merge a mergeable
+//     path, fail the batch if none remain viable, or wait.
 //   - Cancelling        → cancel any in-flight Build entity, respeculate
 //     dependents, CAS to terminal Cancelled, publish to conclude. The
 //     cancel controller hands the batch off in this state and speculate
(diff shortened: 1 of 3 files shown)


submitqueue/orchestrator/controller/speculate/speculate.go
```diff
@@ -442,6 +448,195 @@ func (c *Controller) applyScores(batch entity.Batch, tree entity.SpeculationTree
...
+	for i, p := range paths {
+		if isTerminalPathStatus(p.Status) || p.Status == entity.SpeculationPathStatusCancelling {
+			continue
+		}
+		if !pathDead(p.Path, depByID) {
+			continue
+		}
+
+		if p.Status == entity.SpeculationPathStatusBuilding {
+			p.Status = entity.SpeculationPathStatusCancelling
```
albertywu, line 533: **High risk: this persisted cancellation intent can become unreachable before `runner.Cancel` runs.** Reconciliation changes a dead `Building` path to `Cancelling`, but same-pass finalization treats that path as nonviable and can move the batch to terminal `Failed`. Once terminal, prioritize no longer loads the batch and build skips it, so the cancellation is never enacted and the external build can continue. The external leak existed on this PR's base, but this changed flow leaves it unclosed and adds a persisted cancel intent that the same pass can strand. Do not terminalize while any path is `Cancelling`; route cancellation and wait for terminal runner observation. This remains present through PR #380.

behinddwalls:

PR #374, opened 2026-07-15
docs(rfc): consumer gate — runtime stop/start of individual queue controllers

sbalabanov:
This RFC proposes a first-class **stop/observe/start** primitive for individual queue controllers: halt exactly one controller from taking new messages, observe some condition while it is stopped, then start it again — without stopping the service that hosts it. Motivations are deterministic e2e scenario control (interleaving in-flight messages across controllers) and operational pause of a consuming stage during incidents.

The mechanism is **consumer middleware**: a decorator installed around every registered controller consults gate state before invoking `Process`. When the gate is closed, the delivery is parked in-flight — the decorator blocks and periodically extends the visibility timeout, so parked messages never burn retry budget, never hit the DLQ, and are redelivered normally if the process dies. When the gate opens, the delivery proceeds as the same attempt, in partition order.

Gate state lives in a new shared extension at `platform/extension/consumergate`, with a **file-based first implementation** usable by both e2e tests and operators. This is a design doc only — no code changes.

### DIFF
--- a/doc/rfc/consumer-gate.md
+++ b/doc/rfc/consumer-gate.md
@@ -1,0 +1,89 @@
+# Consumer Gate
+
+Stopping and starting individual queue controllers at runtime — for deterministic e2e scenario control and for operational pause of a consuming stage — without stopping the service that hosts them.
+
+## Problem
+
+The pipeline is a set of queue controllers spread across three services, and several needs reduce to the same primitive: *stop one controller from taking new messages, observe some condition while it is stopped, then start it again.*
+
+- **E2e scenario control.** A test that must interleave two in-flight messages ("the batch controller must not consume its message before the cancel controller has finished") needs to halt exactly one controller while its siblings keep running. Stopping the whole service is too coarse: it kills every controller in the process, re-assigns ephemeral ports, and turns a scenario step into a container lifecycle event.
+- **Operational pause.** During an incident, the safest first move is often "stop consuming" — park a stage that is processing poison input or hammering a struggling dependency — without redeploying or scaling to zero, and while keeping the rest of the service (RPC surface, other controllers) alive.
+
+Nothing in the system expresses this today. The queue can be manipulated from outside (e.g. starving a consumer by occupying its partition leases), but data-plane tricks of that kind are structurally limited: they only work if arranged before the controller ever touches the partition, they couple the caller to one backend's internals, and they are invisible to the service — no first-class semantics, no metrics, nothing an operator can reuse. This RFC makes stop/observe/start a small, coherent, first-class mechanism instead.
+
+## Decisions
+
+### The gate is consumer middleware, acting on deliveries before the controller
+
+The consumer framework already owns the two facts that make an in-process gate clean. Dispatch is **serial per partition** — `consumeLoop` routes each delivery to a per-partition goroutine, and the next delivery of a partition is not started until the current one completes — so holding one delivery blocks exactly that partition and nothing else. And the framework **owns ack/nack** — controllers signal outcome only through `Process`'s return value — so a delivery can be held simply by not yet invoking the controller.
+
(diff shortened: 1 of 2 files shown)


doc/rfc/consumer-gate.md
```diff
@@ -0,0 +1,89 @@
...
+
+The gate is a decorator installed by the consumer around every registered controller. Before invoking `Process`, it consults gate state for the controller's consumer group. If the gate is closed, the delivery is **parked**: the decorator blocks in place, keeping the delivery in-flight and periodically calling `ExtendVisibilityTimeout` — already part of the `Delivery` contract, and specified to *not* increment the retry count — until the gate opens or the consumer shuts down. When the gate opens, the parked delivery proceeds into the controller as the same attempt, in partition order. Parked messages therefore never burn retry budget, never touch the DLQ, and are never lost: if the process dies while parked, extension stops, visibility lapses, and the queue redelivers normally.
+
+Stopping is a barrier, not preemption: a message already inside `Process` when the gate closes runs to completion; the gate guarantees no *new* message enters the controller.
```
behinddwalls, line 22: the one other i have noticed which is not testing from functional perspective we need, when we have limits on how much a stage can do and requires queue to be halted..but that signal is generally coming from Process itself on consumption of the message, like merge serialization...  so yeah, extending visibility but don't account to retry (explicit) would be great where we consume the message and return with some nack+visibilityTimeout, i thought about adding it when i added PublishAfter semantic for build status polling but as I am doing the merge/finalize stuff..seems like we are missing this construct.. so I am hoping we can use the same there as well

doc/rfc/consumer-gate.md
```diff
@@ -0,0 +1,89 @@
...
+
+One bounded side effect is accepted and documented: the routing loop feeds each partition through a channel buffered at the subscription's batch size, so if messages keep arriving for a parked partition, the topic's routing loop eventually stalls once that buffer fills. For a fully closed gate this is moot (every partition parks anyway), and at test volumes it never triggers.
```
behinddwalls, line 24: i think even for actual volume, it comes down to buffer size and other wise everything remain in DB i guess for later delivery.

doc/rfc/consumer-gate.md
```diff
@@ -0,0 +1,89 @@
...
+
+Keeping the contract separate from any backend is what lets the storage medium be chosen per deployment: a filesystem directory first (below), a database- or config-service-backed implementation later if fleet-wide coordination demands it — with the middleware, the wiring shape, and every test written against the contract unchanged.
+
+### First implementation: files in a shared directory
+
+The first implementation stores gate state as plain files under a configured directory. Presence of a gate file means the gate is closed; deleting the file opens it. Parked deliveries are recorded as JSON files. The layout:
+
+```
```
behinddwalls, line 40: if we want to use it for backoff business logic, we may need to store them in DB

behinddwalls (approved):

doc/rfc/consumer-gate.md
sbalabanov: Ai decided to bother it as a side effect, I wouldn't. Same principles apply to when the service is down and queue accumulates.

doc/rfc/consumer-gate.md
sbalabanov: Claude's first implementation was DB. Given closing gates in prod is extremely rare, and in test everything is on the same machine, I've decided files is faster and more control.
Prod gates may be implemented also through files, delivered as configuration. TBD.

doc/rfc/consumer-gate.md
```diff
@@ -0,0 +1,89 @@
...
+
+Tests do not depend on that latency. The deterministic patterns are two: **arrange first** (close the gate before publishing the message that must be caught — exact by construction), or **await the observed effect** (the parked record, below) instead of assuming timing.
+
+### Observation: parked deliveries are recorded
+
+Parking writes the parked record before blocking. This record is the "observe" half of stop/observe/start: a test awaits the record to *know* the stop caught its message (there is otherwise no signal distinguishing "gated and parked" from "not arrived yet"), can assert on the recorded payload, and can decide what to do next while the controller is provably stopped. For an operator, the same records answer "what is this paused controller holding?". Records are bounded by parked messages, which are bounded by gate usage; the directory is empty whenever no gate is in use.
```
albertywu, line 64: Can we specify how the RFC handles situations where there are multiple parked records (not just one).

For example let's say there are 10 buffered messages for controller X while it is in "gated" mode (paused). And now let's assume we let the system sit in that state for 1 hour.

Is there some mechanism to prevent each of those 10 buffered messages from expiring and entering DLQ?

PR #375, opened 2026-07-15
feat(consumergate): runtime stop/start of queue controllers

sbalabanov:
Implements the consumer-gate RFC (#374) and showcases a Request Cancellation e2e test using it.

## Summary

- Adds `platform/extension/consumergate` with the `Gate`, `Entry`, `Admin`, `Factory`, and configuration contracts, plus generated mocks.
- Adds file-backed and no-op implementations. The file backend is explicitly scoped to E2E tests and single-host development; it reads gate state directly from shared files for every delivery and uses atomic temp-file-plus-rename writes.
- Clears each queue delivery that reaches controller dispatch through its consumer-group/partition gate before processing. A delivery blocked at the gate remains in flight through visibility extensions without consuming retry attempts or breaking partition ordering; protection of other prefetched/buffered deliveries is outside this E2E-focused change.
- Fails open when gate reads, monitoring, or visibility extensions fail. Consumer shutdown instead leaves a blocked delivery unacked for normal redelivery.
- Keeps parked records and payload files only while a delivery is actively blocked.
- Uses polling as the portable convergence mechanism. Filesystem events may be added later as an optimization alongside polling because event delivery varies across operating systems, bind mounts, overlay/network filesystems, rootless Docker, and Docker Desktop.
- Documents that the file backend coordinates only processes sharing its directory, survives restarts only with surviving/shared storage, and is not a fleet-wide production pause control plane. Gating never acknowledges, nacks, rejects, removes, or moves the source queue delivery; queue ownership and normal redelivery semantics remain intact.
- Gateway, orchestrator, and runway select the file backend only when `CONSUMER_GATE_DIR` is explicitly configured; otherwise they use the no-op implementation. Compose opts in for E2E and local development. Stovepipe always uses no-op.
- Runs application containers under the host test UID/GID with rootful Docker, while using `0:0` under rootless Docker so bind-mounted gate files remain readable by the host test process.

## Deterministic cancellation E2E coverage

`TestCancel_CaughtPreBatch_NeverLands` closes the runway merge-conflict-check gate for the test queue before landing a request. The test observes the exact parked delivery, cancels the request to its terminal state, opens the gate, waits for the parked record to disappear, and uses a sentinel request to prove the stale message was consumed without reviving or landing the cancelled request.

## Testing

- `make test` — passed (76/76 Bazel test targets).
- `//test/integration/...` — passed (8/8 integration targets).
- `//test/e2e/...` — passed (2/2 E2E targets).
- `git diff --check` — passed.


platform/consumer/consumer.go
```diff
@@ -86,14 +91,21 @@ type activeSubscription struct {
 // consumers such as DLQ reconciliation that must redeliver on any failure.
 // processor must not be nil; callers that genuinely want no transformation
 // can pass errs.NewClassifierProcessor() with no classifiers.
-func New(logger *zap.SugaredLogger, scope tally.Scope, registry TopicRegistry, processor errs.ErrorProcessor) Consumer {
-	return &consumer{
+//
+// opts configure optional behavior, e.g. WithGate to install the consumer-gate
+// middleware.
+func New(logger *zap.SugaredLogger, scope tally.Scope, registry TopicRegistry, processor errs.ErrorProcessor, opts ...Option) Consumer {
```
sbalabanov: Pass Gate interface directly and not through options. Like other interface-typed arguments, it should never be null by convention and not need to check that. If no gates are required, the caller would wire up a no-op gate implementation. Please create one.

service/submitqueue/orchestrator/server/main.go
```diff
@@ -785,6 +794,20 @@ func getEnv(key, defaultVal string) string {
...
 
+// gateOptions returns the consumer options enabling the consumer gate when
+// CONSUMER_GATE_DIR is set: gate state is then read from that directory (see
+// platform/extension/consumergate/file), letting tests and operators stop and
+// start individual queue controllers at runtime. When unset, the returned
+// slice is empty and the gate middleware is absent.
+func gateOptions(logger *zap.Logger) []consumer.Option {
+	dir := os.Getenv("CONSUMER_GATE_DIR")
+	if dir == "" {
+		return nil
```
sbalabanov: prefer a default folder instead

platform/extension/consumergate/file/store.go
```diff
@@ -0,0 +1,266 @@
...
+		return fmt.Errorf("failed to create dir %s: %w", dir, err)
+	}
+	tmp, err := os.CreateTemp(dir, filepath.Base(path)+".tmp")
+	if err != nil {
+		return fmt.Errorf("failed to create temp file in %s: %w", dir, err)
+	}
+	tmpName := tmp.Name()
+	if _, err := tmp.Write(data); err != nil {
+		tmp.Close()
+		os.Remove(tmpName)
```
sbalabanov: do not swallow potential error, instead concatenate with errors.Join
also, redesign a code to not require to call os.Remove(tmpName) for every unsuccessful operation. Converge error flows and do it once.

platform/consumer/gate.go
```diff
@@ -0,0 +1,218 @@
...
+	gated bool
+	at    time.Time
+}
+
+// hold blocks while the controller's gate is closed for the delivery's
+// partition. It returns true when the delivery may proceed into the
+// controller, and false when the consumer is shutting down while parked — in
+// that case the caller must not process, ack, or nack: extension stops,
+// visibility lapses, and the queue redelivers normally.
+func (g *gateMiddleware) hold(ctx context.Context, controller Controller, delivery extqueue.Delivery) bool {
```
sbalabanov: Middleware always relies on periodic checks, however the potential implementation may be able to open the gate as soon as admin signal arrives if it supports notifications. Leave blocking the gate and unblocking to the consumer gate instead - you have to redesign the consumer gate interface to make it a part of the contract. The middleware in this case could be very thin or not needed itself, as Consumer can just check Gates open/close as part of its logic.
Please debate internally whether to implement recording part in the consumer or make it a part of the consumer gate interface responsibility. Make a judgement and implement the best variant.

platform/extension/consumergate/consumergate.go
```diff
@@ -0,0 +1,148 @@
...
+	// (immediately when the gate is open). Implementations own the wait
+	// mechanism (a polling implementation re-reads gate state on a timer; a
+	// notification-capable implementation can release the instant the gate
+	// opens) AND the parked-delivery observation records: an implementation
+	// records the delivery when it actually blocks and stamps its release
+	// when it returns.
+	//
+	// Returns ctx.Err() when ctx is cancelled while blocked (the delivery
+	// must not proceed). Any other error means gate state could not be read
+	// or recorded — the consumer fails open.
```
sbalabanov: do not comment how consumer operates the gate. Only comment the contract that Gate should satisfy.

platform/extension/consumergate/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Contract
+
+A gate is identified by a consumer group (every controller subscribes with a unique one, so it is the controller's stable runtime name), optionally narrowed to a single partition. The gate owns both the blocking mechanism and the parked-delivery observation records. While a gate is closed, its `Wait` method blocks the delivery, stamping `ParkedAtMs` when it actually blocks and `ReleasedAtMs` when the gate opens. The consumer owns queue mechanics (visibility extension via `keepInFlight`, keeping the delivery in-flight without burning retry budget) and the fail-open posture (a `Wait` error lets the delivery through). Stopping is a barrier, not preemption — a message already inside the controller runs to completion.
+
+The package defines two interfaces plus the `Config`:
+
+- `Gate` exposes a single `Wait(ctx, Parked) error` method. Polling implementations (see `file/`) re-check gate state on a timer; notification-capable implementations can release the instant the gate opens. The consumer passes `Gate` as a required argument to `consumer.New`; callers that do not need gating wire the `noop/` implementation.
```
sbalabanov: also here, do not comment how Gate is to be used by other packages, only the interface and contract it should satisfy. At least, stay abstract and do not mention concrete packages and types outside the extension.

platform/extension/consumergate/README.md
```diff
@@ -0,0 +1,23 @@
...
+- `Admin` is the write surface tests and tooling use: close a gate, open it, list what a stopped controller is holding.
+
+Parked records are the "observe" half of stop/observe/start: awaiting one is the only way to *know* a stop caught a specific message (as opposed to the message not having arrived yet), and the `ReleasedAtMs` stamp proves the delivery later proceeded.
+
+## Failure posture
+
+Gating is auxiliary: if `Wait` returns a non-ctx error, the consumer logs, counts the error, and lets the delivery through (fail open). A broken gate medium must not become a pipeline stall.
```
sbalabanov: same, failure posture is determined by the client. Gate should expose the error without considerations.

platform/consumer/gate.go
```diff
@@ -0,0 +1,125 @@
+// Copyright (c) 2026 Uber Technologies, Inc.
```
sbalabanov: move this file content (gate.go) to the consumer.go
having methods of consumer implemented in a different file is confusing
update test files accordingly

platform/consumer/gate.go
```diff
@@ -0,0 +1,125 @@
...
+		PartitionKey:  msg.PartitionKey,
+		Payload:       msg.Payload,
+		Attempt:       delivery.Attempt(),
+	}
+
+	stopKeepAlive := m.keepInFlight(ctx, delivery, keepInFlightInterval)
+	start := time.Now()
+	err := m.gate.Wait(ctx, parked)
+	stopKeepAlive()
+	metrics.NamedTimer(scope, opName, "wait_latency", time.Since(start))
```
sbalabanov: use histogram

platform/consumer/gate.go
```diff
@@ -0,0 +1,125 @@
...
+
+	if err == nil {
+		return true
+	}
+
+	if ctx.Err() != nil {
```
sbalabanov: check the actual error returned instead

platform/consumer/gate.go
```diff
@@ -0,0 +1,125 @@
...
+	parked := consumergate.Parked{
+		ConsumerGroup: controller.ConsumerGroup(),
+		Topic:         controller.TopicKey().String(),
+		MessageID:     msg.ID,
+		PartitionKey:  msg.PartitionKey,
+		Payload:       msg.Payload,
+		Attempt:       delivery.Attempt(),
+	}
+
+	stopKeepAlive := m.keepInFlight(ctx, delivery, keepInFlightInterval)
```
sbalabanov: this is expensive call with goroutine spawned, and it only needs to be called if gate is blocked
redesign to account for this. Change Gate interface if needed.

platform/consumer/gate.go
```diff
@@ -0,0 +1,125 @@
...
+		ticker := time.NewTicker(interval)
+		defer ticker.Stop()
+		for {
+			select {
+			case <-done:
+				return
+			case <-ctx.Done():
+				return
+			case <-ticker.C:
+				if err := delivery.ExtendVisibilityTimeout(ctx, parkExtensionMs); err != nil {
```
sbalabanov: this failure needs to propagate and gate to open immediately as a part of fail-open contract

platform/consumer/consumer.go
```diff
@@ -485,6 +517,127 @@ func (m *consumer) processDelivery(ctx context.Context, controller Controller, d
...
+	// The wait context is cancellable by the keep-in-flight timer: when a
+	// visibility extension fails, the delivery can no longer be held safely
+	// (the queue may already be re-delivering it), so the wait is cancelled
+	// with the extension failure as the cause.
+	waitCtx, cancelWait := context.WithCancelCause(ctx)
+	defer cancelWait(nil)
+
+	stopExtending := m.keepInFlight(waitCtx, delivery, cancelWait)
+	start := time.Now()
+	err := m.gate.Wait(waitCtx, parked)
```
sbalabanov: design Gate interface differently. Instead of Wait() function to block, have a function that returns a future-type of object and the customer can check synchronously if the gate was closed or not. If the gate is open, the customer proceeds normally. If the gate is closed, consumer runs all the delivery extension logic as it sees fits, including creation of "parked" record. After that it calls a method Wait() on a future object itself and it blocks until the gate to open.

platform/consumer/consumer.go
```diff
@@ -485,6 +517,146 @@ func (m *consumer) processDelivery(ctx context.Context, controller Controller, d
...
+// It is timer-driven rather than goroutine-driven: nothing runs unless the
+// delivery actually stays blocked past the first extend interval, so the
+// common case — an open gate — costs one timer arm/disarm and no goroutine.
+//
+// An extension failure means the delivery can no longer be held safely (its
+// visibility may lapse and the queue redeliver it), so the failure propagates
+// through cancelWait — unblocking the gate wait immediately — and the timer
+// chain stops. The returned stop function disarms the timer and waits out any
+// in-flight extension, guaranteeing no extension races the controller.
+func (m *consumer) keepInFlight(ctx context.Context, delivery extqueue.Delivery, cancelWait context.CancelCauseFunc) (stop func()) {
```
sbalabanov: redesign keepInFlight which lifecycle could be controlled by context.
algorithm:
- keepInFlight starts a goroutine that listens to both context and timer ticker. On each tick, it updates the visibility. It should have a channel back to main routine to signal errors.
- when gate unblocks, consumer cancels goroutine and joins its thread
- if keepInFlight fails to extend, it signals back to main thread which cancels the context for the Wait() and opens the gate with the error, recording it.
- if parent context is cancelled, consumer cancels both contexts, wait for them, then returns without acking.

platform/extension/consumergate/consumergate.go
```diff
@@ -0,0 +1,197 @@
...
+	Blocked() bool
+
+	// Watch records delivery as parked, adding the gate identity captured by
+	// Enter and the implementation-owned ParkedAtMs, and begins monitoring the
+	// gate. It returns a channel that yields exactly one value: nil when the gate
+	// opens, or a non-nil error if gate state could not be read or the record
+	// written — without further interpretation, as what to do with a failed wait
+	// is the caller's policy. If ctx is cancelled while monitoring, the channel
+	// yields ctx.Err(). The implementation removes the parked record before
+	// yielding on every terminal path, so ListParked contains only deliveries
```
behinddwalls, line 139: not sure if we need to document it, message gated is never removed from the queue for a given consumer and partition

platform/extension/consumergate/consumergate.go
```diff
@@ -0,0 +1,197 @@
...
+func Wait(ctx context.Context, entry Entry, descriptor DeliveryDescriptor) error {
+	if !entry.Blocked() {
+		return nil
+	}
+	return <-entry.Watch(ctx, descriptor)
+}
+
+// Admin is the write surface used by tests and tooling to operate gates and
+// inspect what a stopped controller is holding.
+type Admin interface {
```
behinddwalls, line 168: should we rename it to Manage? or Manager maybe

albertywu:
Could you clarify what happens in the situation where there's 10 buffered messages awaiting controller X and the controller is paused?

Codex is telling me that the first message would wait at the closed gate and have visibility continually extended but the trailing 9 would remain buffered without visibility extension. Every 60 seconds, those 9 expire and are redelivered, consuming retry attempts and creating duplicate buffered delivery. And after 3 attempts they move to DLQ.

Assuming this is true, wouldn't this cause the undesirable impact of dropping the first 9 buffered messages while controller X is in a prolonged paused state?

albertywu:
It seems that with this design that uses file-based marking of gated controllers, the gating would not survive node crashes.

For example, let's say:
1) Controller X is paused on Node A (so marker is left on Node A on the filesystem)
2) Node A crashes and is replaced with Node B with fresh filesystem
3) Controller X is unpaused now (pausing does not survive node crashes)

Similarly, if the fleet has 10 nodes, I don't think we fan-out the file write to all 10 nodes -- so the pause behavior would only impact the single node that the request landed on. Basically this design seems to work with a single-node only and assumes the filesystem is stateful and persistent across the lifetime of gate usage.

We could fix this by moving the gating implementation from file-based to a storage interface backed with a db. It seems to fit right in with existing Queue implementation which uses db storage so seems better to include as part of that vs a special-snowflake file-based implementation. Something like this could work as the core:
```
CREATE TABLE consumer_gates (
    consumer_group VARCHAR(255) NOT NULL,
    partition_key  VARCHAR(255) NOT NULL, -- empty means all partitions
    reason         TEXT NOT NULL,
    created_by     VARCHAR(255) NOT NULL,
    created_at_ms  BIGINT NOT NULL,
    PRIMARY KEY (consumer_group, partition_key)
);
```

sbalabanov:
> Could you clarify what happens in the situation where there's 10 buffered messages awaiting controller X and the controller is paused?
> 
> Codex is telling me that the first message would wait at the closed gate and have visibility continually extended but the trailing 9 would remain buffered without visibility extension. Every 60 seconds, those 9 expire and are redelivered, consuming retry attempts and creating duplicate buffered delivery. And after 3 attempts they move to DLQ.
> 
> Assuming this is true, wouldn't this cause the undesirable impact of dropping the first 9 buffered messages while controller X is in a prolonged paused state?

It depends on a delivery order. For synchronized queues (only one message could be processed at a time), other messages would be held and won't be extended. For parallel queues, if there is enough controller instances on a production or e2e test system, multiple messages would be parked and their life extended as a result of this.

It is true that messages not accepted will not be extended. For e2e tests, this is not a problem as we control the number of messages to be validated.

sbalabanov:
> It seems that with this design that uses file-based marking of gated controllers, the gating would not survive node crashes.
> 
> For example, let's say:
> 
> 1. Controller X is paused on Node A (so marker is left on Node A on the filesystem)
> 2. Node A crashes and is replaced with Node B with fresh filesystem
> 3. Controller X is unpaused now (pausing does not survive node crashes)
> 
> Similarly, if the fleet has 10 nodes, I don't think we fan-out the file write to all 10 nodes -- so the pause behavior would only impact the single node that the request landed on. Basically this design seems to work with a single-node only and assumes the filesystem is stateful and persistent across the lifetime of gate usage.
> 
> We could fix this by moving the gating implementation from file-based to a storage interface backed with a db. It seems to fit right in with existing Queue implementation which uses db storage so seems better to include as part of that vs a special-snowflake file-based implementation. Something like this could work as the core:
> 
> ```
> CREATE TABLE consumer_gates (
>     consumer_group VARCHAR(255) NOT NULL,
>     partition_key  VARCHAR(255) NOT NULL, -- empty means all partitions
>     reason         TEXT NOT NULL,
>     created_by     VARCHAR(255) NOT NULL,
>     created_at_ms  BIGINT NOT NULL,
>     PRIMARY KEY (consumer_group, partition_key)
> );
> ```

File-based implementation is for e2e tests only running on the same machine. While it is theoretically possible to hook it with configuration delivery (to let it survive a node crash / restart), it is indeed easier to come up with a centralized database-backed implementation. The cost would be a single synchronous database call per each controller run - files are way faster and simpler too - the gate could be closed or open with basic shell commands.

albertywu:
> It depends on a delivery order. For synchronized queues (only one message could be processed at a time), other messages would be held and won't be extended. For parallel queues, if there is enough controller instances on a production or e2e test system, multiple messages would be parked and their life extended as a result of this.
> It is true that messages not accepted will not be extended. For e2e tests, this is not a problem as we control the number of messages to be validated.

Thanks, the parallel-partition case makes sense. Since the RFC documents production debugging use-case ("Operational pause during incident"), my concern is multiple messages in the same partition, where one lease owner processes serially. Only the head reaches the gate; the prefetched tail remains in flight without visibility extensions and can exhaust retries.

Potential fixes:

1. Require `BatchSize=1` for gateable subscriptions.
2. Extend visibility for every buffered delivery in the gated partition.
3. Prevent additional prefetch while a partition is gated.

For the E2E-only scope, option 1 seems simplest, at the cost of throughput

behinddwalls:
> > It depends on a delivery order. For synchronized queues (only one message could be processed at a time), other messages would be held and won't be extended. For parallel queues, if there is enough controller instances on a production or e2e test system, multiple messages would be parked and their life extended as a result of this.
> > It is true that messages not accepted will not be extended. For e2e tests, this is not a problem as we control the number of messages to be validated.
> 
> Thanks, the parallel-partition case makes sense. Since the RFC documents production debugging use-case ("Operational pause during incident"), my concern is multiple messages in the same partition, where one lease owner processes serially. Only the head reaches the gate; the prefetched tail remains in flight without visibility extensions and can exhaust retries.
> 
> Potential fixes:
> 
> 1. Require `BatchSize=1` for gateable subscriptions.
> 2. Extend visibility for every buffered delivery in the gated partition.
> 3. Prevent additional prefetch while a partition is gated.
> 
> For the E2E-only scope, option 1 seems simplest, at the cost of throughput

ideally we just extend for everything in the buffer...so the semantics remain usable for batched polling as well

sbalabanov:
> > > It depends on a delivery order. For synchronized queues (only one message could be processed at a time), other messages would be held and won't be extended. For parallel queues, if there is enough controller instances on a production or e2e test system, multiple messages would be parked and their life extended as a result of this.
> > > It is true that messages not accepted will not be extended. For e2e tests, this is not a problem as we control the number of messages to be validated.
> > 
> > 
> > Thanks, the parallel-partition case makes sense. Since the RFC documents production debugging use-case ("Operational pause during incident"), my concern is multiple messages in the same partition, where one lease owner processes serially. Only the head reaches the gate; the prefetched tail remains in flight without visibility extensions and can exhaust retries.
> > Potential fixes:
> > 
> > 1. Require `BatchSize=1` for gateable subscriptions.
> > 2. Extend visibility for every buffered delivery in the gated partition.
> > 3. Prevent additional prefetch while a partition is gated.
> > 
> > For the E2E-only scope, option 1 seems simplest, at the cost of throughput
> 
> ideally we just extend for everything in the buffer...so the semantics remain usable for batched polling as well

Do we have facilities to do that? Could be quite consuming.
Regardless my main goal is to have e2e tests working. Operational pauses will be future work and even before them there is always an option to just pause the server.

platform/extension/consumergate/consumergate.go
sbalabanov: Documented this explicitly on Entry.Watch, in the extension README, in the consumer wait path, and in the RFC: gating never acknowledges, nacks, rejects, removes, or moves the source delivery; it remains queue-owned and becomes eligible for normal redelivery if visibility lapses.

platform/extension/consumergate/consumergate.go
sbalabanov: I kept Admin. It names the capability-oriented write/inspection surface paired with the read-side Gate interface. Manager would imply ownership of the gate lifecycle, while Manage is an action rather than an interface role.

sbalabanov:
Rebased onto latest `main` and force-pushed the scoped update.

Changes in this revision:
- file-backed gating is now explicit opt-in via `CONSUMER_GATE_DIR`; services otherwise use the no-op implementation
- Compose enables the file backend only for E2E/local development
- RFC and extension docs now explicitly limit the file backend to E2E and single-host development, including restart/node/fleet limitations
- queue lifecycle is documented explicitly: gating does not ack/nack/reject/remove/move the source delivery; the queue retains ownership and normal redelivery applies after visibility lapses
- the broader production operational-pause and buffered-prefetch semantics are intentionally left for future work

Local validation passed: targeted unit tests, all four service builds, lint/tidy/Gazelle/mock checks, and both E2E suites.

behinddwalls (approved):

PR #379, opened 2026-07-16
feat(speculation): cancel batches through the tree via prioritize

behinddwalls:
### Why?

cancelBatch's build-cancellation step predates the speculation tree: it flipped a single Build entity keyed directly by batch.ID, with no build runner in the loop and no tree consulted. Builds are now found through the tree's per-path build mapping — a batch can hold more than one in-flight build across its paths — so batch-level cancellation has to work through the tree like every other stage.

### What?

Cancellation becomes a convergent sweep through the tree and the existing stage seams. Each cancelBatch pass resolves every non-terminal path through the path->build mapping and either settles it to Cancelled (no build, dangling mapping, or terminal build) or captures intent (Cancelling) while its build is still in flight. While anything is pending the batch stays BatchStateCancelling and the pass publishes the queue to prioritize; only a pass that finds nothing pending CASes the batch terminal, fans out dependents, and publishes conclude. A terminal-transition guard re-checks Cancelled paths on would-be-final passes and pulls any that still hides a live build (a pre-build Cancel racing the build stage's trigger) back to Cancelling, so the batch never settles over a running build.

Three stages close the loop: prioritize loads Cancelling batches' trees for routing only (their paths are never candidates, and rogue decisions naming them are skipped as illegal), the build stage cancels every in-flight build of a Cancelling batch regardless of the path's recorded status and never triggers builds for it, and buildsignal keeps polling and re-publishing speculate for Cancelling batches so the wind-down is recorded and the sweep re-woken.

The branch keeps its review history as commits: the first commit cancels through the tree with a direct build-runner call in speculate (a documented exception to "only the build stage calls the runner"), the second removes that exception per review — decisions live in the tree, prioritize is the one channel from tree state to build messages, the build stage stays the sole runner caller — and the third fixes comment wording (build system, not CI; credit speculation-level cancels alongside the tree-sweep race).

Note for the stack: the cancel-flow CAS-race retry classification (preetam/int/cancel-retryable-cas) is not an ancestor of this branch; without it a benign version race on the terminal CAS still dead-letters, so that fix should land first. Metric renames: skipped_halted -> skipped_terminal (build, buildsignal), no_speculating_batches -> no_batches (prioritize).

## Test Plan
- ✅ `go build ./...` and `go vet` clean; `go test ./submitqueue/... ./service/... ./platform/...` all pass
- ✅ `bazel test //submitqueue/... //service/...` — 56/56
- ✅ `make e2e-test` — stovepipe and submitqueue suites pass
- ✅ `make lint` / `make check-tidy` / `make check-gazelle` clean
- speculate: CancellingFlow rewritten around the two-phase convergence — pending passes publish prioritize only, settled passes run the terminal sequence, and the orphaned-build guard re-opens a poisoned Cancelled path
- build: Cancelling-batch test proves every in-flight build (including a still-Building path) is cancelled and nothing triggers
- buildsignal: halted test split into terminal-only skip plus Cancelling-keeps-polling
- prioritize: routed-not-ranked test including a rogue decision on a Cancelling batch's path

## Stack
1. #353
1. #365
1. #371
1. @ #379
1. #380
1. #381
1. #382


submitqueue/orchestrator/controller/speculate/speculate.go
```diff
@@ -921,33 +1004,142 @@ func (c *Controller) cancelBatch(ctx context.Context, batch entity.Batch) error
...
+			changed = true
+		}
+	}
+
+	if !changed {
+		return pending, nil
+	}
+
+	newVersion := tree.Version + 1
+	if err := c.store.GetSpeculationTreeStore().Update(ctx, batch.ID, tree.Version, newVersion, paths); err != nil {
```
albertywu, line 1101: **High risk: a normal cancel-tree CAS race can force `Cancelling` to `Failed`.** A stale speculate or prioritize worker can advance the tree after this pass reads it. This conditional update then returns wrapped `storage.ErrVersionMismatch`. Production does not classify that sentinel as retryable, so the cancellation message can enter DLQ; DLQ reconciliation changes the non-terminal cancelling batch to `Failed`, potentially before a live build is stopped. Reload and recompute on expected optimistic conflicts rather than dead-lettering them. This remains present through PR #380.

submitqueue/orchestrator/controller/speculate/speculate.go
```diff
@@ -921,33 +1004,142 @@ func (c *Controller) cancelBatch(ctx context.Context, batch entity.Batch) error
...
+	// buildsignal loop, so the orphan's stop would never be recorded. So on
+	// a pass that would otherwise settle, resolve pre-existing Cancelled
+	// paths' mappings too and pull any that still has a live build back to
+	// Cancelling for the build stage to enact. Only Cancelled can hide a
+	// live build — Passed and Failed are derived from terminal builds — and
+	// this is the one deliberate downgrade of a terminal path status in the
+	// system, costing extra point reads only on would-be-final passes.
+	if pending == 0 {
+		for i, p := range paths {
+			if resolved[i] || p.Status != entity.SpeculationPathStatusCancelled {
```
albertywu, line 1075: **High risk: this terminal guard skips the exact path a stale build worker can race.** A build worker can pass its early batch-state check, while this sweep observes no mapping and resolves the path to `Cancelled`. Because `resolved[i]` is true, the guard skips it. The stale worker can then trigger and create the mapping before the batch's terminal CAS, leaving a live build under a terminal batch. The trigger-before-marker race is inherited, but this PR's new guard does not close it. A durable pre-trigger reservation or fence is needed. This remains present through PR #380.

submitqueue/orchestrator/controller/prioritize/prioritize.go
```diff
@@ -193,10 +217,16 @@ func (c *Controller) loadTrees(ctx context.Context, batches []entity.Batch) (map
...
-func candidatesOf(trees map[string]entity.SpeculationTree) []entity.SpeculationPathInfo {
+// Prioritized/Building (holds one). Trees of batches in the cancelling set
+// are skipped wholesale — their paths are doomed, so they neither want a
+// slot nor hold one worth ranking; those trees are loaded only so
+// republishBuilds can route their persisted Cancelling intents.
+func candidatesOf(trees map[string]entity.SpeculationTree, cancelling map[string]bool) []entity.SpeculationPathInfo {
 	var candidates []entity.SpeculationPathInfo
-	for _, tree := range trees {
+	for batchID, tree := range trees {
+		if cancelling[batchID] {
```
albertywu, line 227: **High risk: cancelling work stops counting against capacity before the external build stops.** Skipping the whole batch removes its `Building` or `Cancelling` paths from occupancy immediately. With a limit of 1, batch A can still be stopping in the runner while batch B is promoted and triggered, producing two active external builds. The omission existed on this PR's base, but it remains unsafe under the new asynchronous wind-down protocol. Separate eligibility from occupancy and count a slot until the runner reports a terminal state. This remains present through PR #380.

behinddwalls:

PR #381, opened 2026-07-16
test(e2e): assert speculation-tree parity on the landing pipeline

behinddwalls:
### Why?

The tree-driven speculation stack (speculate rework, prioritize stage, path-driven build) claims exact behavioral parity with the previous naive pipeline. The e2e suite proved the black-box half (requests still land, timelines unchanged) but nothing verified the new persisted machinery: that the tree exists, holds exactly the parity chain path, and that exactly one build ran per batch.

### What?

White-box e2e assertions on top of the existing flows (no new scenarios): the suite now constructs the batch, speculation-tree, and build mysql stores against the app DB like the other stores, with harness helpers to map a request to its batch and fetch its tree/builds. The happy-path landing test additionally asserts the landed batch's tree has exactly one path (Head = batch, Base = the batch's dependencies in order), reached `passed` with a stamped BuildID and a positive naive score, a version past the initial write, and that exactly one build exists whose speculation path matches and succeeded. The cancel test asserts no non-terminal build survives cancellation and, when the batch did reach Cancelled with a tree, that every path is `cancelled` (tolerating cancel-before-speculation leaving no tree). Comparisons normalize the nil-vs-empty-slice JSON artifact between `batch.Dependencies` and `path.Base` while keeping ordered-content equality.

## Test Plan
✅ `make gazelle && make fmt` clean; ✅ `make e2e-test` — both e2e targets pass with the new assertions.

## Stack
1. #353
1. #365
1. #371
1. #379
1. #380
1. @ #381
1. #382

### DIFF
--- a/test/e2e/submitqueue/harness_test.go
+++ b/test/e2e/submitqueue/harness_test.go
@@ -34,7 +34,23 @@ import (
 	mergestrategypb "github.com/uber/submitqueue/api/base/mergestrategy/protopb"
 	gatewaypb "github.com/uber/submitqueue/api/submitqueue/gateway/protopb"
 	"github.com/uber/submitqueue/submitqueue/entity"
+	"github.com/uber/submitqueue/submitqueue/extension/storage"
 )
+
+// allBatchStates enumerates every entity.BatchState, active and terminal. Neither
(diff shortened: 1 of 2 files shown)


behinddwalls:

PR #382, opened 2026-07-16
fix(orchestrator): re-trigger publishes need distinct message IDs

behinddwalls:
The mysql message queue makes publishes idempotent on the (topic, partition_key, id) unique key by documented design (safe publish retries), and rows are only removed by offset-based garbage collection — so a publish reusing a prior message ID is silently coalesced against any not-yet-collected row, including already-consumed ones. The new speculation pipeline reuses IDs in exactly the places that must RE-trigger work, and the Phase-B respeculation e2e run caught all of them at once: prioritize rounds published under the bare queue name meant only the first round per queue ever ran (a batch becoming Speculating a moment later was never promoted, never built, and waited forever), and the dependency-failure fan-out published under the dependent's bare batch ID — the same key the score stage had already used — so the wake-up vanished and the dependent stayed stuck at scored.

Every re-trigger publish now mints an ID that is unique exactly when a new round/wake-up is needed and coalescing when it is not. Prioritize rounds: `{queue}/{batch}/v{treeVersion}` — every persisted tree change is guaranteed a round (and every prioritizer input change is some tree's version bump), while no-change passes coalesce. Dependent fan-out: `{dependent}/dep-terminal/{dep}` — one wake-up per (dependent, terminal dependency), self-heal redeliveries coalesce among themselves. Prioritize's build republish: `{batch}/v{treeVersion}` — promotion rounds are guaranteed delivery, pure heal republishes coalesce.

The terminal self-heal branch now re-fans dependents for every terminal state, not just Cancelled/Failed: Succeeded is how dependents learn a dep landed and dead-path the siblings that bet against it — mergesignal already publishes the terminal batch to speculate on success, so this branch is precisely its landing spot.

✅ `make fmt && make test` (79/79) — speculate suite updated for the new IDs (per-test versions derived from each test's own Update expectations) and for Succeeded's dependent re-fan. The Phase-B respeculation e2e (next in stack) is the regression test that caught both drops: it goes from permanently stuck at scored to landing.

## Stack
1. #353
1. #365
1. #371
1. #379
1. #380
1. #381
1. @ #382

### DIFF
--- a/submitqueue/orchestrator/controller/prioritize/prioritize.go
+++ b/submitqueue/orchestrator/controller/prioritize/prioritize.go
@@ -392,7 +392,15 @@ func (c *Controller) republishBuilds(ctx context.Context, queue string, trees ma
 		if !needsBuild {
 			continue
 		}
-		if err := c.publishBuild(ctx, batchID, queue); err != nil {
+		// The message ID encodes (batch, tree version): a round that changed
+		// the tree (new promotions) mints a fresh ID and is guaranteed
+		// delivery, while pure heal republishes reuse the last one and
+		// coalesce under the queue's publish idempotency on
+		// (topic, partition_key, id) — a bare batch ID would coalesce a
+		// later promotion round away against the first round's
+		// not-yet-collected row.
+		msgID := fmt.Sprintf("%s/v%d", batchID, tree.Version)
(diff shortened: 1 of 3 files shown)


behinddwalls:

PR #383, opened 2026-07-16
[1/N] docs(sqsim): propose local scenario simulator

albertywu:
## Summary

RFC for SQSim (a Submitqueue Simulation Tool)

## Experience Video

https://github.com/user-attachments/assets/d6132330-6885-49ac-833a-97fa9079a193

## Stack
1. @ #383
1. #384
1. #385
1. #386
1. #387
1. #388
1. #389
1. #390
1. #391
1. #392
1. #393


sbalabanov (commented):
With all the live components, realtime clock and scenarios defined through code, the system is very much close to the e2e test rather than simulator.
In the past, simulator was useful to replay a certain traffic from a production system but at shrunk time intervals, which would require adjustable clock and some components mocked - like a CI/build integration.
Let's discuss a proper designation for the framework.

sbalabanov (changes requested):
back for discussions, see above comment

PR #394, opened 2026-07-16
docs(stovepipe): record stage RFC

mnoah1:
Updating record step doc to align with current implementation (now mostly implemented through phase 1), and forward looking plans for Hooks and future phases.

### DIFF
--- a/doc/rfc/stovepipe/steps/record.md
+++ b/doc/rfc/stovepipe/steps/record.md
@@ -1,0 +1,302 @@
+# Record stage
+
+`record` turns a terminal build outcome into a durable validation fact.
+
+- Phase 1 records whole-repository greenness. On green it also advances the Queue's last-green bookmark and promotes the commit onto the Queue's promotion ref. This is implemented: see [stovepipe/controller/record/record.go](../../../../stovepipe/controller/record/record.go).
+- Phase 2 records greenness per project instead of per repository. Sketched here, to be expanded before implementation.
+
+Notifying downstream systems is **not** implemented in either phase. It will ride the cross-domain hook framework instead of a Stovepipe-specific extension; see [Hooks](#hooks).
+
+See [workflow.md](../workflow.md) for the whole pipeline, [build.md](build.md) for how builds are created, and [buildsignal.md](buildsignal.md) for the terminal-only handoff into this stage.
+
+`record` owns the validation facts and the two caches derived from them, the last-green bookmark and the promotion ref. It does not decide build scope, poll a build, release the Queue's build slot, interpret a target graph, or map targets to projects. Those belong to `process`, `buildsignal`, and `analyze`.
+
+## Phase 1 algorithm
+
+For a delivery carrying a `Record` payload:
+
+```
+1. Deserialize the payload into (request id R, queue name Q).
+   - malformed -> return raw; non-retryable.
+
+2. Resolve the queue's storage from Q.
+   - unresolvable -> return raw; non-retryable. A queue with no storage is a
+     malformed message or a config error, not a transient condition.
+
+3. Load Request R.
+   - ErrNotFound       -> return raw; non-retryable.
+   - other store error -> return raw; the classifier decides.
+
+4. If Q is non-empty and disagrees with R.Queue, return a non-retryable error.
+   The Request is the authoritative binding; Q is a routing hint, empty on
+   payloads written before the field existed.
+
+5. Inspect R.State.
+   - succeeded / failed -> continue. buildsignal stamps the outcome before it
+     publishes here, and both values are verdicts about the code.
+   - cancelled -> ack, no fact: the build decided nothing about the commit (see
+     "When to record an outcome").
+   - superseded -> ack, no fact. Unreachable in practice.
+   - accepted / processing / anything else -> return a non-retryable invariant error.
+
+6. Map R.State to a whole-repository degree and create the ValidationFact keyed by
+   (R.URI, empty project) in the queue-bound fact store.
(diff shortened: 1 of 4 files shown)


doc/rfc/stovepipe/steps/record.md
```diff
@@ -0,0 +1,228 @@
...
+Phase 2 inserts a non-terminal `analyzing` state ("whole-repository fact recorded, slot released, project analysis in flight") between `processing` and the recorded states, and broadens the recorded states to mean "all planned facts recorded".
+
+## Hooks
+
+Hooks are the notification boundary, not the source of truth; the GreennessStore is authoritative and commits before any hook fires.
+
+Following the extension rule "identity in, resolve internally," the Hooks contract takes a thin greenness identity rather than a controller-assembled external payload:
+
+```
+Notify(ctx, GreennessRef{Queue, URI, Project}) error
```
behinddwalls: we may want to have this as a separate RFC for Hook and hooks could anything including notifications so we may want to name it agnostic of the function it serves

doc/rfc/stovepipe/steps/record.md
```diff
@@ -0,0 +1,228 @@
...
+
+## Greenness is an immutable fact
+
+A greenness fact answers "how broken was this scope at this Queue URI?" Its identity is:
+
+```
+(queue, uri, project)
+```
+
+`project` is empty for whole-repository greenness and is a stable project id in Phase 2. The key is derivable from identity the controller already holds, so storage requires no query by attribute or secondary index.
```
behinddwalls: it records the status at the given commit? how do we answer the question of what is prev or next green commit for a given commit? how would that work? should that be a separate rfc?

doc/rfc/stovepipe/steps/record.md
```diff
@@ -0,0 +1,228 @@
...
+| `cancelled`           | `1`    |
+
+Intermediate degrees are reserved for project analysis and deferred with the project mapping contract; Phase 1 does not manufacture fractional values.
+
+## Phase 1 algorithm
+
+For a delivery carrying build id `B`:
+
+```
+1. Load Build B.
```
behinddwalls: why Build? why not the request? Shouldn't the buildsignal already capturing the status of request?

doc/rfc/stovepipe/steps/record.md
mnoah1: Yeah, I think we need a separate one to elaborate on how hooks will work. Left his high level section and will start something separate for this.

doc/rfc/stovepipe/steps/record.md
mnoah1: Now using request.

doc/rfc/stovepipe/steps/record.md
mnoah1: Added a high level section ("Coverage of intermediate commits") with the current update, but I think this needs its own RFC as it could be more involved and there are several different possible approaches.  I'll start on that separately.

doc/rfc/stovepipe/steps/record.md
```diff
@@ -0,0 +1,215 @@
...
+
+
+| Field       | Meaning                                                                             |
+| ----------- | ----------------------------------------------------------------------------------- |
+| `Queue`     | Stable Queue name that namespaces the validation                                    |
+| `URI`       | Commit URI under validation                                                         |
+| `Project`   | Empty for the whole repository; stable project id for Phase 2                       |
+| `Degree`    | Health degree in the closed interval `[0, 1]`; `0` is green and `1` is fully broken |
+| `RequestID` | Request that established the fact                                                   |
+| `CreatedAt` | Millisecond timestamp at which the fact was first recorded                          |
```
mnoah1: Thoughts on timestamp?  We don't track timestamps elsewhere in the system but for these more long-term persisted facts, do we want to maintain some kind of timestamped history?

doc/rfc/stovepipe/steps/record.md
```diff
@@ -0,0 +1,215 @@
...
+
+Competing verdicts cannot reach this point: duplicate builds for one Request are resolved a stage earlier, where `buildsignal`'s first-writer-wins outcome write discards the losing build's verdict, so `record` only ever sees one.
+
+Absence remains distinct from degree `0`. Callers gating deployments must treat absence as not green.
+
+### When to record an outcome
+
+A fact is written only when the request reaches a `succeeded` or `failed` verdict.  A `cancelled` build  is acked with no fact recorded. Callers gating deployments treat an absent fact as not green.
+
+The fail-closed path also produces no fact, as the DLQ reconciler forces `failed` and never publishes to `record`.
```
roychying: I don't think `never publishes to record` holds when the fail-close originates in `build_dlq`. if `build` ack keeps failing, once `retry_count` hits `MaxAttempts`, the subscriber itself moves it to the DLQ, leaving one or more live poll chains behind in `buildsignal` step, and `publishRecord` fires anyway, `record` then writes a degree-1 fact regardless of the build result.

mnoah1: Updated this. `publishRecord` sits outside the guard that makes `finishRequest` idempotent, and the entry guard admits a request that already carries an outcome — deliberately, so a redelivery after the outcome was stamped but before the publish landed still reaches `record`. So a forced `failed` gets carried into a fact by any delivery arriving afterwards. "The reconciler never publishes to record" was only ever true of the reconciler, not of the path, and it contradicted the late-success edge case a few sections down.

behinddwalls (approved):

PR #397, opened 2026-07-16
refactor(metrics): replace NamedTimer with histograms

sbalabanov:
## What

Removes the `NamedTimer` metrics helper and stops recording **any** duration as a tally timer. All durations are now **histograms**, with buckets chosen per operation.

- Deletes `metrics.NamedTimer`.
- `Op.Complete` no longer emits a `latency` timer and takes a **required** `buckets` parameter — there is no baked-in default, since operations differ widely in expected latency. Every callsite passes an appropriate set.
- Exports three common duration bucket sets so callers have ready-made choices:
  - `FastLatencyBuckets` (~100µs–5s) — fast in-process work (scoring, cache lookups).
  - `StorageLatencyBuckets` (~1ms–1m) — DB / message-queue round-trips, RPC handlers.
  - `LongLatencyBuckets` (~5ms–4h) — long-running pipeline work and external calls (builds, merges, git, providers).
- One-shot duration metrics (consumer `controller_latency`/`ack_nack_latency`, mysql subscriber `poll.latency`/`poll.message_age`) use `NamedHistogram(..., buckets, ...).RecordDuration(d)` with an operation-appropriate set. No dedicated one-shot helper is added.
- Updates unit tests and `platform/metrics/README.md`.

## Why timers are a bad fit for a distributed system

A tally **timer** ships raw per-datapoint durations, and the monitoring backend derives percentiles (p50/p99/max) **per time series** — one series per unique combination of metric name and tag values, so every distinct tag value (region, zone, …) multiplies the series count. The moment a dashboard or alert spans more than one series — rolling up a tag you didn't pin to a single value — the backend has to combine already-aggregated per-series statistics, and that is where timers fall apart:

- **Percentiles don't combine.** The p99 across N series is *not* the average, the max, or any function of each series' p99. `avg(series_a.p99, series_b.p99)` is mathematically meaningless — you're combining summary statistics of different underlying distributions with different sample counts. The only aggregate that survives is the (count-weighted) **mean**.
- **`max` is lossy too.** `max(per-series maxes)` happens to be correct, but everything between the mean and the max — exactly the tail latency you care about during an incident — is imprecise.
- **Precision degrades with the number of series aggregated.** With every tag pinned (a single series) a timer is fine; the error grows as you roll up more series, and since each tag value adds series, high-cardinality tags make it worse. The rollups you reach for during an incident ("p99 across the whole region") are exactly the imprecise ones.

A **histogram** avoids this because it emits **bucket counts**, not pre-aggregated statistics. Bucket counts are additive: the backend merges any set of series by simply **summing each bucket** across them, reconstructing the true combined distribution. Percentiles are then read off the merged buckets, so p50/p99/max stay accurate at *any* aggregation level and over *any* time window. The trade-off is bucket granularity — hence per-operation bucket sets rather than one broad default, so resolution lands where the data is and empty far-out buckets don't waste series cardinality.

## Testing

- `go test ./platform/... ./submitqueue/... ./runway/... ./stovepipe/...` — pass
- `make fmt`, `make gazelle`, `go build ./...`, `go vet` (changed pkgs) — clean

### DIFF
--- a/platform/metrics/metrics.go
+++ b/platform/metrics/metrics.go
@@ -34,41 +34,88 @@ func NewTag(key, value string) Tag {
 	return Tag{Key: key, Value: value}
 }
 
-// defaultLatencyBuckets provides pre-defined duration buckets for common latency histograms.
-// Covers sub-millisecond to multi-hour ranges suitable for RPC calls, queue processing,
-// and long-running operations like builds and merges.
-var defaultLatencyBuckets = tally.DurationBuckets{
-	5 * time.Millisecond,
-	10 * time.Millisecond,
-	25 * time.Millisecond,
-	50 * time.Millisecond,
-	100 * time.Millisecond,
-	250 * time.Millisecond,
-	500 * time.Millisecond,
-	1 * time.Second,
-	2500 * time.Millisecond,
-	5 * time.Second,
-	10 * time.Second,
-	30 * time.Second,
-	1 * time.Minute,
-	2 * time.Minute,
-	5 * time.Minute,
-	10 * time.Minute,
-	30 * time.Minute,
-	1 * time.Hour,
-	2 * time.Hour,
-	4 * time.Hour,
-}
+// Common duration bucket sets for latency histograms. Operations differ widely
+// in expected latency, so there is no single default — pick the set whose range
+// matches the operation and pass it to Op.Complete or NamedHistogram. Buckets
+// far outside an operation's real latency waste series cardinality and lose
+// resolution where the data actually lands.
+var (
+	// FastLatencyBuckets suits fast in-process operations (~microseconds to
+	// seconds): scoring, cache lookups, and other CPU-bound work.
+	FastLatencyBuckets = tally.DurationBuckets{
+		100 * time.Microsecond,
+		250 * time.Microsecond,
+		500 * time.Microsecond,
+		1 * time.Millisecond,
+		2500 * time.Microsecond,
+		5 * time.Millisecond,
+		10 * time.Millisecond,
+		25 * time.Millisecond,
+		50 * time.Millisecond,
+		100 * time.Millisecond,
+		250 * time.Millisecond,
+		500 * time.Millisecond,
+		1 * time.Second,
+		2500 * time.Millisecond,
+		5 * time.Second,
+	}
+
+	// StorageLatencyBuckets suits storage and message-queue round-trips
+	// (~1ms to a minute): database reads/writes, publish/consume, and RPC
+	// handlers whose latency is dominated by such calls.
+	StorageLatencyBuckets = tally.DurationBuckets{
+		1 * time.Millisecond,
+		2500 * time.Microsecond,
+		5 * time.Millisecond,
+		10 * time.Millisecond,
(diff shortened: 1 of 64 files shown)


platform/metrics/metrics.go
```diff
@@ -110,18 +112,27 @@ func (o Op) Complete(err error) {
...
+// that don't fit the Begin/Complete lifecycle.
+//
+// Prefer histograms over timers for durations: timer percentiles (p50/p99/max)
+// are computed per time series and cannot be combined, so aggregating across
+// series — rolling up regions, zones, or any tag not pinned to one value —
+// yields imprecise percentiles (only the mean survives). Since every tag value
+// adds series, the error grows with how many series you aggregate. Bucketed
+// histograms merge exactly by summing buckets, so percentiles stay accurate at
+// any aggregation level.
+func NamedDurationHistogram(scope tally.Scope, name string, histogram string, d time.Duration, tags ...Tag) {
```
sbalabanov: There is no need to have NamedDurationHistogram helper. Instead, have a client always pass buckets to NamedHistogram if needed.

platform/metrics/metrics.go
```diff
@@ -87,19 +87,21 @@ func Begin(scope tally.Scope, name string, tags ...Tag) Op {
...
+// percentiles cannot be combined across time series (see NamedDurationHistogram).
 func (o Op) Complete(err error) {
 	elapsed := time.Since(o.start)
 
 	if err == nil {
 		o.scope.Counter("succeeded").Inc(1)
 		s := o.scope.Tagged(map[string]string{"result": "success"})
-		s.Timer("latency").Record(elapsed)
-		s.Histogram("latency_histogram", defaultLatencyBuckets).RecordDuration(elapsed)
+		s.Histogram("latency", defaultLatencyBuckets).RecordDuration(elapsed)
```
sbalabanov: Do not use default buckets - all the operations are very different. Instead, modify Complete helper to accept histogram buckets as required parameter. Modify all of the callsites.

platform/metrics/metrics.go
sbalabanov: metrics package could export some common duration buckets like this to implement what is required by other comments

platform/metrics/metrics.go
sbalabanov: Done — removed `NamedDurationHistogram`. One-shot callsites now use `NamedHistogram(scope, name, hist, buckets, ...tags).RecordDuration(d)` with an explicit bucket set.

platform/metrics/metrics.go
sbalabanov: Done — `Op.Complete(err, buckets)` now takes buckets as a required parameter with no default, and all ~80 callsites were updated to pass an operation-appropriate set (Fast/Storage/Long).

platform/metrics/metrics.go
sbalabanov: Done — the package now exports `FastLatencyBuckets` (~100µs–5s), `StorageLatencyBuckets` (~1ms–1m), and `LongLatencyBuckets` (~5ms–4h). Callsites pick the set matching the operation.

behinddwalls (approved):

PR #398, opened 2026-07-17
docs(orchestrator): document controller correctness

albertywu:
## Summary
Document controller correctness guidelines

## Test Plan

### DIFF
--- a/submitqueue/orchestrator/controller/README.md
+++ b/submitqueue/orchestrator/controller/README.md
@@ -1,0 +1,112 @@
+# Controller Correctness
+
+SubmitQueue controllers are built around eventual consistency. Controllers advance a workflow through durable state checkpoints, and every component must tolerate retries before the next checkpoint is recorded.
+
+The core model is:
+
+> Load durable state, reconcile it toward this controller's checkpoint, then replay the checkpoint's fanout until it is accepted.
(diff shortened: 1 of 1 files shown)


submitqueue/orchestrator/controller/README.md
```diff
@@ -0,0 +1,115 @@
...
+Prefer one reconciliation pass per delivery. The consumer framework is the retry loop:
+
+```text
+controller returns error
+    -> error processor classifies it
+    -> consumer nacks retryable errors
+    -> redelivery re-enters Process and reloads durable state
+```
+
+Do not add an unbounded retry loop inside a controller. On `storage.ErrVersionMismatch`, return the wrapped sentinel and let the shared domain classifier mark it retryable. Redelivery then starts from a fresh read and decides whether the transition is still needed, already happened, or was superseded.
```
sbalabanov: I do not think this is a necessary statement to make.

sbalabanov (approved):

submitqueue/orchestrator/controller/README.md
albertywu: Agreed. Removed this statement and the matching checklist item to keep the guideline focused on the reconciliation model.

PR #399, opened 2026-07-17
fix(orchestrator): fix score reconciliation bugs

albertywu:
## Summary
Fix score reconciliation so redelivery converges from the batch's durable state instead of repeating or regressing work.

### Bugs

- The score controller treated every non-cancelling, non-terminal batch as needing to be scored. A stale score delivery could therefore move a batch that had already advanced to `Speculating` or `Merging` back to `Scored`.
- If the `Scored` state was persisted but publishing request logs or the `speculate` message failed, redelivery recalculated the score and wrote another `Scored` version instead of replaying fanout from the persisted checkpoint.
- `storage.ErrVersionMismatch` was not marked retryable. An optimistic-lock race could therefore be rejected and sent to the DLQ instead of being redelivered to reconcile from fresh state.

### Fix

- Reconcile explicitly from the current batch state:
  - `Created`: calculate the score and conditionally transition to `Scored`.
  - `Scored`: preserve the persisted score and replay request logs plus the `speculate` message.
  - `Speculating` or `Merging`: acknowledge the stale score delivery without regressing state.
  - `Cancelling` or terminal: acknowledge because another controller owns recovery.
- Wrap `storage.ErrVersionMismatch` with `errs.NewRetryableError` so the consumer nacks the delivery and reloads durable state on redelivery.
- Add regression coverage for downstream-state protection, version-conflict redelivery, and replay of fanout for an already-`Scored` batch.

## Test Plan
- `make fmt`
- `make lint`
- `make check-tidy`
- `make check-gazelle`
- `./tool/bazel test //submitqueue/orchestrator/controller/score:go_default_test --test_output=errors`

### DIFF
--- a/submitqueue/orchestrator/controller/score/score.go
+++ b/submitqueue/orchestrator/controller/score/score.go
@@ -16,11 +16,13 @@ package score
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/uber-go/tally"
 	entityqueue "github.com/uber/submitqueue/platform/base/messagequeue"
 	"github.com/uber/submitqueue/platform/consumer"
+	"github.com/uber/submitqueue/platform/errs"
 	"github.com/uber/submitqueue/platform/metrics"
 	corerequest "github.com/uber/submitqueue/submitqueue/core/request"
 	"github.com/uber/submitqueue/submitqueue/core/topickey"
@@ -104,10 +106,10 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) (r
 		"partition_key", msg.PartitionKey,
 	)
 
-	// Short-circuit when the batch is in BatchStateCancelling — the cancel
-	// controller has handed the batch off to speculate, which owns the terminal
-	// write to Cancelled and the downstream dependent / conclude publishes. We
-	// must not race it to conclude (conclude requires terminal). Silently ack.
+	var batchScore float64
+	// Short-circuit when the batch is in BatchStateCancelling. The cancel
+	// controller has handed the batch off to speculate, which owns the
+	// terminal write and downstream fanout.
 	if batch.State == entity.BatchStateCancelling {
 		c.metricsScope.Counter("skipped_cancelling").Inc(1)
 		c.logger.Infow("skipping score for cancelling batch",
@@ -116,12 +118,8 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) (r
 		return nil
 	}
 
-	// Short-circuit if the batch is already terminal. Score never writes a
-	// terminal state, so it owns no recovery here: whichever controller wrote
-	// the terminal state (speculate.cancelBatch / failOnDependency, or merge)
-	// already published to conclude, and speculate's terminal self-heal
-	// republishes conclude on every redelivery of a terminal batch. Silently
-	// ack — same pattern as build / buildsignal on halted.
+	// Score owns no terminal-state recovery. The controller that wrote the
+	// terminal state owns its remaining fanout.
 	if batch.State.IsTerminal() {
 		c.metricsScope.Counter("skipped_terminal").Inc(1)
 		c.logger.Infow("skipping score for terminal batch",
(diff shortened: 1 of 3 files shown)


behinddwalls:
The retry loop shouldn't be here. The consumer framework is already the retry loop: a nack re-enters `Process`, re-reads the batch, and re-dispatches on state — same convergence as the `continue`, via the broker.

The actual gap is classification: `storage.ErrVersionMismatch` is unclassified, so it defaults to non-retryable and dead-letters. Fix that once instead of patching it per controller:

- Drop the loop. On `ErrVersionMismatch`, bump `version_conflicts` and return the wrapped error.
- Add a domain classifier mapping `storage.ErrVersionMismatch → InfraRetryable`, composed into the pipeline consumers' `ClassifierProcessor` in the orchestrator wiring (domain-side — `platform/errs` can't see the sentinel).

That covers every CAS-writing controller (score, speculate, merge, cancel — the cancel DLQ race is this same bug). Controllers stay thin state machines; retryability stays in classifiers, which is where this repo puts it.

Keep the new state branches — that's the real fix. With the loop gone, add a direct test for redelivery of an already-`Scored` batch; it's currently only covered via the reload path.

sbalabanov:
ErrVersionMismatch is likely always retryable?
In which case it could be created as infra_retryable right at the place where it is declared

albertywu:
@behinddwalls @sbalabanov Addressed feedback in `5904769`:

- Removed the controller-local retry loop. `Process` now performs one reconciliation pass, increments `version_conflicts` on `storage.ErrVersionMismatch`, and returns the wrapped error.
- Kept the explicit state branches. An already-`Scored` batch reuses its persisted score and replays logs plus `speculate`; `Speculating` and `Merging` are acknowledged without regressing state.
- Added direct tests for version-conflict return and redelivery of an already-`Scored` batch.

sbalabanov (commented):
If Runway and Stovepipe use the same optimistic concurrency pattern with database checkpoints, then we have to
- either duplicate such a classifier across all services
- or come up with a platform alternative - but it should not have access to storage-specific sentinels
- or go with "classify at declaration" path

which one to prefer?

submitqueue/orchestrator/controller/score/score.go
```diff
@@ -131,25 +128,56 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) (r
...
+		err = c.store.GetBatchStore().UpdateScoreAndState(
+			ctx,
+			batch.ID,
+			batch.Version,
+			newVersion,
+			batchScore,
+			entity.BatchStateScored,
+		)
+		if errors.Is(err, storage.ErrVersionMismatch) {
+			c.metricsScope.Counter("version_conflicts").Inc(1)
```
sbalabanov: why needed? in all other places we just bubble the error up

submitqueue/orchestrator/controller/score/score.go
```diff
@@ -131,25 +128,56 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) (r
...
-	if err := c.store.GetBatchStore().UpdateScoreAndState(ctx, batch.ID, batch.Version, newVersion, batchScore, entity.BatchStateScored); err != nil {
-		metrics.NamedCounter(c.metricsScope, opName, "storage_errors", 1)
-		return fmt.Errorf("failed to update score for batch %s: %w", batch.ID, err)
-	}
-	batch.Version = newVersion
+	case entity.BatchStateScored:
+		// The durable transition already happened, but its fanout may be
+		// incomplete. Preserve the committed score and replay all outputs.
+		batchScore = batch.Score
+		c.metricsScope.Counter("replayed_scored").Inc(1)
```
sbalabanov: probably not needed metric

submitqueue/orchestrator/controller/score/score.go
```diff
@@ -131,25 +128,56 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) (r
...
 
-	c.logger.Infow("scored batch",
-		"batch_id", batch.ID,
-		"score", batchScore,
-	)
+	case entity.BatchStateSpeculating, entity.BatchStateMerging:
+		// A downstream controller consumed the handoff, so this score
+		// delivery is stale and must not regress the batch to Scored.
+		c.metricsScope.Counter("skipped_downstream").Inc(1)
+		return nil
```
sbalabanov, line 180: can it ever happen by any reasonable flow of events? If not, let's fail it hard.

sbalabanov (changes requested):

submitqueue/orchestrator/controller/score/score.go
albertywu: Agreed. Removed the scorer-specific `version_conflicts` check and metric. The controller now wraps and returns the storage error directly. The shared SubmitQueue classifier recognizes `storage.ErrVersionMismatch`, and the consumer nacks it for redelivery.

submitqueue/orchestrator/controller/score/score.go
albertywu: Agreed. Removed `replayed_scored`. The `Scored` reconciliation branch remains because it reuses the durable score and replays incomplete log and speculate fanout.

submitqueue/orchestrator/controller/score/score.go
albertywu: Yes, this is reachable under normal at-least-once delivery. A score attempt can persist `Scored` and publish to speculate, downstream can advance the batch to `Speculating` or `Merging`, and then a lost score acknowledgement can cause the original delivery to be redelivered. I kept this as a successful stale no-op and expanded the comment with that sequence. Failing here would send a valid duplicate to the DLQ.

albertywu:
> If Runway and Stovepipe use the same optimistic concurrency pattern with database checkpoints, then we have to either duplicate such a classifier across all services
or come up with a platform alternative - but it should not have access to storage-specific sentinels
or go with "classify at declaration" path
which one to prefer?

@sbalabanov explored a bit, and currently in favor of status quo which is "classify at declaration" path. Updated the code to mark the CAS ErrVersionMismatch as Retryable when returning it in the controller.

sbalabanov (approved):

behinddwalls:

The rebase succeeded but force-push failed for `wua/scorer-controller-audit`. This may be due to a concurrent push. Please rebase manually.

PR #401, opened 2026-07-20
feat(stovepipe): add BuildRunner extension contract and fake backend

roychying:
## Why?
Stovepipe's build pipeline step needs a vendor-agnostic way to trigger and poll builds against an external CI system before it can wire up a real backend (e.g. Buildkite/GitHub Actions) or the controller that drives it. SubmitQueue already has this shape via its own buildrunner extension, but Stovepipe validates a single commit against a baseline rather than a stack of dependency batches, so it needs its own interface with URI-based identity instead of batch identity — see doc/rfc/stovepipe/steps/build.md's "Why separate contracts" for the full rationale. This PR lands that contract plus a stateless fake implementation.

## What?
- `stovepipe/extension/buildrunner/buildrunner.go`: the BuildRunner interface (Trigger/Status/Cancel), Config, and the Factory interface, per the repo's extension rules (concrete factories and per-queue routing stay in the wiring layer). Trigger(ctx, headURI, baseURI, metadata) (entity.BuildID, error) mints a fresh build id every call — no caller-supplied dedup key; Status may round-trip to the backend; Cancel is defined for contract parity but not yet called by any controller.
- `stovepipe/extension/buildrunner/fake/fake.go`: a stateless BuildRunner for tests/examples. Defaults to succeeding immediately; failure modes are injected via a buildrunner-fake=<token> marker embedded in headURI (trigger-error, build-fail, build-error). The outcome is encoded directly in the returned BuildID (with a random suffix for uniqueness) so Status needs no per-build bookkeeping and can even be served by a different runner instance/process than the one that triggered it.
- `stovepipe/extension/buildrunner/mock/buildrunner_mock.go`: generated gomock mock for BuildRunner, for controller-level unit tests.

## Test Plan
make gazelle, make build, make test - all clean.


stovepipe/extension/buildrunner/buildrunner.go
```diff
@@ -0,0 +1,86 @@
...
+//     platform/errs.
+type BuildRunner interface {
+	// Trigger starts a new build every call and mints the build's identity —
+	// there is no caller-supplied dedup input. headURI is the commit under
+	// validation; baseURI is the incremental baseline, empty for a full
+	// build; both are opaque tokens owned by SourceControl. metadata is
+	// caller-supplied annotation the runner may echo back via Status but must
+	// not depend on. Trigger is async: it must return promptly with the
+	// runner-assigned id, not an outcome — callers learn progress via Status.
+	Trigger(ctx context.Context, headURI, baseURI string, metadata entity.BuildMetadata) (entity.BuildID, error)
```
behinddwalls: maybe we reverse the order `baseURI` and then `headURI`?

stovepipe/extension/buildrunner/README.md
```diff
@@ -0,0 +1,21 @@
+# BuildRunner
```
behinddwalls, line 1: you could ask claude to trim this and only document behavioural things

stovepipe/extension/buildrunner/buildrunner.go
roychying: Reversed

stovepipe/extension/buildrunner/README.md
roychying: sure. removed other contents

behinddwalls (approved):

PR #407, opened 2026-07-20
feat(stovepipe): add the build stage controller

roychying:
## Why?
Stovepipe needs a controller that actually triggers builds using the BuildRunner contract.

## What?
Adds `stovepipe/controller/build`. `Process` unmarshals a `BuildRequest` (request id only), reloads the `Request`, no-ops if already terminal, resolves the runner via `Factory.For(Config{QueueName})`, reads the already-decided `BuildStrategy`/`BaseURI` (retryable if not yet visible — never re-derives full-vs-incremental itself), triggers via `buildRunner.Trigger(ctx, baseURI, request.URI, nil)`, persists a `Build` row (`BuildStore.Create`, tolerant of `ErrAlreadyExists`), and publishes the resulting build id to the `buildsignal` topic, partitioned by build id.

## Test Plan
`stovepipe/controller/build/build_test.go` covers: terminal-request no-op, unresolved build strategy (retryable), `Factory.For`/`Trigger`/`BuildStore.Create`/publish failure paths, and the happy path for both full and incremental strategies. `go build ./...` and `go test ./stovepipe/controller/build/...` pass.

### DIFF
--- a/stovepipe/controller/build/build.go
+++ b/stovepipe/controller/build/build.go
@@ -138,7 +138,7 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) (r
 		return fmt.Errorf("BuildController failed to publish build signal for %s: %w", build.ID, err)
 	}
 
-	c.logger.Infow("triggered build",
+	c.logger.Debugw("triggered build",
 		"request_id", request.ID,
 		"build_id", build.ID,
 		"queue", request.Queue,
--- a/stovepipe/controller/build/build.go
+++ b/stovepipe/controller/build/build.go
@@ -147,16 +147,13 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) (r
 	return nil
 }
 
-// loadRequest returns the request for id. A not-yet-visible row is retryable.
+// loadRequest returns the request for id.
 func (c *Controller) loadRequest(ctx context.Context, id string) (entity.Request, error) {
 	got, err := c.store.GetRequestStore().Get(ctx, id)
-	if err == nil {
-		return got, nil
+	if err != nil {
+		return entity.Request{}, fmt.Errorf("BuildController failed to load request %s: %w", id, err)
 	}
-	if errors.Is(err, storage.ErrNotFound) {
-		return entity.Request{}, errs.NewRetryableError(fmt.Errorf("request %s not found yet: %w", id, err))
-	}
-	return entity.Request{}, fmt.Errorf("BuildController failed to load request %s: %w", id, err)
+	return got, nil
 }
 
 // publishBuildSignal publishes buildID to the buildsignal stage, partitioned by
--- a/stovepipe/controller/build/build.go
+++ b/stovepipe/controller/build/build.go
@@ -111,7 +111,7 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) (r
 
 	// process decided the scope; build never re-derives incremental-vs-full.
 	if request.BuildStrategy == entity.BuildStrategyUnknown {
-		metrics.NamedCounter(c.metricsScope, _opName, "storage_errors", 1)
+		metrics.NamedCounter(c.metricsScope, _opName, "strategy_not_visible", 1)
 		return errs.NewRetryableError(fmt.Errorf("request %s has no build strategy yet", request.ID))
 	}
 	baseURI := ""
--- a/stovepipe/controller/build/build.go
+++ b/stovepipe/controller/build/build.go
@@ -119,7 +119,7 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) (r
 		baseURI = request.BaseURI
 	}
 
-	buildID, err := buildRunner.Trigger(ctx, request.URI, baseURI, nil)
+	buildID, err := buildRunner.Trigger(ctx, baseURI, request.URI, nil)
 	if err != nil {
 		return fmt.Errorf("BuildController failed to trigger build for request %s: %w", request.ID, err)
 	}
--- a/stovepipe/controller/build/build.go
+++ b/stovepipe/controller/build/build.go
@@ -1,0 +1,196 @@
+// Copyright (c) 2025 Uber Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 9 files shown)


stovepipe/controller/build/build.go
```diff
@@ -0,0 +1,196 @@
...
+
+	buildRunner, err := c.buildRunners.For(buildrunner.Config{QueueName: request.Queue})
+	if err != nil {
+		// A queue with no registered builder is a config error.
+		return fmt.Errorf("BuildController failed to resolve build runner for queue %s: %w", request.Queue, err)
+	}
+
+	// process decided the scope; build never re-derives incremental-vs-full.
+	if request.BuildStrategy == entity.BuildStrategyUnknown {
+		metrics.NamedCounter(c.metricsScope, _opName, "storage_errors", 1)
```
mnoah1: Give this its own error `strategy_not_visible` or similar

stovepipe/controller/build/build.go
```diff
@@ -0,0 +1,196 @@
...
+	if request.BuildStrategy == entity.BuildStrategyIncrementalSinceGreen {
+		baseURI = request.BaseURI
+	}
+
+	buildID, err := buildRunner.Trigger(ctx, baseURI, request.URI, nil)
+	if err != nil {
+		return fmt.Errorf("BuildController failed to trigger build for request %s: %w", request.ID, err)
+	}
+
+	build := entity.Build{
```
mnoah1, line 127: I guess it's ok for MVP, but in a future state where we might end up generating more than one build per request, we might want to persist the base URI that was used for that specific build.  For now it should be fine to get from the request as it's still 1:1.

mnoah1 (approved):

stovepipe/controller/build/build.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+	if err := c.store.GetBuildStore().Create(ctx, build); err != nil && !errors.Is(err, storage.ErrAlreadyExists) {
+		return fmt.Errorf("BuildController failed to persist build %s: %w", build.ID, err)
+	}
+
+	if err := c.publishBuildSignal(ctx, build.ID); err != nil {
+		return fmt.Errorf("BuildController failed to publish build signal for %s: %w", build.ID, err)
+	}
+
+	c.logger.Infow("triggered build",
```
behinddwalls: do we need this as Info log? or should it be debug?

stovepipe/controller/build/build.go
```diff
@@ -0,0 +1,196 @@
...
+
+// loadRequest returns the request for id. A not-yet-visible row is retryable.
+func (c *Controller) loadRequest(ctx context.Context, id string) (entity.Request, error) {
+	got, err := c.store.GetRequestStore().Get(ctx, id)
+	if err == nil {
+		return got, nil
+	}
+	if errors.Is(err, storage.ErrNotFound) {
+		return entity.Request{}, errs.NewRetryableError(fmt.Errorf("request %s not found yet: %w", id, err))
+	}
```
behinddwalls: why would request be not found?

stovepipe/controller/build/build.go
```diff
@@ -0,0 +1,196 @@
...
+	return entity.Request{}, fmt.Errorf("BuildController failed to load request %s: %w", id, err)
```
behinddwalls: do we need, BuildController market here?

behinddwalls: actually applicable to all the errors, can we remove the market?

stovepipe/controller/build/build.go
roychying: Could you elaborate on the scenario where one Request maps to multiple builds with different base URIs? Is this referring to the Phase 2 per-project builds? If so, will those share the same base URI as the Phase 1 build, since it's decided once per Request at admit?

Originally each BUILD has its own baseURI & headURI in storage, but was discarded per @behinddwalls 's comment: https://github.com/uber/submitqueue/pull/370#discussion_r3590381364

mnoah1: I guess I was thinking about if we implement any speculation/overlapping builds in the future but we can leave that decision for later.  For the current flow, the values from request should be good.

stovepipe/controller/build/build.go
roychying: it matches convention for one-shot per-build events elsewhere (in sq orchestrator build step). my preference is   reserving Debug for high-frequency logs like buildsignal's poll loop?

stovepipe/controller/build/build.go
roychying: Shouldn't happen in the common case as per we discussed in rfc (https://github.com/uber/submitqueue/pull/336#discussion_r3573363301). I kept it as a defensive guard but it's also making sense to remove here...

stovepipe/controller/build/build.go
roychying: also it's matching the existing convention across the codebase. happy to change it repo-wide if you want to revisit the convention, but that'd be a separate PR.

stovepipe/controller/build/build.go
roychying: good catch. renamed this one

stovepipe/controller/build/build.go
mnoah1: I replied on the other one - I think the defensive guard is ok here... it could be missing due to replication lag if this task hits a different replica than the one that inserted the row.

stovepipe/controller/build/build.go
mnoah1: Agree with keeping in Info for now, it would be nice to be able to scroll through all the logs of builds that got triggered and frequency of these shouldn't be unreasonably high.

behinddwalls: eventually it's a problem for ratining logs over time when they become too much due to all the info logs

stovepipe/controller/build/build.go
behinddwalls: i think we should not add that logic here, we likely going to pin to primary region and require a consistent data store for this

stovepipe/controller/build/build.go
behinddwalls: yes.. we should remove that..it's a noise

stovepipe/controller/build/build.go
roychying: ok then shall we do same change for submitqueue orchestrator build as well?

stovepipe/controller/build/build.go
roychying: ack. will raise another pr later for this issue

roychying: modified here: https://github.com/uber/submitqueue/pull/420

behinddwalls (approved):

behinddwalls:

The rebase of `chenghan.ying/stovepipe-build-controller` completed without conflicts, but the resulting code diff does not match the original. This means the rebase silently altered code content. The branch was **not** force-pushed.

PR #409, opened 2026-07-20
feat(stovepipe): add and wire the buildsignal stage controller

roychying:
## What?
Adds `stovepipe/controller/buildsignal` plus the new `Record` queue message/proto and `TopicKeyRecord`. `Process` consumes a `BuildSignal` (build id), loads `Build`+`Request`, no-ops if the request is terminal, polls `buildRunner.Status`, persists a real status transition via CAS (`BuildStore.Update`) with terminal status **write-once** (a later poll of a flaky backend can never overwrite an already-committed terminal status), and either reschedules itself via `PublishAfter` (5s while `accepted`, 2s while `running`) or, once terminal, publishes to `record`. Wires the controller into `main.go` and registers the `buildsignal` (subscribed) and `record` (publish-only, no consumer yet) topics.

## Test Plan
`make build`, `make test` - all clean.


stovepipe/controller/buildsignal/buildsignal.go
```diff
@@ -0,0 +1,285 @@
...
+// TODO: move these behind a queueconfig-style extension so operators can
+// tune poll cadence per queue without a code change.
+var (
+	// PollDelayAcceptedMs is the delay before the next Status call while the
+	// build is queued by the runner but has not started executing.
+	PollDelayAcceptedMs int64 = 5000
+	// PollDelayRunningMs is the delay before the next Status call while the
+	// build is executing.
+	PollDelayRunningMs int64 = 2000
+)
```
behinddwalls, line 52: i think this needs to be configurable and come via controller inputs eventially, can you add TODO to do so

stovepipe/controller/buildsignal/buildsignal.go
```diff
@@ -0,0 +1,285 @@
...
+	// The request is done (record already ran, or the head was superseded);
+	// stop polling.
+	if request.State.IsTerminal() {
+		return nil
+	}
+
+	buildRunner, err := c.buildRunners.For(buildrunner.Config{QueueName: request.Queue})
+	if err != nil {
+		// A queue with no registered builder is a config error.
+		return fmt.Errorf("BuildSignalController failed to resolve build runner for queue %s: %w", request.Queue, err)
```
behinddwalls, line 132: BuildSignalController marker? can we remove those?

stovepipe/controller/buildsignal/buildsignal.go
```diff
@@ -0,0 +1,285 @@
...
+	if errors.Is(err, storage.ErrNotFound) {
+		return entity.Build{}, errs.NewRetryableError(fmt.Errorf("build %s not found yet: %w", id, err))
+	}
+	return entity.Build{}, fmt.Errorf("BuildSignalController failed to load build %s: %w", id, err)
+}
+
+// loadRequest returns the request for id. A miss here is almost certainly a
+// lagging read behind the Build that referenced it, so it is retryable; a
+// genuinely orphaned Build still dead-letters at MaxAttempts.
+func (c *Controller) loadRequest(ctx context.Context, id string) (entity.Request, error) {
```
behinddwalls, line 202: i see it being duplicated between build and buildsignal

stovepipe/controller/buildsignal/buildsignal.go
```diff
@@ -0,0 +1,285 @@
...
+	got, err := c.store.GetRequestStore().Get(ctx, id)
+	if err == nil {
+		return got, nil
+	}
+	if errors.Is(err, storage.ErrNotFound) {
```
behinddwalls: same as other PR, why do we need to handle not found?

mnoah1: I've noticed agents tend to add this as a defense against DB replication lag, maybe won't happen much in the real world but feels reasonable to guard against.

stovepipe/controller/buildsignal/buildsignal.go
mnoah1: Maybe we should align on this more generally.  Do we want to make each controller own its own logic for fetching required entities, or have some kind of intermediate later - e.g. how we typically put all database / retrieval operations behind a repository package in our internal services and it just returns the entity.

mnoah1: It looks like in submitqueue most controllers have been handling it themselves so far but it could lead to duplication.

stovepipe/controller/buildsignal/buildsignal.go
roychying: same as this thread: https://github.com/uber/submitqueue/pull/407#discussion_r3623744558
happy to refactor all in another pr

stovepipe/controller/buildsignal/buildsignal.go
roychying: it's true. also discussed in build rfc pr: https://github.com/uber/submitqueue/pull/336#discussion_r3573363301. but it's also making sense to remove it as well.. will do

stovepipe/controller/buildsignal/buildsignal.go
roychying: will introduce a shared helper in stovepipe for all loadX pattern

roychying: Fixed

stovepipe/controller/buildsignal/buildsignal.go
roychying: Fixed

stovepipe/controller/buildsignal/buildsignal.go
roychying: Modified here: https://github.com/uber/submitqueue/pull/420

behinddwalls (approved):

PR #413, opened 2026-07-21
docs(rfc): queue-scoped speculation design

behinddwalls:
### Why?

A merge queue that verifies one change at a time is bounded by its slowest build. Speculation builds a batch early against a guess at how its unresolved dependencies will resolve, so a matching build is ready when reality settles while merging stays strict. Getting that right on an eventually consistent, transactionless store needs an explicit design for how a path is represented, which paths build under a bounded CI budget, what is persisted, and where each decision lives.

### What?

This RFC designs speculation for SubmitQueue end to end. Each queue change fires one run per queue: the speculate controller reads the current facts once, reconciles path state, refutes broken bets, finalizes verdicts, asks one pluggable extension which paths to build or preempt, and writes the result. Only chosen paths are persisted; every other possibility is recomputed each run.

The Speculator is the only controller-facing extension. It chooses which paths to build and which running ones to cancel, while verdicts and correctness cancels remain controller rules. The default Speculator internally composes a Generator, which produces a lazy best-first candidate stream, and an Allocator, which spends the CI budget and settles in-flight paths by ID; alternate Speculators need not preserve that split.

A path is self-describing: it carries an included, excluded, or dropped bet on every dependency. Its identity hashes the head plus its bets, build dispatches are replayed idempotently by path and attempt, and pending, building, and cancelling paths continue to charge the hard CI budget until terminal. The design also covers conflict relaxation and bypassing a large dependency once passed builds cover every possible resolution.

The RFC fills the previously empty `doc/rfc/submitqueue/speculation.md`.


sbalabanov (changes requested):
back to you based on brainstorming session

doc/rfc/submitqueue/speculation.md
```diff
@@ -0,0 +1,220 @@
+# Speculation
+
+A merge queue that verifies one change at a time is limited by its slowest build. Speculation builds batches early — before the conflicting batches ahead of them resolve — by guessing how those will resolve. So a valid build is often ready when its dependencies finish. Work enters SubmitQueue as **batches** (changes verified and merged together). Batches that touch the same code **conflict**, making earlier conflicting batches their **dependencies**. A speculative assumption about how those dependencies will resolve is called a **path**, and the batch being built is the path's **head**. On every queue update, the **speculate controller** recomputes everything: it reads the latest state, applies incoming signals, asks a pluggable **Speculator** which paths are worth building within the available CI budget, and persists only those selected paths. It recomputes all other possibilities on demand. Despite speculative execution, merging remains strict: a batch lands only after its dependencies have resolved and a matching speculative build has passed.
```
JamyDev: The short sentences read like this is a bulleted list that got collapsed. I think it would actually be more legible if most of these were facts about speculation (baring the first line)

doc/rfc/submitqueue/speculation.md
```diff
@@ -0,0 +1,220 @@
...
+
+### Finalization
+
+The speculate controller computes every **verdict**. A verdict is a batch's outcome. The facts fix it: there is one correct answer, and the Speculator can neither compute nor veto it:
+
+- **Merge (strict).** Each path carries a bet on every dependency — *included* (built on top of), *excluded* (built without), or *dropped* (relaxed away). Once a path's build has passed and every *included* dependency has merged, the head merges — it waits only on the dependencies it was built on top of, not the head's full dependency list. It moves the head to Merging and hands it to runway (the service that performs merges). The head is marked merged only when runway reports the merge actually happened, never optimistically. Down a chain, each head waits for its included predecessors, so a chain lands one at a time. (The runway merge is idempotent: re-sending a lost hand-off for an already-merged change is an empty merge runway skips while reporting success.)
```
JamyDev:
```suggestion
- **Merge (strict).** Each path carries a bet on every dependency — *included* (built on top of), *excluded* (built without), or *dropped* (relaxed away). Once a path's build has passed and every *included* dependency has merged, the head merges — it waits only on the dependencies it was built on top of, not the head's full dependency list. It moves the head to Merging and hands it to Runway (the service that performs merges). The head is marked merged only when Runway reports the merge actually happened, never optimistically. Down a chain, each head waits for its included predecessors, so a chain lands one at a time. (The runway merge is idempotent: re-sending a lost hand-off for an already-merged change is an empty merge Runway skips while reporting success.)
```

JamyDev: Additionally, what does "relaxed away" mean? Can we use simpler terms?

JamyDev: Ah I see, I had to use [Future Sight](https://pokemondb.net/move/future-sight)..

doc/rfc/submitqueue/speculation.md
```diff
@@ -0,0 +1,220 @@
...
+- **Cancel.** When a batch is cancelled, the controller cancels its in-flight paths and, once all have stopped, marks it Cancelled.
+
+Two smaller duties:
+
+- **Merging supervision.** If the hand-off to runway is lost, the next run re-sends it; the outcome returns via mergesignal.
+- **No dependent fanout.** A merge publishes one dirty signal. The next run recomputes every batch against the new reality — dependents that bet against the merged batch get cancelled, those that bet on it keep building.
+
+### Conflict relaxation
+
+Conflict analysis is conservative — it flags any *possible* conflict — so heads carry dependencies that rarely matter and over-serialize. Relaxation lets the Speculator **drop** the weakest: the path tags that dependency *dropped* and ignores it — it neither gates the merge nor refutes the path if it lands. Fewer bets, shorter chains, more parallelism. Which to drop is a per-run Speculator policy.
```
JamyDev:
```suggestion
Conflict analysis is conservative — it flags any *possible* conflict — so heads carry dependencies that rarely matter and over-serialize. Relaxation lets the Speculator **drop** the weakest: the path tags that dependency *dropped* and ignores it — it neither gates the merge nor refutes the path if it lands. Which to drop is a per-run Speculator policy.
```

drop the fluff

doc/rfc/submitqueue/speculation.md
```diff
@@ -0,0 +1,220 @@
...
+
+The drop lives on the path, so the path stays self-describing: finalization needs no external relaxed set, and dropped dependencies don't count toward the depth bound (relaxing is what shrinks a head's depth).
+
+Example: `H` conflicts with `B1` and weak `B2`. Drop `B2`, and `H` merges once `B1` lands and its build passes — even if `B2` later lands. Without it, `H` waits on both.
+
+### Bypass large diff
+
+If a batch's passed builds cover *every* way its dependencies could resolve, the outcome is the same either way — so it can merge now, ahead of them. Classic case: a small change stuck behind a slow one is built both with and without it; both pass, and it merges immediately.
+
+The Generator decides when covering the whole space is cheap enough and funds those builds; the controller merges once a *passed* path exists for every combination, read straight off the records. Distinct bets have distinct identities, so a cover can't be faked, and one that isn't complete just falls back to the normal wait.
```
JamyDev: > Distinct bets have distinct identities, so a cover can't be faked, and one that isn't complete just falls back to the normal wait.

I've read this sentence 5 times, I know from context clues what you're probably trying to say, but this sentence doesn't say it..

doc/rfc/submitqueue/speculation.md
```diff
@@ -0,0 +1,220 @@
...
+
+## Extensions
+
+**One extension: the Speculator.** It decides *which paths to build and which running ones to cancel* — nothing more. Everything else stays controller code: reconciling facts, cancelling ruled-out paths, the verdicts, checking the output. Swapping the Speculator changes which paths run; it can never change whether a batch merges or fails.
```
JamyDev:
```suggestion
## Speculator Extension

Decides *which paths to build and which running ones to cancel*. Everything else stays controller code: reconciling facts, cancelling ruled-out paths, the verdicts, checking the output. Swapping the Speculator changes which paths run; it can never change whether a batch merges or fails.
```

Could also just be written as:

```
Single responsibility, calculate which paths to build, and which builds to cancel. Controller handles the rest (reconciling facts, cancelling ruled-out paths, the verdicts, checking the output)
```

doc/rfc/submitqueue/speculation.md
```diff
@@ -0,0 +1,220 @@
...
+
+**The contract** is `Speculate(batches, pathSets) → []Speculation`. In: the queue's live batches (each with its prior score, dependency list, and state) and every path set for those batches — live and recently finished — so the Speculator sees which paths already ran and won't re-propose a failed or passed one. Out: a list of build/cancel actions, nothing else. Budget, depth bound, and clock are injected when the Speculator is built. An impl may read extra data (also injected); the controller checks the output, so extra data never affects correctness.
```
JamyDev: Can you make this a list of: 


```suggestion
**The contract** is `Speculate(batches, pathSets) → []Speculation`. 
- In: 
  - The queue's live batches (each with its prior score, dependency list, and state) E
  - Every path set for those batches — live and recently finished — so the Speculator sees which paths already ran and won't re-propose a failed or passed one. 
- Out: a list of build/cancel actions. 
- Budget, depth bound, and clock are injected when the Speculator is built. An impl may read extra data (also injected); the controller checks the output, so extra data never affects correctness.
```

Not perfect, but you get the gist

doc/rfc/submitqueue/speculation.md
```diff
@@ -0,0 +1,220 @@
...
+
+### The default Speculator
+
+Composed in wiring from two swappable interfaces — swap the whole Speculator for a joint optimizer or learned model, or tune behavior by replacing just the Generator or Allocator. Its entire body: open the candidate stream over the batches and their path sets, then hand the stream and the path sets to the Allocator.
```
JamyDev: > swap the whole Speculator for a joint optimizer or learned model, or tune behavior by replacing just the Generator or Allocator

I feel like this is unnecessary, the extension model implies this, which begs the question: why are generator and allocator not extensions by themselves? Don't we intend to try and swap those out?

doc/rfc/submitqueue/speculation.md
```diff
@@ -0,0 +1,220 @@
...
+
+- **Generator** — the queue-wide ordered source of candidate paths. Given the batches and their path sets, it returns a single iterator that yields coherent, depth-capped candidates best-first across every head. Scoring and depth capping are internal. It prices each candidate as the product of its dependencies' landing chances: a dependency's chance starts at its `Batch.Score` (forced to 1 or 0 once resolved). Build evidence from its path set then sharpens it: a dependency whose speculative build has passed is likelier to land. It skips candidate paths already terminal in the path sets (passed, failed, cancelled), so best-first never re-emits a dead future. It caps **depth** — the number of unresolved dependencies a head's paths range over — so one deep chain can't blow up a run. A head beyond the cap is skipped this run and enters as its dependencies resolve.
```
JamyDev: This also just reads like either a list of requirements, or maybe an implementation. 

If I implement my own generator, am I expected to conform to all of these? Are some of these more flexible?

doc/rfc/submitqueue/speculation.md
```diff
@@ -0,0 +1,220 @@
...
+- **Allocator** — spends the build budget (the queue's cap on concurrent builds) over the single iterator. It pulls candidates in order until the budget fills. It matches each against the path sets by ID: a path already building is kept, not re-triggered. Building and cancelling paths charge the budget (a cancelling build holds CI until observed terminal); passed, failed, and cancelled paths charge none. The sticky default fills only free slots and leaves every viable in-flight build running; the preempting default cancels in-flight paths that fall below the funded set. Budget is the only rationing lever: there is no score floor, so the default spends every slot on the best remaining candidates, however long the shot.
```
JamyDev: > Building and cancelling paths charge the budget (a cancelling build holds CI until observed terminal)

This is interesting.. let's say I have a 15 build budget, 14 ongoing. I speculate, and note that I can cancel 10 builds. Given that there is only 1 build left in the budget, that means I can't return more than 1 build to start, even though I know the budget will be there shortly (given I'm canceling 10 of em). Seems suboptimal that we'd have to re-run the speculator to make use of that budget?

doc/rfc/submitqueue/speculation.md
```diff
@@ -0,0 +1,220 @@
...
+ CI runs; the result lands back at buildsignal — the loop closes
+```
+
+### State reconciliation
+
+The controller reads all live batches, path sets, executions, and builds once, never re-reading mid-run. It sets each path's status from its build result and refutes — cancels — any path whose bet a resolved dependency has broken: an included dependency that failed, or an excluded one that landed. Every write is a compare-and-swap. A lost race replans that head on a later run. No run depends on an earlier one, so trigger messages may be lost, duplicated, or reordered without harm.
+
+### Speculator
+
+The controller calls the one extension, `Speculator.Speculate(batches, path sets)`, which returns paths to build and in-flight paths to preempt — nothing else. The controller validates that result, dropping any refuted, incoherent, or already-failed build.
```
JamyDev: Can we document an rough time-deadline for the Speculate call? Our controller should implement a bailout if a speculate call is taking too long (and effectively misbehaving).

behinddwalls: I would prefer not to enforce a Speculator-specific deadline in the controller. Like our other extension calls, the controller passes its context and the implementation is responsible for honoring cancellation and keeping the call within its operational budget. That leaves each implementation free to make the right latency/quality tradeoff instead of baking an arbitrary timeout into the contract; we can instrument duration and optimize or bound a concrete implementation if it becomes necessary.

doc/rfc/submitqueue/speculation.md
behinddwalls: They are swappable, but intentionally remain composition points inside the default Speculator rather than controller-facing extensions. The controller depends only on the Speculator contract, allowing an alternative joint optimizer to replace it without preserving the Generator/Allocator split. I clarified this in the RFC.

doc/rfc/submitqueue/speculation.md
behinddwalls: Yes, this intentionally requires another run. Cancellation is best-effort and the CI slot remains occupied until terminal; allocating against expected capacity could exceed the hard cap if cancellation is slow or loses a race. The terminal result publishes dirty, so the next run fills the released slots. I clarified this in the RFC.

doc/rfc/submitqueue/speculation.md
JamyDev: I would still think it's good to give ballpark guidance (not enforced)? While it obviously depends on expected queue size; 1 min / 100 batches is probably too slow, whereas 100ms / 100 batches is too aggressive.

behinddwalls: i think would leave it as is for now on this... to be honest, the current V1 tick runs every 10s or so a minute to run wouldn't be that bad but yes...it needs to meet the demand of the queue.

JamyDev (approved):

doc/rfc/submitqueue/speculation.md
```diff
@@ -0,0 +1,209 @@
...
+
+Work enters SubmitQueue as **batches** — changes verified and merged together. Two batches **conflict** when they touch the same code, which makes the earlier one a **dependency** of the later. A **path** is one guess at how a batch's dependencies resolve, and the batch it builds is the path's **head**.
+
+On every queue update the **speculate controller** reruns from scratch: it reads the current state, applies the incoming signals, asks a pluggable **Speculator** which paths are worth building within the CI budget, and persists only those. Everything else is recomputed next time, never stored. Merging stays strict: a batch lands only after its dependencies resolve and a matching build has passed.
+
+## The speculation run
+
+The speculate controller runs whenever the queue changes — after a new batch, a completed build, a completed merge, a cancel, or DLQ recovery. Each publishes a **dirty signal** carrying the changed batch ID, partitioned by the queue so a queue's runs happen one at a time. The dirty signal is an internal queue contract — payload in `submitqueue/core/messagequeue`, topic key in `submitqueue/core/topickey`.
```
sbalabanov: DLQ never recovers, only reconciles to "failed"
even after this, DLQ does not signal anything to non-DLQ controller (messaging layers isolation).

doc/rfc/submitqueue/speculation.md
```diff
@@ -0,0 +1,209 @@
...
+   - a merge completes
+   - a cancel or DLQ recovery
+   │
+   │  carries the changed batch ID, partitioned by queue, so a
+   │  queue's runs happen one at a time
+   ▼
+ speculate  (orchestrator/controller/speculate) — all controller code,
+ │          except step 3 (the one call out to the extension)
+ │
+ │ 1 read   the queue's live batches, their path sets, executions, builds
```
sbalabanov: path sets and builds are probably the same thing?

doc/rfc/submitqueue/speculation.md
```diff
@@ -0,0 +1,209 @@
...
+
+## The speculation run
+
+The speculate controller runs whenever the queue changes — after a new batch, a completed build, a completed merge, a cancel, or DLQ recovery. Each publishes a **dirty signal** carrying the changed batch ID, partitioned by the queue so a queue's runs happen one at a time. The dirty signal is an internal queue contract — payload in `submitqueue/core/messagequeue`, topic key in `submitqueue/core/topickey`.
+
+```
+ dirty(queue) — "trigger a run" — published after:
+   - a new batch
+   - a build completes (success/failure/cancellation)
+   - a merge completes
```
sbalabanov: can also be success/failure/cancel

doc/rfc/submitqueue/speculation.md
```diff
@@ -0,0 +1,209 @@
...
+
+The drop lives on the path, so the path stays self-describing: finalization needs no external relaxed set, and dropped dependencies don't count toward the depth bound (relaxing is what shrinks a head's depth).
+
+Example: `H` conflicts with `B1` and weak `B2`. Drop `B2`, and `H` merges once `B1` lands and its build passes — even if `B2` later lands. Without it, `H` waits on both.
+
+### Bypass large diff
+
+If a batch's passed builds cover *every* way its dependencies could resolve, the outcome is the same either way — so it can merge now, ahead of them. Classic case: a small change stuck behind a slow one is built both with and without it; both pass, and it merges immediately.
+
+The Generator decides when covering the whole space is cheap enough and enumerates the extra candidates; the Allocator funds them within the build budget. The controller merges early only when a passed path exists for every combination of the dependencies — it reads that straight off the path records. If any combination is missing or unbuilt, the head waits normally.
```
sbalabanov: Generator is new term on a doc and comes up without context. May be give a TLDR of the design a single paragraph above.

doc/rfc/submitqueue/speculation.md
```diff
@@ -0,0 +1,209 @@
...
+	Attempt   int                   // build attempt number, >= 1; builds and messages key on (ID, Attempt)
+	Version   int                   // for compare-and-swap writes
+	CreatedAt int64                 // ms
+	UpdatedAt int64                 // ms
+}
+
+// SpeculationPath is a head plus a bet on each of its dependencies; every
+// dependency appears exactly once, in queue order, so the path is self-describing.
+type SpeculationPath struct {
+	Head string          // the batch being built
```
sbalabanov: batch ID ?

doc/rfc/submitqueue/speculation.md
```diff
@@ -0,0 +1,209 @@
...
+	Bets []DependencyBet // one bet per dependency of Head, in queue order
```
sbalabanov: doesn't it duplicate SpeculationPath ?

doc/rfc/submitqueue/speculation.md
```diff
@@ -0,0 +1,209 @@
...
+type DependencyBet struct {
+	Batch string            // the dependency batch ID
+	Bet   DependencyBetType // included | excluded | dropped
+}
+
+// DependencyBetType is how a path treats one dependency.
+type DependencyBetType int
+
+const (
+	BetUnknown  DependencyBetType = iota // zero-value sentinel; never valid
```
sbalabanov: give it zero explicitly

doc/rfc/submitqueue/speculation.md
```diff
@@ -0,0 +1,209 @@
...
+// base-3 code the depth bound keeps to a small integer.
+type SpeculationPathSet struct {
+	BatchID string                 // primary key; the head batch
+	Paths   []SpeculationPathEntry // the head's chosen paths
+	Version int                    // for compare-and-swap writes
+}
+```
+
+```go
+// Speculator — the one extension the controller calls. It can never express a verdict.
```
sbalabanov: "It can never express a verdict." - what is this about?

doc/rfc/submitqueue/speculation.md
```diff
@@ -0,0 +1,209 @@
...
+type Speculator interface {
+	Speculate(ctx context.Context, batches []entity.Batch, pathSets []entity.SpeculationPathSet) ([]Speculation, error)
```
sbalabanov, line 167: in real implementations, have to explain the inputs. i.e. batches[] are everything in flight or finalized before deps
pathSets are everything materialized so far wrt batches[], completed or not

doc/rfc/submitqueue/speculation.md
```diff
@@ -0,0 +1,209 @@
...
+	// Next yields the next-best candidate across all heads; ok = false means the
+	// queue's coherent, depth-capped space is exhausted. Candidates descend in
+	// score, never repeat, and never contradict a resolved fact.
+	Next(ctx context.Context) (c CandidatePath, ok bool, err error)
+}
+
+// CandidatePath is a path plus its transient ordering score for this run.
+type CandidatePath struct {
+	Path  entity.SpeculationPath // head + one bet per dependency
+	Score float64                // the ordering score; transient, carried within the run,
```
sbalabanov: typical term is "ranking score" or just "rank"

sbalabanov (approved):

doc/rfc/submitqueue/speculation.md
behinddwalls: It just captures all deps, regardless of dropped or being excluded in a path so finalization and bypassing logic can make those decisions based on the bet itself.. as when we relax, Batch.Deps would be different from what Bet gets

PR #414, opened 2026-07-21
docs(rfc): propose shared platform errors

albertywu:
## Summary
Define stable platform error identities for conditions with common semantics across domains. Keep error identity separate from retry classification, use platform errors directly rather than through domain aliases, and illustrate the distinction with ErrVersionMismatch and ErrNotFound.

## Stack
1. @ #414
1. #421

### DIFF
--- a/doc/rfc/index.md
+++ b/doc/rfc/index.md
@@ -8,6 +8,7 @@ Design documents and technical proposals, grouped by scope. Shared/cross-cutting
 - [Message Queue Contract](messagequeue-contract.md) - How queue payloads are defined (Protobuf, serialized as protobuf JSON), located by audience (external in `api/{domain}/messagequeue/`, internal in `{domain}/core/messagequeue/`), bound to topics (the `topics` proto option), and enforced by Bazel visibility
 - [Consumer Gate](consumer-gate.md) - Stopping and starting individual queue controllers at runtime via consumer middleware: parked deliveries held in-flight with visibility extension, gate state as a separate extension with a file-based first implementation shared by tests and operators
 - [Change URIs](change-uri.md) - Identity of a code change: `scheme://{host[:port]}/{path}` per provider (GitHub PR, Phabricator Diff, git ref/commit) and canonical-form rules
+- [Shared Platform Errors](shared-platform-errors.md) - Cross-domain error identities with common classification only where recovery behavior is universal
 
 ## SubmitQueue
 
--- a/doc/rfc/shared-platform-errors.md
+++ b/doc/rfc/shared-platform-errors.md
@@ -1,0 +1,50 @@
+# Shared Platform Errors
+
+Define shared error identities for conditions that have the same meaning across domains. A shared identity enables consistent `errors.Is` checks and, where appropriate, a common platform classification. Sharing an error identity does not by itself make the error retryable.
+
+## Problem
+
+SubmitQueue and Stovepipe define duplicate sentinels for common conditions such as a missing resource and an optimistic version conflict. Callers must know which domain package produced the error, and platform code cannot apply a common policy without importing domain packages or adding domain-specific classifiers.
+
+| Concern | Current | Proposed |
+|---|---|---|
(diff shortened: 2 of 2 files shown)


doc/rfc/error-classification.md
```diff
@@ -0,0 +1,98 @@
...
+| Consumer and transport | Apply the verdict through ack, nack, reject, retry limits, and DLQ policy |
+
+This separation puts retry classification in the common runtime because that runtime sees every returned error and already owns the transport action. Controllers remain free to interpret a conflict in workflow terms without duplicating transport policy.
+
+### Optimistic-conflict marker
+
+`platform/errs` defines a semantic marker:
+
+```go
+type OptimisticConflict interface {
```
behinddwalls: who should return this error, storage impl or controller?

doc/rfc/error-classification.md
```diff
@@ -0,0 +1,98 @@
...
+    return fmt.Errorf("update batch %s: %w", batch.ID, err)
+}
+```
+
+The classifier marks the conflict retryable, the consumer nacks the delivery, and redelivery reloads durable state before making another decision.
+
+A controller may still handle the conflict locally when it has workflow knowledge the platform lacks. For example, if another writer already committed the desired terminal transition, the controller can acknowledge successful convergence:
+
+```go
+if errors.Is(err, storage.ErrVersionMismatch) && desiredStateAlreadyCommitted {
```
behinddwalls: should controller us this vs OptimisticConflict error for checking?

doc/rfc/error-classification.md
```diff
@@ -0,0 +1,98 @@
...
+The marker describes the cause, not the action. It does not itself mean nack, choose a delay, or configure a retry limit. The platform classifier owns the policy that an unhandled optimistic conflict should be retried from fresh state.
+
+SubmitQueue and Stovepipe retain their own `storage.ErrVersionMismatch` values, implemented as typed errors satisfying this marker. Storage callers can continue using `errors.Is(err, storage.ErrVersionMismatch)`, and platform code does not import either storage package.
+
+### Generic classification
+
+The existing `genericerrs.Classifier` recognizes the marker while continuing to classify backend-independent errors such as `context.Canceled`:
+
+```text
+OptimisticConflict -> InfraRetryable
```
behinddwalls: we also have things like `storage.ErrNotFound`?

albertywu: Yes, but `ErrNotFound` and `ErrAlreadyExists` are call-site dependent, so they should not receive one global verdict. Added clarification that platform markers are for errors with a stable cross-domain recovery policy.

doc/rfc/error-classification.md
albertywu: The storage implementation returns its domain-specific `storage.ErrVersionMismatch`; the concrete error type implements the platform marker. Controllers do not construct `OptimisticConflict`. Clarified this ownership in the RFC.

doc/rfc/error-classification.md
albertywu: Controllers should check their extension-specific `storage.ErrVersionMismatch` for workflow handling. `OptimisticConflict` is used by the platform classifier for cross-domain classification. Clarified this distinction in the RFC.

doc/rfc/error-classification.md
behinddwalls: when we say concrete error type? what that would be `storage.ErrVersionMismatch` error type implementing this marker or some other error...

classifier would need to know the error type? https://github.com/uber/submitqueue/blob/main/platform/errs/generic/generic.go#L52? OptimisticConflict would be an error type in that case? right

albertywu:
After exploring a bit I'm coming back in favor of status quo -- classifying errors as retry-able in the controller logic itself vs relying on platform alternatives and duplicating classifiers across services.

Closing out these PRs, don't think it's necessary.

Status quo pattern: storage / extension layers return domain-specific errors, and controller logic classifies as retry-able when returning errors.

PR #421, opened 2026-07-21
feat(errs): share cross-domain error identities

albertywu:
## Summary
Define ErrNotFound and ErrVersionMismatch in platform/errs for conditions with stable semantics across domains. Migrate SubmitQueue and Stovepipe storage and queue configuration code to use the platform errors directly, and keep only ErrVersionMismatch automatically classified as retryable.

## Test Plan
- make fmt
- make build
- make test
- make lint
- make check-tidy
- make check-gazelle

## Revert Plan
Revert this commit.

## API Changes
Remove domain ErrNotFound and ErrVersionMismatch sentinels plus storage IsNotFound and WrapNotFound helpers. Callers use platform/errs directly.

## Monitoring and Alerts
N/A

## Stack
1. #414
1. @ #421


albertywu:
After exploring a bit I'm coming back in favor of status quo -- classifying errors as retry-able in the controller logic itself vs relying on platform alternatives and duplicating classifiers across services.

Closing out these PRs, don't think it's necessary.

Status quo pattern: storage / extension layers return domain-specific errors, and controller logic classifies as retry-able when returning errors.

PR #423, opened 2026-07-22
feat(buildrunner extension): add shared platform buildkite/githubactions BuildRunner clients

roychying:
## Summary
- Extract shared Buildkite and GitHub Actions clients into `platform/extension/buildrunner/` and refactor the existing submitqueue adapters to use them.
- Add stovepipe-side Buildkite and GitHub Actions `BuildRunner` adapters on top of the shared clients.

## Test plan
- [x] `go build ./...`
- [x] `go test ./platform/extension/buildrunner/... ./submitqueue/extension/buildrunner/... ./stovepipe/extension/buildrunner/...`
- [x] `make gazelle` (no drift)


platform/extension/buildrunner/buildkite/client.go
```diff
@@ -0,0 +1,237 @@
...
+	_, _ = io.Copy(io.Discard, resp.Body)
+
+	switch resp.StatusCode {
+	case http.StatusOK:
+		return nil
+	case http.StatusUnprocessableEntity:
+		// Already terminal — no-op per BuildRunner.Cancel contract.
+		return nil
+	default:
+		return fmt.Errorf("unexpected status %d from cancel", resp.StatusCode)
```
sbalabanov: would be nice to also include the message into the error, typically in a body of a response; can be truncated if needed.

platform/extension/buildrunner/buildkite/client.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+	c.setHeaders(req)
+
+	resp, err := c.httpClient.Do(req)
+	if err != nil {
+		return fmt.Errorf("send request: %w", err)
+	}
+	defer resp.Body.Close()
+
+	respBody, err := io.ReadAll(resp.Body)
```
sbalabanov: this may not properly respond to context cancellation. AI analysis:

> Behavior: Vulnerable to blocking/hanging.

> Why: io.ReadAll reads from resp.Body until it hits io.EOF. However, once httpClient.Do() successfully receives response headers, the HTTP transport hands the raw stream body to you. Standard response bodies (net/http.bodyEOFSignal or TCP socket readers) do not automatically monitor the request ctx for subsequent reads.
> 
> If the server stalls, sends data infinitely slowly, or hangs while streaming the payload after headers have been sent, calling io.ReadAll(resp.Body) will block the goroutine until the connection times out at the network level—even if your ctx was canceled long ago.

The poor man's alternative is buffered read with cancellation checks:
https://github.com/uber/tango/blob/1ed2b86888fc4afc59494d9d97fdd6dc23c2638e/core/storage/ctxreader.go#L25
At least it cancels somewhere in between for long transfers.

platform/extension/buildrunner/buildkite/client.go
```diff
@@ -0,0 +1,237 @@
...
+	if err != nil {
+		return fmt.Errorf("read response: %w", err)
+	}
+
+	if resp.StatusCode == http.StatusNotFound {
+		return fmt.Errorf("build not found")
```
sbalabanov: http helper function do() is not necessarily about build. You may want to define sentinel errors to transform http codes into.

platform/extension/buildrunner/buildkite/client.go
```diff
@@ -0,0 +1,237 @@
...
+// without any local state. Returns an empty non-nil map when the key is
+// absent or its value cannot be decoded — a corrupt env var must not fail the
+// caller.
+func DecodeMetadataEnv(env map[string]string, key string) map[string]string {
+	meta := make(map[string]string)
+	raw, ok := env[key]
+	if !ok || raw == "" {
+		return meta
+	}
+	_ = json.Unmarshal([]byte(raw), &meta)
```
sbalabanov: let's not swallow errors without proper visibility
because this is an exported function, probably better design is to expose an error and let caller handle it as they see fit

platform/extension/buildrunner/buildkite/client_test.go
```diff
@@ -0,0 +1,218 @@
...
+	require.NoError(t, err)
+	return NewClient(c)
+}
+
+func buildJSON(number int, state, webURL string) []byte {
+	return buildJSONWithEnv(number, state, webURL, nil)
+}
+
+func buildJSONWithEnv(number int, state, webURL string, env map[string]string) []byte {
+	b, _ := json.Marshal(BuildResponse{Number: number, State: state, WebURL: webURL, Env: env})
```
sbalabanov:
```
func buildJSONWithEnv(t *testing.T, number int, state, webURL string, env map[string]string) []byte {
  t.Helper()
  b, err := json.Marshal(BuildResponse{Number: number, State: state, WebURL: webURL, Env: env})
  require.NoError(t, err)
  return b
}
```

platform/extension/buildrunner/buildkite/client_test.go
```diff
@@ -0,0 +1,218 @@
...
+// --- CreateBuild ---
+
+func TestCreateBuild_SubmitsPayloadAndReturnsResponse(t *testing.T) {
+	var capturedMethod string
+	var capturedBody []byte
+	c := newTestClient(t, http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		capturedMethod = req.Method
+		capturedBody, _ = io.ReadAll(req.Body)
+		w.Header().Set("Content-Type", "application/json")
+		_, _ = w.Write(buildJSON(42, "scheduled", "https://buildkite.com/test-org/my-pipeline/builds/42"))
```
sbalabanov: require.NoError()

platform/extension/buildrunner/githubactions/client.go
```diff
@@ -0,0 +1,308 @@
...
+func (c *Client) runPath(runID int64) string {
+	return fmt.Sprintf(
+		"/repos/%s/%s/actions/runs/%s",
+		url.PathEscape(c.owner),
+		url.PathEscape(c.repo),
+		url.PathEscape(strconv.FormatInt(runID, 10)),
+	)
+}
+
+func (c *Client) do(ctx context.Context, method, rawURL string, body []byte, out any) error {
```
sbalabanov, line 178: duplicated code with buildkite impl, can deduce?

stovepipe/extension/buildrunner/buildkite/buildkite.go
```diff
@@ -0,0 +1,196 @@
...
+
+// NewBuildRunner constructs a Buildkite-backed BuildRunner bound to a single
+// pipeline.
+func NewBuildRunner(params Params) (buildrunner.BuildRunner, error) {
+	if params.Client == nil {
+		return nil, fmt.Errorf("buildkite client is required")
+	}
+	if params.Logger == nil {
+		return nil, fmt.Errorf("logger is required")
+	}
```
sbalabanov: Submitqueue design assumes all wiring to be non-nil, so there is no need to explicitly check it. Panic will do just fine.

stovepipe/extension/buildrunner/buildkite/buildkite.go
```diff
@@ -0,0 +1,196 @@
...
+// Buildkite build number as the build ID. Errors are propagated to the
+// caller so the queue consumer can nack and retry.
+func (r *runner) Trigger(ctx context.Context, baseURI, headURI string, metadata entity.BuildMetadata) (entity.BuildID, error) {
+	env := map[string]string{
+		EnvKeyHeadURI: headURI,
+		EnvKeyBaseURI: baseURI,
+		EnvKeyQueue:   r.cfg.QueueName,
+	}
+	if len(metadata) > 0 {
+		metaJSON, _ := json.Marshal(metadata)
```
sbalabanov: plz do not swallow errors

stovepipe/extension/buildrunner/githubactions/githubactions_test.go
```diff
@@ -0,0 +1,254 @@
...
+}
+
+func TestTrigger_ErrorsWhenDispatchResponseHasNoRunID(t *testing.T) {
+	r := newTestRunner(t, http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		_ = json.NewEncoder(w).Encode(platformgithubactions.DispatchWorkflowResponse{})
+	}))
+
+	_, err := r.Trigger(context.Background(), "", "github://repo/head/aaa", nil)
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "response missing workflow_run_id")
```
sbalabanov: Do not assume on error messages - messages are not part of application logic and it makes test unnecessary fragile. AGENTS.md has an instruction to avoid it, surprising it was not picked up.

sbalabanov (changes requested):

platform/extension/buildrunner/buildkite/client.go
roychying: good call. dug a bit on this one. Per the [_stdlib_ docs](https://pkg.go.dev/net/http#Request.WithContext), `Request`'s context _"controls the entire lifetime of a request and its response: obtaining a connection, sending the request, and reading the response headers and body"_ , and [`Client.Timeout`](https://pkg.go.dev/net/http#Client) (implemented as context cancellation internally) is documented to _"interrupt reading of the Response.Body"_ . So I believe a canceled `ctx` does unblock an in-progress `io.ReadAll(resp.Body)` here, not wait for a network-level timeout.

one thing is that nothing sets a deadline on `ctx` or a `Timeout` on the `http.Client` for this path.  `platform/http.NewClient`'s doc already calls that out as the caller's job, but no actual service wiring exists yet to enforce it (Buildkite isn't wired into any `main.go` today).

I prefer to leave that to whichever change wires up a real client, but happy to follow tango's implementation since it's a cheap safe guard. lmk if anything is wrong on my understanding

platform/extension/buildrunner/buildkite/client.go
roychying: done, already included the body in the cancel error

platform/extension/buildrunner/buildkite/client.go
roychying: Done, do() now returns ErrNotFound instead of a hardcoded string.

platform/extension/buildrunner/buildkite/client_test.go
roychying: done

platform/extension/buildrunner/buildkite/client_test.go
roychying: done

platform/extension/buildrunner/githubactions/client.go
roychying: sure, now both use `extracted platform/http.SendRequest`

stovepipe/extension/buildrunner/buildkite/buildkite.go
roychying: I see, removed those unnecessary check

stovepipe/extension/buildrunner/buildkite/buildkite.go
roychying: the metadata marshal error is now propagated. Also fixed the same swallowed-marshal-error pattern in submitqueue's buildkite `Trigger`

stovepipe/extension/buildrunner/githubactions/githubactions_test.go
roychying: dropped the message assertion, kept `require.Error`.

stovepipe/extension/buildrunner/buildkite/buildkite.go
roychying: Also applied the same fix to submitqueue's `buildkite` and `githubaction`'s `NewBuildRunners`, which had identical checks.

mnoah1 (approved):

sbalabanov (approved):

platform/extension/buildrunner/buildkite/client.go
sbalabanov: Confirmed experimentally: canceling the request context interrupts an in-progress response-body read on this client path. The proof test waits until `io.ReadAll` has entered `Response.Body.Read`, cancels the context, and asserts that the call returns an error matching `context.Canceled`.

Proof-only PR (intentionally closed; do not review or merge): https://github.com/uber/submitqueue/pull/510

This works because `platform/http.BaseURLTransport` preserves the request context and delegates to `net/http.DefaultTransport`. A custom `RoundTripper` that ignores the context would remain a caveat, and a timeout/deadline is still needed if the caller never cancels.

PR #432, opened 2026-07-23
test: remove hardcoded queue integration timeout and parallelize workers

sbalabanov:
Improve Integration Test Runtime
Remove hardcoded timeout in lieu of Bazel's timeout

## Summary

- coordinate concurrent MySQL message-queue test receivers through one shared result channel
- keep message aggregation and duplicate tracking exclusively on the main test goroutine, without mutexes or a wait group
- acknowledge each delivery before reporting it, cancel on successful completion or the first real worker error, and consume every worker's terminal result before returning
- distinguish expected context cancellation after completion/error from standalone cancellation that must fail the test
- remove local integration-test safety timeouts and rely on Bazel's test timeout for stuck tests

## Benchmark

Uncached Bazel test attempt runtime:

- before: 58.3s
- after: 38.3s
- improvement: 20.0s (~34%)

## Tests

- `./tool/bazel test //test/integration/extension/messagequeue/mysql:go_default_test --nocache_test_results --test_output=errors --strategy=TestRunner=local ...`
- `./tool/bazel test //test/integration/extension/messagequeue/mysql:go_default_test --nocache_test_results --test_output=errors --strategy=TestRunner=local --@rules_go//go/config:race ...`
- `make fmt`
- `make lint`
- `make check-tidy`
- `make check-gazelle`

### DIFF
--- a/test/integration/extension/messagequeue/mysql/queue_test.go
+++ b/test/integration/extension/messagequeue/mysql/queue_test.go
@@ -17,9 +17,9 @@ package mysql
 import (
 	"context"
 	"database/sql"
+	"errors"
 	"fmt"
 	"strconv"
-	"sync"
 	"testing"
 	"time"
 
@@ -133,6 +133,13 @@ func receiveN(
 	}
 }
 
+type concurrentReceiveResult struct {
+	workerID int
+	message  string
+	err      error
+	done     bool
+}
+
 // receiveConcurrently collects deliveries across workers until the expected
 // number of unique messages has been acknowledged. Completion or a worker
 // error cancels every receiver immediately.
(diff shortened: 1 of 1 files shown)


test/integration/extension/messagequeue/mysql/queue_test.go
sbalabanov: remove hardcoded timeout altogether. Let the test rely on upstream timeout setting (i.e. Bazel's) instead. If needed, wait on channel or poll indefinitely.

sbalabanov: Addressed in c5c0fbb. I removed the file-level `testTimeout` and all local `time.After`/`context.WithTimeout` safety waits from this integration suite. Channel and signal waits now rely on Bazel's test timeout. The concurrent receiver helper still uses event-driven context cancellation for successful completion and worker errors, and propagates channel-close and ack failures to the test goroutine. The focused target and race-detector run both pass.

test/integration/extension/messagequeue/mysql/queue_test.go
```diff
@@ -114,40 +110,95 @@ func testSubConfig(subscriberName, consumerGroup string) extqueue.SubscriptionCo
...
-			require.Fail(t, "Timeout waiting for message", "%d/%d after %v", i+1, count, testTimeout)
-		}
+		handler(receive(t, deliveryChan), i)
 	}
 }
 
+// receiveConcurrently collects deliveries across workers until the expected
+// number of unique messages has been acknowledged. Completion or a worker
+// error cancels every receiver immediately.
+func receiveConcurrently(
```
sbalabanov, line 146: refactor the implementation - do not use mutex. Instead use a common shared channel to send back results. In which case no need to use a waitgroup, just expect N results in a shared channel from N goroutines, which would include cancellations. The main thread should run cancel() once it sees the first error, and still waits for other goroutines to return with context cancelled error.

sbalabanov: Addressed in ffdad06. Each worker now sends acknowledged-message events and exactly one terminal result through a common result channel. The main goroutine exclusively owns message aggregation, cancels on the expected unique-message count or the first real worker error, and continues until all N terminal worker results are received. `context.Canceled` is ignored only after successful completion or after another error triggered cancellation; otherwise cancellation is propagated as the failure. Deliveries are acknowledged before their message event is reported, so only successful acknowledgements contribute to completion and duplicate detection. The mutex, WaitGroup, and local timeouts remain removed. The focused target and race-detector run both pass.

behinddwalls (approved):

PR #438, opened 2026-07-23
test: make Docker suites hermetic

sbalabanov:
## Summary

- cross-compile container service binaries through Bazel and expose explicit Docker test-context filegroups
- resolve Compose files only from declared runfiles and stage minimal temporary Docker build contexts
- remove repository-root discovery plus `MODULE.bazel` / `go.mod` marker data dependencies
- make Docker test targets declare network access and remove pre-build Make prerequisites
- derive test image names from declared input contents to keep caching deterministic

## Sandbox evaluation

The Docker suites already run under Bazel's sandboxed strategy (`processwrapper-sandbox` locally). I did not enable Bazel's experimental hermetic Linux sandbox globally: it omits the host root and would require nonportable mounts for the Docker CLI, plugin, socket, shared libraries, and certificate/config paths. The tests now get source inputs only through Bazel runfiles while retaining the host Docker boundary they inherently require.

## Validation

- `make build`
- `make test`
- `make integration-test`
- `make e2e-test`
- `make lint`
- `make check-tidy`
- `make check-gazelle`

PR #439, opened 2026-07-24
refactor(test): make all Docker tests hermetic

sbalabanov:
## Intent

Make every Docker-based test fully hermetic: all inputs are explicit Bazel `data` dependencies, no test resolves the repository root, and the Bazel sandbox is stricter than before.

> Note: branch is `e2e-hermetic-2` because `e2e-hermetic` (#438) was taken by a concurrent attempt at the same task; this PR is an independent implementation.

## Problem

Docker suites resolved the source checkout at runtime — `testutil.FindRepoRoot` walked runfile symlinks of `MODULE.bazel`/`go.mod` (declared as `data` solely for that purpose) or shelled out to git — and handed the checkout to `docker compose` as the image build context. The images `COPY`'d pre-built `.docker-bin/*` binaries produced by `make build-*-linux` **outside** Bazel's action graph, so `bazel test` results depended on undeclared, stale-able state.

## Changes

**Hermetic inputs**
- New `go_cross_binary` targets (`gateway_linux`, `orchestrator_linux`, `runway_linux`, `stovepipe_linux`): Bazel builds the exact Linux binary that lands in each image; tests depend on them as `data`.
- `testutil.WithBuildContext` stages a minimal docker build context (binary + Dockerfile + configs, all from declared runfiles) into a test temp dir; compose's `REPO_ROOT` variable now points at that staged context. Existing compose files and `local-*` Makefile flows are unchanged.
- `testutil.Runfile` resolves compose files from runfiles; `FindRepoRoot` is deleted and every `//:MODULE.bazel` / `//:go.mod` data dep is removed.
- Image tag prefix derives from `TEST_TARGET` (stable per target) instead of the checkout path, so parallel targets no longer share tags.
- Compose V2 plugin is preferred over standalone `docker-compose` (legacy V1 lacks `up --wait`); `DOCKER_CONFIG` falls back to the invoking user's config dir so user-level CLI plugins are found despite Bazel's scratch `HOME`.

**Stricter sandbox** (beneficial now that tests are hermetic)
- `--incompatible_strict_action_env`: static action/test environment — no client env leaks into results.
- `--sandbox_default_allow_network=false`: sandboxed actions get their own network namespace by default; Docker tests opt back in via `tags = ["requires-network"]`.
- Dropped the `external` tag: hermetic test results are now cacheable — unchanged integration/e2e suites skip entirely (verified locally: `(cached) PASSED`).
- `DOCKER_HOST` / `SKIP_CLEANUP` are explicit `--test_env` pass-throughs.

**Makefile / docs**
- `integration-test*` / `e2e-test` no longer depend on `build-all-linux` (still used by `local-*` compose flows).
- Updated AGENTS.md / TESTING.md guidance.

## Test Plan

- [x] `bazel test` on all 9 Docker suites passes locally (5 MySQL-only + gateway + orchestrator + stovepipe + both e2e)
- [x] `make test` (79 unit tests) passes
- [x] `make lint`, `make check-tidy`, `make check-gazelle` pass
- [x] Test-result caching verified (second run: `(cached) PASSED`)
- [ ] CI green on all required checks


behinddwalls (approved):

sbalabanov:
I am running this PR (Claude Fable) against this: https://github.com/uber/submitqueue/pull/438 (GPT Sol).
It should not be landed quite yet :-/

PR #444, opened 2026-07-27
feat(entity): speculation path and run entities

behinddwalls:
Add the speculation domain model in submitqueue/entity/speculation.go: SpeculationPath (keyed by a content hash), DependencyBet/DependencyBetType, SpeculationPathStatus, SpeculationPathEntry, and SpeculationPathSet, plus the run vocabulary PathAction/Speculation/CandidatePath. Enums are string-valued with empty-string sentinels; the path's ID is a SHA-256 over the head and its ordered bets, computed on each call. Covered by table tests.

Also aligns the speculation RFC with the implemented design: field names match the entities (Head, Batch, string enums), the Generator opens over the batches alone and enumerates the whole path space, terminal-path suppression moves to the Allocator (the piece that reconciles candidates against the path sets by ID), and the stale depth-bound language is removed — lazy generation removed the 2^n cost the bound existed to cap.

No storage, no wiring, and no serialization helpers — queue payloads get their own contract types when a message needs one. These are the entities the speculation extension and controller build on.

## Stack
1. @ #444
1. #445
1. #446
1. #450
1. #451

### DIFF
--- a/submitqueue/entity/speculation.go
+++ b/submitqueue/entity/speculation.go
@@ -1,0 +1,196 @@
+// Copyright (c) 2025 Uber Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package entity
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+	"strings"
+)
+
+// DependencyBetType is how a path treats one of its head's dependencies.
+type DependencyBetType string
+
+const (
+	// BetUnknown is the zero-value sentinel; it is never a valid bet.
+	BetUnknown DependencyBetType = ""
+	// BetIncluded bets the dependency lands: the head is built on top of it, and
+	// the path is invalidated if the dependency ultimately fails.
+	BetIncluded DependencyBetType = "included"
+	// BetExcluded bets the dependency does not land: the head is built without it,
+	// and the path is invalidated if the dependency ultimately lands.
+	BetExcluded DependencyBetType = "excluded"
+	// BetDropped marks a dependency ignored by conflict relaxation: whether it
+	// lands or fails never affects the path — it neither gates the merge nor
+	// invalidates the path.
+	BetDropped DependencyBetType = "dropped"
+)
+
+// DependencyBet is a path's bet on one dependency of its head.
+type DependencyBet struct {
+	// Batch is the dependency batch ID this bet is about.
+	Batch string
+	// Bet is how the path treats the dependency (included, excluded, or dropped).
+	Bet DependencyBetType
+}
+
+// SpeculationPath is one guess at how a batch's dependencies resolve: a head
+// batch plus a bet on each of its dependencies. Every dependency of the head
+// appears exactly once, in queue order, so the path is self-describing — its
+// full meaning can be read without consulting any external relaxed set or
+// dependency list.
+type SpeculationPath struct {
+	// Head is the ID of the batch being built along this path.
+	Head string
+	// Bets is one bet per dependency of Head, in queue order.
+	Bets []DependencyBet
+}
+
+// ID returns the path's stable identity: a hex-encoded SHA-256 over the head
+// and its bets in order. Two paths with the same head and the same ordered
+// bets share an ID; any difference in head, dependency, or bet yields a
+// different ID. The hash is computed on every call and never cached, so a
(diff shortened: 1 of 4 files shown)


submitqueue/entity/speculation.go
```diff
@@ -0,0 +1,229 @@
...
+	Bet DependencyBetType
+}
+
+// SpeculationPath is one guess at how a batch's dependencies resolve: a head
+// batch plus a bet on each of its dependencies. Every dependency of the head
+// appears exactly once, in queue order, so the path is self-describing — its
+// full meaning can be read without consulting any external relaxed set or
+// dependency list.
+type SpeculationPath struct {
+	// Head is the batch being built along this path.
```
sbalabanov: wording should say "batch id" ? May be even field name.

submitqueue/entity/speculation.go
```diff
@@ -0,0 +1,229 @@
...
+	Head string
+	// Bets is one bet per dependency of Head, in queue order.
+	Bets []DependencyBet
+}
+
+// ID returns the path's stable identity: a hex-encoded SHA-256 over the head
+// and its bets in order. Two paths with the same head and the same ordered
+// bets share an ID; any difference in head, dependency, or bet yields a
+// different ID.
+func (p SpeculationPath) ID() string {
```
sbalabanov, line 66: may be GetID() to indicate that the operation is always computational and there is no cache?

submitqueue/entity/speculation.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+}
+
+// SpeculationPathEntry is the stored record of one chosen speculation path,
+// keyed by the hash of its content. It holds no build reference (that lives on
+// the separate execution record, keyed by (ID, Attempt)) and no score (a score
+// is meaningful only within a single speculation run).
+type SpeculationPathEntry struct {
+	// ID is the primary key: the hash of the path's content (head plus its
+	// bets). It equals Path.ID().
```
sbalabanov: is there a reason to duplicate from path.ID?

submitqueue/entity/speculation.go
```diff
@@ -0,0 +1,229 @@
...
+	// Version is the version of the object. It is used for optimistic locking.
+	// Versioning starts at 1 and is incremented for each change to the object.
+	Version int32
+	// CreatedAtMs is the creation time in Unix epoch milliseconds.
+	CreatedAtMs int64
+	// UpdatedAtMs is the last-update time in Unix epoch milliseconds.
+	UpdatedAtMs int64
+}
+
+// ToBytes serializes the SpeculationPathEntry to JSON bytes for queue message payload.
```
sbalabanov: is it ever transmitted over queue? what serialization is for? May be add it when needed?

submitqueue/entity/speculation.go
```diff
@@ -0,0 +1,229 @@
...
+}
+
+// SpeculationPathSet is one head's chosen speculation paths under a single
+// version. It holds both live paths and recently finished ones — finished
+// entries linger briefly so that a re-run cannot collide with an old build.
+// Every path in the set shares the same head and bets over the same ordered
+// dependency list.
+type SpeculationPathSet struct {
+	// BatchID is the primary key: the head batch these paths speculate on.
+	BatchID string
```
sbalabanov: to name is consistently with SpeculationPath - "Head" ? or at least the same

sbalabanov (approved):

submitqueue/entity/speculation.go
behinddwalls: Reworded — the doc now reads "Head is the ID of the batch being built along this path." Kept the field name `Head`: with the path-set rename from your other comment, the path and its set now share the same `Head` name, and the "ID of" wording plus the string type make the ID-ness explicit.

submitqueue/entity/speculation.go
behinddwalls: Kept `ID()` — Go convention avoids the `Get` prefix for accessors (Effective Go), and `GetID` would read as a plain getter rather than signal computation. Instead the doc now states it explicitly: the hash is computed on every call and never cached, so callers that compare IDs repeatedly should keep the result (the bestfirst heap already caches it per candidate for exactly this reason).

submitqueue/entity/speculation.go
behinddwalls: Intentional, and the doc now says why: the entry is a stored record and `ID` is its primary key, materialized so lookups and comparisons read it without rehashing the path. The invariant `ID == Path.ID()` is stated on the field.

submitqueue/entity/speculation.go
behinddwalls: You are right — nothing transmits these yet, they were used only by their own tests. Removed all three ToBytes/FromBytes pairs and their tests; when a queue message needs a path payload it gets its own contract type per the messagequeue-contract RFC.

submitqueue/entity/speculation.go
behinddwalls: Renamed `SpeculationPathSet.BatchID` → `Head`, matching `SpeculationPath.Head`. Also aligned the RFC's API block with the entities while in there (it still showed `HeadBatchID`/`BatchID` and int enums).

PR #445, opened 2026-07-27
feat(speculation): speculator extension contract

behinddwalls:
Add submitqueue/extension/speculation/speculator: the Speculator interface — the one speculation extension the speculate controller calls — with Config, Factory, and mocks. A Speculator decides which paths to build and which running ones to cancel within the queue's build budget; verdicts stay controller-owned, so a swapped-in implementation changes which paths run, never a batch's outcome.

Speculate's doc spells out the contract formally: what batches and pathSets contain, what the returned actions mean (empty result = nothing to change; order carries no meaning), and the error contract — plain unclassified errors for infrastructure failures only, context errors passed through, and an abandoned run retried by the next dirty signal.

## Stack
1. #444
1. @ #445
1. #446
1. #450
1. #451


submitqueue/extension/speculation/speculator/speculator.go
```diff
@@ -0,0 +1,54 @@
...
+// speculate controller calls. It can never express a verdict: whether a batch
+// merges or fails is fixed by the facts and computed by the controller, so a
+// swapped-in Speculator changes which paths run, never a batch's outcome.
+type Speculator interface {
+	// Speculate is handed the queue's in-flight batches plus any finalized
+	// batches still referenced as dependencies (each with its dependency list and
+	// state) and every path set for them — live and recently finished, so it will
+	// not re-propose a path that already passed or failed. It returns the build
+	// and cancel actions it proposes; a path it wants left as-is has no entry in
+	// the result.
```
sbalabanov: a formal definition of arguments as a list? and return values, including error contract (i.e. context errors passed out).

sbalabanov (approved):

submitqueue/extension/speculation/speculator/speculator.go
behinddwalls: Done — the doc on Speculate now lists the contract formally: what `batches` and `pathSets` contain; the return semantics (build funds a path, cancel preempts one, a kept path has no entry, empty result = nothing to change, order carries no meaning); and the error contract — errors are reserved for infrastructure failures, returned plain/unclassified per the platform/errs rules, ctx cancellation/expiry passes through so `errors.Is(err, context.Canceled)` holds, and on error the actions are nil and the run is abandoned for the next dirty signal to retry with fresh state.

PR #446, opened 2026-07-27
feat(speculation): generator contract and bestfirst impl

behinddwalls:
### Why?

The standard Speculator needs a ranked stream of candidate speculation paths to fund. Nothing enumerates or ranks the path space today.

### What?

Adds `submitqueue/extension/speculation/generator` — the Generator/Iterator contract — plus the `bestfirst` implementation and mocks.

The generator is a pure enumerator: `Generate` takes only the queue’s batches and offers every coherent path, including ones whose builds already ran. Suppressing those belongs to the Allocator, which reconciles candidates against stored path sets by ID. The snapshot is a strict contract: every batch a head’s direct dependencies reference must be present, and a snapshot with a missing dependency, empty or duplicate IDs, or a self or repeated dependency errors instead of opening a stream — any defaulting for a batch that is hard to score belongs to the scorer, not the generator.

`bestfirst` ranks each path by how likely all its assumptions are to hold and generates lazily in two levels: each head enumerates its own paths through a per-head stream — extend/swap moves over a cheapest-first flip list, worked out only when the head is first pulled — and a global heap holds each head’s current best, so pulling k paths does O(k) work however large the space behind them. Scores are summed log probabilities so wide heads cannot underflow into ties, every score is summed fresh from the head’s best score so the ordering holds in floating point as computed, and the extend/swap tree emits each path exactly once in non-increasing score order. The full algorithm, tie-breaks, derivation, and floating-point constraints live in `doc/rfc/submitqueue/speculation-generator-best-first.md`; the package README records only operational behavior.

Also clarifies `scorer.Scorer`’s contract as final success: the probability that the batch ultimately succeeds — reaches its terminal Succeeded state with its changes landed — not merely that its build passes. That is the quantity bestfirst consumes: a path’s assumptions resolve at terminal batch states, and a Merging dependency has already passed its build yet can still fail to land.

## Test Plan

✅ `./tool/bazel test //submitqueue/extension/speculation/...`
✅ `make fmt`
✅ `make lint`
✅ `make check-tidy`
✅ `make check-gazelle`
✅ `make check-mocks` — fails only on the pre-existing `storage/mock/request_batch_store_mock.go` header drift that exists on `main`; the new `generator/mock` is up to date


submitqueue/extension/speculation/generator/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+**The best path is not "bet everything lands."** Each guess takes the *likelier* outcome, so a dependency that will probably fail is excluded. Every other path flips some of those guesses, and each flip costs a known factor. One consequence worth knowing: flipping two cheap guesses can beat flipping one expensive guess, and the ordering handles that correctly.
+
+**Laziness at two levels.** Across heads, a heap holds each head's current offer — a real, already-built path, never an estimate — so the top of the heap is the best path in the queue. Within a head the same shape repeats: it keeps a small heap, hands out its best, and builds only the one or two paths that come after it. Pulling *k* candidates builds about *k* paths, whatever the size of the set behind them.
+
+`bestfirst` discards nothing: pull long enough and every path within the depth bound comes out. It does no conflict relaxation (`dropped` bets) — that is the `Speculator`'s call, not the generator's.
```
sbalabanov: why need a depth bound? Should it be simply guarded by build budget?

submitqueue/extension/speculation/generator/generator.go
```diff
@@ -0,0 +1,45 @@
...
+	// Open starts the candidate stream. The consumer pulls candidates lazily from
+	// the returned iterator.
+	Open(ctx context.Context, batches []entity.Batch, pathSets []entity.SpeculationPathSet) (PathIterator, error)
+}
+
+// PathIterator is a pull-based stream of candidate paths. The producer computes
+// only what is pulled.
+type PathIterator interface {
+	// Next yields the next candidate in the generator's order. ok is false once
+	// the coherent, depth-capped space is exhausted. Candidates never repeat and
```
sbalabanov: "depth-capped space is exhausted"
contradicts to the contract above saying that generator yields candidates however it likes

submitqueue/extension/speculation/generator/generator.go
```diff
@@ -0,0 +1,45 @@
...
+	// never contradict a resolved fact.
```
sbalabanov: probably document, may be example, what "resolved fact" is (like failed spec path)

submitqueue/extension/speculation/generator/bestfirst/bestfirst.go
```diff
@@ -0,0 +1,525 @@
...
+	"sort"
+
+	"github.com/uber/submitqueue/submitqueue/entity"
+	"github.com/uber/submitqueue/submitqueue/extension/scorer"
+	"github.com/uber/submitqueue/submitqueue/extension/speculation/generator"
+)
+
+// defaultLandingChance is what we assume for a dependency we cannot score
+// because it is not among the live batches: a coin flip.
+const defaultLandingChance = 0.5
```
sbalabanov: I'd do default to at least 95%, or what is practically observed conservative success rate is

submitqueue/extension/speculation/generator/bestfirst/bestfirst.go
```diff
@@ -0,0 +1,525 @@
...
+	batchByID := make(map[string]entity.Batch, len(batches))
+	for _, b := range batches {
+		batchByID[b.ID] = b
+	}
+	terminalByHead := terminalPathIDs(pathSets)
+
+	// chanceOf remembers each dependency's chance of landing, so a dependency
+	// that several heads wait on is only scored once.
+	chances := make(map[string]float64)
+	chanceOf := func(depID string) (float64, error) {
```
sbalabanov: might be hard to read vs standalone unexported functions

submitqueue/extension/speculation/generator/bestfirst/bestfirst.go
```diff
@@ -0,0 +1,525 @@
...
+			return c, nil
+		}
+		dep, ok := batchByID[depID]
+		if !ok {
+			// Not a live batch, so there is nothing to score. Remember the coin
+			// flip too, so we don't look it up again.
+			chances[depID] = defaultLandingChance
+			return defaultLandingChance, nil
+		}
+		c, err := g.scorer.Score(ctx, dep)
```
sbalabanov: do we rescore each batch for every speculation run? If scoring only depends on immutable inputs, it could be optimized.
At least, put some expectations on contract that it should be fast.

submitqueue/extension/speculation/generator/bestfirst/bestfirst.go
```diff
@@ -0,0 +1,525 @@
...
+// choice is one unfinished dependency that a head has to guess about.
+//
+// The likelier of the two outcomes is the "preferred" bet and the other is the
+// "alternative". Betting the alternative is always the worse guess, and penalty
+// says how much worse: multiply a path's score by it and you get the score of
+// the same path with this one guess flipped. A penalty near 1 means flipping
+// barely costs anything (the dependency was close to a coin flip); a penalty
+// near 0 means flipping it is very expensive.
+type choice struct {
+	// dep is the dependency batch being guessed about.
```
sbalabanov: batch id?

submitqueue/extension/speculation/generator/bestfirst/bestfirst.go
```diff
@@ -0,0 +1,525 @@
...
+	dep string
+	// slot is where this dependency sits in the head's bets, so a flip writes
+	// back to the right place and bets stay in queue order.
```
sbalabanov: elaborate? lost me here

submitqueue/extension/speculation/generator/bestfirst/bestfirst.go
```diff
@@ -0,0 +1,525 @@
...
+	slot int
+	// alternativeBet is the bet used when this guess is flipped.
+	alternativeBet entity.DependencyBetType
+	// preferred is the chance of the likelier outcome, so always >= 0.5.
```
sbalabanov: but less than 1? should say explicitly

submitqueue/extension/speculation/generator/bestfirst/bestfirst.go
```diff
@@ -0,0 +1,525 @@
...
+}
+
+// settledBet reports the bet a dependency's state forces, if that state is
+// final. An unfinished dependency has no forced bet and has to be guessed at.
+func settledBet(state entity.BatchState) (entity.DependencyBetType, bool) {
+	switch state {
+	case entity.BatchStateSucceeded:
+		// Already landed, so including it is a fact, not a guess.
+		return entity.BetIncluded, true
+	case entity.BatchStateFailed, entity.BatchStateCancelled:
```
sbalabanov: if it is still "cancelling", we probably should proactively exclude as well?

submitqueue/extension/speculation/generator/bestfirst/bestfirst.go
```diff
@@ -0,0 +1,525 @@
...
+	// generated counts how many paths have been built. Nothing reads it in
+	// production; it is here so tests can check we really are building lazily.
+	generated int
+}
+
+// settledBet reports the bet a dependency's state forces, if that state is
+// final. An unfinished dependency has no forced bet and has to be guessed at.
+func settledBet(state entity.BatchState) (entity.DependencyBetType, bool) {
+	switch state {
+	case entity.BatchStateSucceeded:
```
sbalabanov: if can probably be even "landed" - for bypass large diffs

submitqueue/extension/speculation/generator/bestfirst/bestfirst.go
```diff
@@ -0,0 +1,525 @@
...
+	default:
+		return entity.BetUnknown, false
+	}
+}
+
+// newHeadStream sets up one head. It sorts the head's dependencies into two
+// piles — ones that already finished, whose bet is a known fact, and unfinished
+// ones, which become choices to guess about — and then builds the head's best
+// path so the stream has something to offer. It returns ok=false when the head
+// has more unfinished dependencies than the depth bound allows.
```
sbalabanov: so it combines multiple iterators into the one final one?
isn't it simpler to build a dependency graph instead and traverse it with a single iterator?

submitqueue/extension/speculation/generator/bestfirst/bestfirst_test.go
```diff
@@ -0,0 +1,876 @@
...
+func streamFor(t *testing.T, iter generator.PathIterator, headID string) *headStream {
+	t.Helper()
+	it, ok := iter.(*iterator)
+	require.True(t, ok, "iterator is not the bestfirst merge")
+	for _, s := range it.streams {
+		if s.head == headID {
+			return s
+		}
+	}
+	t.Fatalf("no stream for head %q", headID)
```
sbalabanov: prefer "require" on nil/nonnil return value

sbalabanov (changes requested):

submitqueue/extension/speculation/generator/README.md
behinddwalls: yes.. i am gonna remove it, i thought it helped limit how deep we went in the deps but i think we should fine with bestfirst approach

submitqueue/extension/speculation/generator/generator.go
behinddwalls: yes getting rid of it

submitqueue/extension/speculation/generator/bestfirst/bestfirst.go
behinddwalls: ACK, will do it and split it for flip case

submitqueue/extension/speculation/generator/bestfirst/bestfirst.go
behinddwalls: will move it out

submitqueue/extension/speculation/generator/bestfirst/bestfirst.go
behinddwalls: Yes.. right now. we can have Scorer contract be the owner on it and will need to be fast and cache anything if needed

submitqueue/extension/speculation/generator/bestfirst/bestfirst.go
behinddwalls: Yes

submitqueue/extension/speculation/generator/bestfirst/bestfirst.go
behinddwalls: i think the risk is if they don't get cancelled and get landed..then we may optimistically assumed a future state which may not be true?

submitqueue/extension/speculation/generator/generator.go
behinddwalls: Added examples: a dependency that already landed is never bet excluded, one that already failed is never bet included, and a path whose build already finished is not offered a second time.

submitqueue/extension/speculation/generator/bestfirst/bestfirst.go
behinddwalls: Done — now 0.95.

Worth flagging what this needed first. The constant was doing double duty: the fallback for an unscoreable dependency *and* the even-odds threshold for picking the preferred bet (`if c < defaultLandingChance`). Raising it on its own would have flipped the preferred bet to *excluded* for every dependency scored below 0.95 — most of the queue — silently inverting the ranking.

Split into `evenOdds = 0.5` for the threshold and `defaultLandingChance = 0.95` for the fallback, so the two can move independently.

submitqueue/extension/speculation/generator/bestfirst/bestfirst.go
behinddwalls: Agreed — extracted into an unexported `chanceCache` type with an `of(ctx, batchID)` method.

submitqueue/extension/speculation/generator/bestfirst/bestfirst.go
behinddwalls: Yes: once per dependency per run, memoized within a run but not carried across runs.

Documented that on `scorer.Scorer.Score` — a run scores each batch at most once, results are not reused next run, and anything expensive to compute belongs behind the implementation's own cache. Note that file is pre-existing on `main` and outside this stack, so it shows up here as an unrelated path.

submitqueue/extension/speculation/generator/bestfirst/bestfirst.go
behinddwalls: Yes, it is. Renamed `dep` → `batchID`, and the doc now says it identifies the dependency batch.

submitqueue/extension/speculation/generator/bestfirst/bestfirst.go
behinddwalls: Rewritten. `slot` is the dependency's position in the head's dependency list, and so in the head's bets.

The reason it has to be recorded: the `choices` slice gets sorted by penalty, which scrambles the original order. Each choice therefore carries the index it came from, and that's how a flipped bet gets written back in queue order rather than penalty order.

submitqueue/extension/speculation/generator/bestfirst/bestfirst.go
behinddwalls: Said explicitly now: `preferred` is in [0.5, 1], `alternative` in [0, 0.5], and `penalty` in [0, 1].

submitqueue/extension/speculation/generator/bestfirst/bestfirst.go
behinddwalls: A dependency graph turns out to be the wrong shape for this. What gets enumerated isn't the dependencies themselves but the power set of *bets* over a head's unfinished dependencies — `{d0:included, d1:excluded}` and so on. Those combinations aren't nodes in the graph, so traversing it in dependency order doesn't produce them in score order.

But the instinct behind the question was right: there was no need for two iterators. Both heaps ordered by the same key, and a head's children only ever landed in its own heap, so they collapse into one heap over `(head, flipSet)` — seed it with every head's best path, then pop the best, emit it, and push that path's children within the same head.

Same output, and it deletes `headStream` as an iterator, the refill dance, and one of the two heap instantiations. It also made `Open` strictly lazier: one path built per head instead of two. Thanks for pushing on it.

submitqueue/extension/speculation/generator/bestfirst/bestfirst_test.go
behinddwalls: Done — `require.NotNil` instead of `t.Fatalf` followed by `return nil`.

submitqueue/extension/speculation/generator/bestfirst/bestfirst.go
behinddwalls: yes,.. succeeded is actually same as landed: https://github.com/uber/submitqueue/blob/main/submitqueue/entity/batch.go#L32

submitqueue/extension/speculation/generator/bestfirst/bestfirst.go
behinddwalls: Yes — once per run, never across runs. The cache lives for one `Open`; a later run scores the same batch again on purpose, since a score can move as the queue does.

The contract already carries the expectation you are asking for: `scorer.Scorer` says implementations should be cheap, that a run scores each batch at most once, and that because results are not carried across runs anything expensive belongs behind the implementation's own cache — which is the right place to exploit immutable inputs, since only the implementation knows what is safe to reuse. I have made the same point explicit on `scoreCache`, since that is where the question naturally comes up.

submitqueue/extension/speculation/generator/bestfirst/bestfirst.go
behinddwalls: Gone entirely. `pathFor` now mutates the copied best-path template in place, so an alternative only needs the index it applies at — the batch ID is already on the `PathDependency` being written to. The struct is down to `at`, `assumption`, `relativeScore`.

submitqueue/extension/speculation/generator/bestfirst/bestfirst.go
behinddwalls: Reworded and renamed. It is `at` now, documented as the index into `bestPath.Dependencies` at which this alternative swaps the assumption — and, importantly, why the index has to be carried at all: the alternatives slice is sorted by `relativeScore`, which scrambles queue order, so each one remembers where it came from.

submitqueue/extension/speculation/generator/bestfirst/bestfirst.go
```diff
@@ -0,0 +1,178 @@
...
+	// moving it would leave that side below probability 0.5 and relativeScore
+	// above 0, which breaks the best-first ordering.
+	scoreCutoff = 0.5
+
+	// defaultScore is used for a dependency that is not among the live batches
+	// and so cannot be scored. Nearly every batch a queue accepts does build
+	// successfully, so a high default is closer to the truth than an even split
+	// and keeps an unscoreable dependency from dragging down every path that
+	// assumes it succeeds.
+	defaultScore = 0.95
```
sbalabanov: this may be configurable per queue

submitqueue/extension/scorer/scorer.go
```diff
@@ -22,11 +22,17 @@ import (
...
-// Scorer computes a success probability score for a batch based on its changes.
+// Scorer computes the probability that a batch's build succeeds, based on its
+// changes.
 type Scorer interface {
-	// Score returns a probability between 0.0 and 1.0 indicating the likelihood
-	// of a successful land for the given batch. It is handed the batch identity
-	// and resolves the batch's changes itself through an injected changeset.Resolver.
+	// Score returns a probability between 0.0 and 1.0 that the given batch's
+	// build succeeds. It is handed the batch identity and resolves the batch's
+	// changes itself through an injected changeset.Resolver.
```
sbalabanov: > injected changeset.Resolver

Should not document implementation internals on what data the scorer should manipulate, as long as the interface is not expected to implement this.

submitqueue/extension/speculation/generator/bestfirst/bestfirst.go
```diff
@@ -0,0 +1,326 @@
...
+	// moving it would leave that side below 0.5 and flipCost above 0,
+	// which breaks the best-first ordering.
+	probabilityCutoff = 0.5
+
+	// defaultProbability is used for a dependency that is not among the live
+	// batches and so cannot be scored. Nearly every batch a queue accepts does
+	// build successfully, so a high default is closer to the truth than an even
+	// split and keeps an unscoreable dependency from dragging down every path
+	// that assumes it succeeds.
+	defaultProbability = 0.95
```
sbalabanov: Shouldn't a scorer return something for each dependency, period?

submitqueue/extension/speculation/generator/bestfirst/bestfirst.go
```diff
@@ -0,0 +1,326 @@
...
+)
+
+const (
+	// probabilityCutoff is where the preferred assumption switches sides: at or
+	// above it a dependency is assumed to succeed, below it assumed to fail.
+	//
+	// Not a tuning knob. The preferred side is whichever one this favors, so
+	// moving it would leave that side below 0.5 and flipCost above 0,
+	// which breaks the best-first ordering.
+	probabilityCutoff = 0.5
```
sbalabanov: Is it needed? Hard to understand. Can we just hardcode that preferred assumption is for which the probability is the highest?

submitqueue/extension/speculation/generator/bestfirst/bestfirst.go
```diff
@@ -0,0 +1,326 @@
...
+// dependency is what one dependency batch contributes to any path over it:
+// which assumption is preferred, what the preferred side scores, and what
+// deviating costs. It follows from the batch alone, so heads that share a
+// dependency share this, and its logarithms are taken once per queue rather
+// than once per head that depends on it.
+type dependency struct {
+	// unresolved reports whether the outcome is still open. A resolved
+	// dependency forces its assumption and offers no flip, so the two
+	// unpreferred fields below mean nothing for it.
+	unresolved bool
```
sbalabanov: prefer "resolved" with default to false?

submitqueue/extension/speculation/generator/bestfirst/bestfirst.go
```diff
@@ -0,0 +1,326 @@
...
+}
+
+// known returns what has already been worked out for the given dependency, and
+// whether anything has. The zero value is not a valid answer, so a caller that
+// might be looking one up for the first time must check ok.
+//
+// Deliberately just the map read, so it is cheap enough for the compiler to
+// inline it: this runs once per dependency of every head, which on a deep queue
+// is millions of times. Pairing it with compute at the call site rather than
+// wrapping the two in one method is what keeps it inlinable.
```
sbalabanov: remove anything about inlining, this is not important enough to bloat the doc

submitqueue/extension/speculation/generator/bestfirst/bestfirst.go
```diff
@@ -0,0 +1,326 @@
...
+		// path's probability by 1, so a resolved dependency adds nothing to any
+		// head's total. unresolved stays false, so it offers no flip and
+		// drops out of the search — the path takes the fact and moves on.
+		d := dependency{preferred: assumption}
+		c.cached[batchID] = d
+		return d, nil
+	}
+
+	// Not a live batch, so there is nothing to score. Its outcome is still
+	// open, though, so it keeps both assumptions and just sits at the default.
```
sbalabanov: scorer may rely on info collected before batch enters speculation ("live"), like number of changed targets. I think we need to require scorer to always give a score, if it needs a default, keep it with scorer implementation.

submitqueue/extension/speculation/generator/bestfirst/bestfirst.go
```diff
@@ -0,0 +1,326 @@
...
+	// unresolved reports whether the outcome is still open. A resolved
+	// dependency forces its assumption and offers no flip, so the two
+	// unpreferred fields below mean nothing for it.
+	unresolved bool
+	// preferred is the assumption a path takes unless it deviates: for a
+	// resolved dependency the one its state forces, otherwise the more probable
+	// of the two.
+	preferred entity.DependencyAssumption
+	// unpreferred is the assumption a path takes by deviating.
+	unpreferred entity.DependencyAssumption
```
sbalabanov: why do we need to keep `unpreferred` if it is always an inverse of `preferred` ?

submitqueue/extension/speculation/generator/bestfirst/head.go
```diff
@@ -0,0 +1,237 @@
...
+// paths. It roots a tree of them but holds no heap of its own — every head's
+// candidates compete in the iterator's single heap.
+//
+// A head starts thin: an ID, its dependency list, and one score. Its
+// flips are worked out the first time it is expanded, so a head nobody
+// pulls from costs no more than that.
+type head struct {
+	// id is the head batch's ID, and the Head of every path it yields.
+	id string
+	// dependencyIDs are the head batch's dependencies in queue order. The slice
```
sbalabanov: specify, all or only direct ones

submitqueue/extension/speculation/generator/bestfirst/head.go
```diff
@@ -0,0 +1,237 @@
...
+	flipsPrepared bool
+}
+
+// newHead totals a head's best score. That is all Open does per head: one
+// pass over its dependencies, adding what each contributes.
+//
+// A cancelled or expired ctx aborts with its error. The check is per dependency
+// rather than per head because Open walks every dependency of every head, which
+// on a deep queue is millions of them.
+func newHead(ctx context.Context, batch entity.Batch, cache *dependencyCache) (*head, error) {
```
sbalabanov: why return a pointer and not an value?

submitqueue/extension/speculation/generator/bestfirst/head.go
```diff
@@ -0,0 +1,237 @@
...
+		if err := ctx.Err(); err != nil {
+			return nil, err
+		}
+		d, ok := cache.known(dep)
+		if !ok {
+			var err error
+			if d, err = cache.compute(ctx, dep); err != nil {
+				return nil, err
+			}
+		}
```
sbalabanov: is there any case when `known` and `compute` are not always called together? Should they just be a single function, like getOrCompute() ?

sbalabanov (approved):
accepting to unblock, please address comments

submitqueue/extension/speculation/generator/bestfirst/bestfirst.go
behinddwalls: moved it the code

submitqueue/extension/speculation/generator/bestfirst/bestfirst.go
behinddwalls: removed

submitqueue/extension/speculation/generator/bestfirst/bestfirst.go
behinddwalls: yes, i am going add that separately

submitqueue/extension/speculation/generator/bestfirst/head.go
behinddwalls: i think we can just do value

submitqueue/extension/speculation/generator/bestfirst/head.go
behinddwalls: yes merged

PR #450, opened 2026-07-28
feat(speculation): allocator contract and sticky impl

behinddwalls:
### Why?

Ranked candidates need a budget policy: something must decide which paths get the queue's limited concurrent-build slots, without double-funding paths that are already running or finished.

### What?

Adds `submitqueue/extension/speculation/allocator` — the budget-spending contract the standard Speculator hands its candidate stream to — plus the `sticky` implementation and mocks.

The Allocator owns terminal-path suppression: it reconciles the stream against the stored path sets by path ID, so an in-flight candidate keeps the slot it already holds and a terminal one is skipped. Budget accounting reads path status and nothing else — pending, building, and cancelling paths hold a slot until their builds actually stop — and the unit is deliberately coarse: one slot per concurrent build.

`sticky` fills only free slots and never preempts a running build; its policy counterpart is a preempting allocator. Output is all-or-nothing (a partial list would fund an arbitrary prefix of the ranking), and a stored path with no status fails fast rather than guessing whether it holds a slot.

## Test Plan
✅ `bazel test //submitqueue/extension/speculation/allocator/...` — terminal candidates skipped per status, in-flight candidates kept, budget exhaustion, context cancellation, corrupt-status failure.

✅ `make fmt`, `make gazelle`, `make mocks`


submitqueue/extension/speculation/allocator/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+`Allocate` pulls candidates from the iterator in best-first order and matches them against the queue's current path sets by path ID, so a pending or building path stays funded rather than starting a new attempt. Pending, building, and cancelling paths all charge the budget (a cancelling build holds its slot until it reaches a terminal state); terminal paths charge nothing. It returns the build and cancel actions that spend the budget.
+
+Because cancellation is best-effort, an `Allocator` should not spend capacity it merely expects a cancel to release — a build cancelled to make room keeps charging the budget until its cancel reaches a terminal state, so the queue converges over successive runs rather than oversubscribing the hard cap on concurrent builds in a single pass.
+
+## `sticky`
```
sbalabanov: move that into sticky/README.md instead

submitqueue/extension/speculation/allocator/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+The `sticky` `Allocator` fills only free budget slots and leaves every in-flight build running. Against the budget it counts the paths that still hold a build slot — `pending`, `building`, and `cancelling` (a cancel is a request, and the build keeps its slot until it actually stops). It then proposes a build for each new candidate, in order, until the budget fills; it never proposes a cancel.
```
sbalabanov: add that 'sticky' operates the "builds" as a build unit budget (so, it is quite coarse).
More interesting implementations may operate with more granular units, understanding the size of the build needed to be done (i.e. based on number of targets, historical data etc).

submitqueue/extension/speculation/allocator/sticky/sticky.go
```diff
@@ -0,0 +1,108 @@
...
+
+// alloc is a sticky allocator.Allocator. budget is the queue's cap on concurrent
+// builds; pending, building, and cancelling paths all charge against it.
+type alloc struct {
+	logger *zap.SugaredLogger
+	budget int
+}
+
+// New returns a sticky allocator.Allocator with the given concurrent-build
+// budget.
```
sbalabanov: explain what unit of measure for the budget is

submitqueue/extension/speculation/allocator/sticky/sticky.go
```diff
@@ -0,0 +1,108 @@
...
+	// reason applies.
+	funded := make(map[string]bool)
+	for _, set := range pathSets {
+		for _, entry := range set.Paths {
+			if entry.Status == entity.SpeculationPathStatusUnknown {
+				// A stored path should always carry a real status, so this is a
+				// data defect. Skip it rather than counting it: a status that
+				// never turns terminal would pin a build slot forever. That leaves
+				// the path free to be proposed again, which is self-correcting,
+				// but the malformed entry is worth knowing about.
```
sbalabanov: SQ default on data inconsistencies is to fail fast and hard

sbalabanov (dismissed):
critical comment about failing fast

submitqueue/extension/speculation/allocator/README.md
behinddwalls: Moved. `sticky/README.md` now carries everything specific to sticky — what it counts against the budget, the unit, the fail-fast on a statusless path, and the never-preempt trade-off. `allocator/README.md` keeps only the contract every `Allocator` owes.

submitqueue/extension/speculation/allocator/README.md
behinddwalls: Added, in both places. `sticky/README.md`: the budget is measured in builds, one slot per path no matter what that build does, which is deliberately coarse — a twelve-hour build and a two-minute build cost the same slot. `allocator/README.md` says the unit is the implementation's choice and names the more granular option: an `Allocator` that understands build size (target count, historical cost) could weight paths instead and pack the budget more tightly.

submitqueue/extension/speculation/allocator/sticky/sticky.go
behinddwalls: Done, on the field itself: budget is the cap on concurrently held build slots, measured in builds — one slot per path whose attempt is pending, building, or cancelling, whatever that build's size. `New`'s doc says the same.

One thing worth calling out while here: the accounting reads path status and nothing else. No batch state enters it, because the budget tracks builds occupying CI, and a cancelling build occupies CI until it actually stops.

submitqueue/extension/speculation/allocator/sticky/sticky.go
behinddwalls: Changed to fail fast. A stored path with no status now makes `Allocate` return an error naming the path and its head, instead of warning and skipping.

That was the right call for a second reason: the old behaviour had to *choose* how to read a corrupt record, and either reading is wrong — count it and a status that never turns terminal pins a slot forever; skip it and the path gets double-built. Failing removes the guess. It also removed sticky's only logging dependency, so `New` now takes just the budget.

behinddwalls:
> critical comment about failing fast

i think we need log but not fail as there is no way to get rid of it afterwords; i will take care of it separately

mnoah1 (approved):

PR #451, opened 2026-07-28
feat(speculation): standard composed speculator

behinddwalls:
### Why?

The Generator ranks and the Allocator spends; something must compose them into the Speculator extension the orchestrator calls.

### What?

Adds `speculator/standard`. It funds the queue's most promising paths first until the build budget is spent: candidates considered in descending likelihood, in-flight paths kept funded rather than restarted, finished paths never re-proposed. Pairing bestfirst with sticky yields the default speculation policy; changing either behavior means swapping a part, not writing a new Speculator.

standard adds no cancellation handling of its own and inherits it from the two parts; a test pins that a cancelled run yields the context error and no actions. Behavior is documented in `standard/README.md`; the speculator README keeps only the extension contract.

## Test Plan
✅ `bazel test //submitqueue/extension/speculation/...`

✅ `make fmt`, `make gazelle`, `make mocks`

### DIFF
--- a/submitqueue/extension/speculation/speculator/README.md
+++ b/submitqueue/extension/speculation/speculator/README.md
@@ -8,6 +8,8 @@ The `speculator` package defines the one speculation extension the speculate con
 
 Like the other extensions, a `Speculator` is selected **per queue** by the wiring layer through the `Config` (queue name) and `Factory` interface. Budget, clock, and any extra data are injected at construction by the integrator, not carried on the contract.
 
+[`standard`](standard/README.md) is the default implementation: it funds the most promising paths first until the build budget is spent.
+
 ## Adding a backend
 
 Create a package under `speculator/<backend>/` whose `New(...)` returns a `speculator.Speculator`, injecting whatever it needs at construction. Resolve any content it requires internally; do not add a `Config` or `Factory` implementation here — per-queue routing and the factory adapter live in the wiring layer.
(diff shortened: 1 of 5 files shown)


submitqueue/extension/speculation/speculator/README.md
```diff
@@ -8,6 +8,12 @@ The `speculator` package defines the one speculation extension the speculate con
 
 Like the other extensions, a `Speculator` is selected **per queue** by the wiring layer through the `Config` (queue name) and `Factory` interface. Budget, depth bound, clock, and any extra data are injected at construction by the integrator, not carried on the contract.
 
+## The `standard` Speculator
```
sbalabanov: move to "standard"'s README
The description should say not how it is composed, but what it does (what behavior it implements wrt prioritizing paths and working with build budget).

sbalabanov (approved):

submitqueue/extension/speculation/speculator/README.md
behinddwalls: Moved and rewritten to lead with behaviour. `standard/README.md` now opens with what it does — funds the queue's most promising paths first until the build budget is spent; paths already pending or building keep the slot they hold rather than restarting; paths whose builds finished are skipped while their records remain in the path sets; everything below the cut waits for a later run, which is safe because a batch's verdict never depends on what was funded.

How it is composed comes after that, and only to say the two halves are swappable. `speculator/README.md` keeps just the extension contract.

Issue #354, opened 2026-07-14
batch: create the batch before updating the dependents reverse index


behinddwalls:
## Problem

Batch creation currently writes in this order: append the new batch's ID to each dependency's `BatchDependent.Dependents` reverse index → create the batch's own (empty) `BatchDependent` row → CAS the request to `Batched` (serialization point vs cancel) → `BatchStore.Create` last.

Because the reverse-index appends happen before the batch exists, two paths leave **dangling IDs** in dependencies' `Dependents` lists — IDs of batches that will never exist (batch IDs are never reused; a retry mints a fresh ID):

- the cancel-race abandon: the request-claim CAS loses to the cancel controller and batch creation is abandoned after the index appends (`batch.go` documents the entry "becomes a dangling BatchDependent — tolerated per the 'downstream should handle stale entries' contract");
- a crash between the CAS and `BatchStore.Create`.

Every reader of `BatchDependent` must therefore implement tolerance. In practice the speculate dependent fan-out publishes wakes for dangling IDs; those messages fail the batch load, reject to the DLQ, and are skipped there as not-found — noise and retry churn on the mainline path.

## Proposal

Reorder batch creation: **create the batch first, then update the dependents' reverse index**, and only then publish the batch into the pipeline.

Rationale: the failure residue flips from "dangling ID that every reader must tolerate" to "created batch whose reverse-index update failed" — and that batch is fine to lose. It has no meaning to the rest of the system: it is never published for speculation and never satisfies any speculation state, so it never advances and nothing downstream acts on it. Dangling IDs, by contrast, leak tolerance requirements into every consumer of the reverse index.

An additional win: with the request-claim CAS moved ahead of all writes (CAS → Create → own BatchDependent row → reverse-index appends → publish), the common failure source — losing the CAS race to cancel — abandons with **zero residue**, instead of leaving danglers behind.

## Caveats to address as part of this change

[...]

PR #452, opened 2026-07-29
feat(storage): add request batch assignments

ubettigole:
## Summary
Batch creation does not currently persist which batch owns a request. If processing fails after claiming the request or partially persisting a batch, redelivery can create a new batch ID, leaving duplicate or orphaned batches and stale dependency-graph entries.

This foundational PR adds an immutable request-to-batch assignment store and its MySQL implementation. Later PRs use this mapping to resume the original batch and allow cancellation to resolve durable batch ownership. This PR does not change controller behavior itself.

## Test Plan
Unit tests

## Issues

## Stack
1. @ #452
1. #453
1. #454
1. #455


sbalabanov (changes requested):
This is by design. When batching failed, a batch with new id is created and pipeline restarted. Batch dependent store should be tolerant to races (non-existing batch ids). 
submitqueue/orchestrator/controller/batch/batch.go has an explanation starting line 200.
Could use a better model to validate if these assumptions still hold true.

ubettigole:
This immutable request-to-batch assignment approach is no longer part of the revised stack. The replacement stack preserves fresh batch IDs on retry and starts at #453.

PR #454, opened 2026-07-29
fix(cancel): cancel all applicable batch attempts

ubettigole:
## Summary
Batch retries mint new batch IDs, so a request may be associated with multiple batch attempts. Previously, cancellation stopped after finding one matching active batch, which could leave other attempts running.
Update cancellation to find every eligible batch containing the request. Created, Speculating, and already-Cancelling attempts are handed to speculate for cancellation, while Merging and terminal attempts retain their existing outcomes for conclude to reconcile. Creating attempts remain ignored because they have not been published and may not be fully initialized.

Matching batches are processed deterministically, and every cancellation failure is logged and counted while preserving the first error for classification.

## Test Plan
Unit tests

## Issues

### DIFF
--- a/submitqueue/entity/batch.go
+++ b/submitqueue/entity/batch.go
@@ -62,6 +62,21 @@ func (s BatchState) IsTerminal() bool {
 	}
 }
 
+var nonCancellableBatchStates = map[BatchState]bool{
+	BatchStateUnknown:   true,
+	BatchStateCreating:  true,
+	BatchStateMerging:   true,
+	BatchStateSucceeded: true,
+	BatchStateFailed:    true,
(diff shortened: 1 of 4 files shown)


submitqueue/entity/batch.go
```diff
@@ -62,6 +62,17 @@ func (s BatchState) IsTerminal() bool {
 	}
 }
 
+var cancellableBatchStates = map[BatchState]bool{
```
behinddwalls: should we do the other way? ideally all state are except Merging and Merged? everything else can be handeled?

behinddwalls (approved):

PR #455, opened 2026-07-29
fix(batch): publish only fully initialized batches

ubettigole:
## Summary
Prevent partially initialized batches from becoming visible to downstream processing or dependency analysis.
New batches are first persisted in the Creating state. The batch controller then creates the batch’s reverse-index row, registers it with each dependency, and transitions it to Created only after structural initialization succeeds. Request-log and speculate publication happen only after that transition.

Retries continue to mint fresh batch IDs. Failed attempts may leave unpublished Creating or Created batches behind, but these attempts remain inert; metrics track their creation so a cleanup job can be added later if necessary.

## Test Plan
Unit tests

## Issues

### DIFF
--- a/submitqueue/orchestrator/controller/batch/batch.go
+++ b/submitqueue/orchestrator/controller/batch/batch.go
@@ -106,6 +106,7 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) er
 	// Short-circuit if the request has been halted — either it already reached a
 	// terminal state, or the cancel controller has recorded a cancellation intent
 	// (RequestStateCancelling). A halted request must never spawn a new batch.
+	// If cancellation races with an attempt already initializing below, speculate re-checks the contained request state before starting work.
 	if entity.IsRequestStateHalted(request.State) {
 		metrics.NamedCounter(c.metricsScope, opName, "skipped_halted", 1)
 		c.logger.Infow("skipping batch for halted request",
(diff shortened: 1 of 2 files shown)


submitqueue/orchestrator/controller/batch/batch.go
```diff
@@ -315,6 +293,45 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) er
 	return nil // Success - message will be acked
 }
 
+// initializeBatch creates the reverse-index structure and marks a Creating batch ready for publication.
+func (c *Controller) initializeBatch(ctx context.Context, batch entity.Batch) (entity.Batch, error) {
```
behinddwalls: maybe we rename to populateBatch as we already created i guess?

submitqueue/orchestrator/controller/batch/batch.go
```diff
@@ -315,6 +293,45 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) er
...
+	batchDependent := entity.BatchDependent{
+		BatchID:    batch.ID,
+		Dependents: []string{},
```
behinddwalls, line 313: can we use the conflict.Analyzer.all for now here to populate?

behinddwalls (approved):

submitqueue/orchestrator/controller/batch/batch.go
ubettigole: Talked offline - no action needed here.

PR #460, opened 2026-07-29
feat(runway): git-backed merger with REBASE

behinddwalls:
### Why?

Runway's `merger` extension has had exactly one implementation — `noop`, which always succeeds. Nothing actually merges anything. This lands the first real backend: a `Merger` driven by the `git` CLI against a local checkout.

It ships REBASE only. The strategy-specific apply paths are small and independent, but the machinery underneath them — the pinned git runtime, object resolution, the reset/apply/push cycle, contention retry, dry-run discard, conflict classification — is shared and is the bulk of what needs review. Landing it with one strategy keeps that review separable from the per-strategy mechanics that follow in this stack.

### What?

Adds `runway/extension/merger/git`.

**A change is a range of commits, not a commit.** A change URI pins a pull request to a single head SHA, but a pull request is routinely several commits. Applying the head alone applies only that commit's diff against its own parent: it conflicts against context its predecessors would have established, or — when the commits touch different files — succeeds while silently dropping everything before the head. So the unit replayed is the range from the change's merge base with the target up to its head. A change already contained in the target has an empty range and is a no-op, which is what keeps redelivery idempotent.

A range runs through git's sequencer, which changes the control flow versus a single pick: the operation stops on a commit it declines and `--skip` advances to the next rather than ending. Two kinds of commit stop it harmlessly — one already present on the target, and one that was empty to begin with — and both are skipped. Anything else is a real conflict, and the in-progress pick is aborted so the checkout stays usable. The commits an apply produced are read back off the checkout rather than tracked per-invocation, which stays correct when the sequencer drops some.

**Referenced commits are guaranteed present before anything is applied.** The default fetch refspec is `+refs/heads/*`, which does not cover a provider's change refs: a pull request head never also pushed as a branch — the normal case for a fork — is simply absent, and the apply then fails with git's "bad object", indistinguishable from a conflict. Every commit the request names is now fetched and verified up front, for all steps, so an unusable request fails without having mutated the checkout. Commits are requested by SHA (relying on the server serving a want for a reachable-but-unadvertised object, which github.com allows), falling back to the provider's canonical ref, with deployment-supplied refspecs as a last resort. Neither fetch is shallow — the range needs ancestry.

A commit a *reachable* remote cannot supply is terminal; a remote that will not answer stays retryable. The first is a property of the request, the second a property of the moment.

**One seam for change providers.** Every URI is reduced to the only three things the merger needs — the commit to apply, the ref the provider publishes it under, and a label for synthesized messages. `github://` and `git://` are supported; an unrecognized scheme is terminal. Adding a provider is one case in that mapping rather than a change to any apply path.

**Optional staleness check.** Fetching by SHA guarantees the merger applies exactly the commit the URI names, not that it is still the change's head — a force-push leaves the superseded commit fetchable on most hosts. When enabled, each change's canonical ref is read (one ref advertisement, no object transfer) and a mismatch is terminal.

**Atomicity, contention, dry run.** Nothing reaches the remote until the final push. If the push is rejected because the remote tip moved, the whole reset/apply/push cycle retries up to a bounded number of attempts. `CheckMergeability` runs the identical apply path but never pushes, then resets and reports empty outputs, committing intermediate steps locally so a multi-step check sees the same conflict surface a real merge would.

**Runtime hygiene.** Every invocation uses an explicitly pinned git runtime and a scrubbed environment — no system or global config, no interactive prompts. That leaves no ambient identity, so the committer is injected per-invocation.

`isConcreteStrategy` currently admits only REBASE; SQUASH_REBASE, MERGE and PROMOTE are rejected as invalid requests until their apply paths land later in this stack. The merger is not wired into the server yet, so this adds no production behavior.

## Test Plan

✅ `bazel test //runway/...` — 6/6 targets pass, including the new `//runway/extension/merger/git` suite (40s)

The suite drives a real git binary against throwaway repositories. Beyond the single-commit cases (stacked URIs, already-landed changes, multi-step requests, conflicts, checkout recovery, contention retry, give-up after max attempts, DEFAULT resolution, dry runs), it covers what this change is actually about: a multi-commit change expressed as **one** URI for both disjoint and overlapping edits, a partially-landed change, an empty commit inside a range, a conflict inside a range leaving no sequencer state behind, stacked multi-commit changes not duplicating each other's commits, a head reachable only via its pull-request ref, an unavailable commit classified as invalid rather than conflicting, the staleness check on and off, and provider resolution across schemes.

The regression tests were verified to fail against head-only picking: reverting just that line fails 5 of them, and they pass with the range applied.

## Stack
1. #459
1. @ #460
1. #461
1. #462
1. #463
1. #476
1. #477

### DIFF
--- a/runway/extension/merger/git/git_merger.go
+++ b/runway/extension/merger/git/git_merger.go
@@ -1,0 +1,800 @@
+// Copyright (c) 2025 Uber Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package git is a Merger implementation backed by a local git checkout. It
+// applies a MergeRequest's ordered steps onto a target branch, honoring each
+// step's merge strategy, and (for a committing Merge) pushes the result.
+//
+// Strategy → git operation:
+//
+//   - REBASE:        cherry-pick each URI's head commit onto the target tip.
+//   - DEFAULT:       resolved to the instance's configured DefaultStrategy
+//     before any step runs.
+//
+// REBASE is the only strategy implemented so far. A step naming any other
+// strategy is rejected as merger.ErrInvalidRequest until its apply path lands.
+//
+// Atomicity: for a committing merge nothing reaches the remote until the final
+// push. A step that fails to apply aborts the in-progress git operation and
+// returns without pushing.
+//
+// Contention: if the push fails because the remote tip moved between reset and
+// push, the whole reset/apply/push cycle is retried up to Params.MaxPushAttempts
+// (default 10). Detection re-fetches the remote tip after a push failure and
+// compares it to the SHA reset to at the start of the cycle.
+//
+// Dry-run (CheckMergeability) applies the exact same steps but never pushes;
+// intermediate steps are committed locally so a cumulative multi-step check
+// sees the same conflict surface, then the checkout is reset to discard them.
+// Reported StepResults carry no Outputs.
+//
+// Change URIs are resolved through resolveChange (see changeref.go), which
+// reduces any supported provider URI to the commit to apply, the ref that
+// commit canonically lives under, and a label. github:// and git:// are
+// supported; an unrecognized scheme is a terminal invalid request.
+//
+// Object availability: every referenced commit is fetched and verified before
+// any step is applied. A change is replayed as the full range from its merge
+// base with the target up to its head, because a URI pins a pull request to a
+// single head SHA while the pull request itself is routinely several commits.
+package git
+
+import (
+	"bytes"
+	"context"
+	"errors"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"sync"
+
+	"github.com/uber-go/tally"
+	"go.uber.org/zap"
+
+	mergestrategypb "github.com/uber/submitqueue/api/base/mergestrategy/protopb"
+	runwaymq "github.com/uber/submitqueue/api/runway/messagequeue"
+	runwaypb "github.com/uber/submitqueue/api/runway/messagequeue/protopb"
+	coremetrics "github.com/uber/submitqueue/platform/metrics"
+	"github.com/uber/submitqueue/runway/extension/merger"
+)
+
+// defaultMaxPushAttempts caps the retry loop when the remote tip keeps moving
+// under us. Bounded to prevent an infinite loop against a busy remote.
+const defaultMaxPushAttempts = 10
+
+// Default committer identity used when Params leaves it unset. The scrubbed
+// environment (GIT_CONFIG_NOSYSTEM, GIT_CONFIG_GLOBAL=/dev/null) leaves no
+// ambient identity, so the commit-creating REBASE strategy needs one supplied
+// explicitly.
+const (
+	defaultCommitterName  = "SubmitQueue Runway"
+	defaultCommitterEmail = "runway@submitqueue.invalid"
+)
+
+// GitRuntime identifies the explicitly provided Git runtime used by the Merger.
+type GitRuntime struct {
+	// Executable is the absolute path to the Git executable.
+	Executable string
+	// ExecPath is the absolute directory containing Git's helper executables.
+	ExecPath string
+	// TemplateDir is the absolute directory containing Git's repository
+	// templates.
+	TemplateDir string
+	// PassthroughEnv names additional environment variables to inherit from
+	// the parent process, on top of the auth and transport ones always passed
+	// through. For a deployment whose remote needs something unusual; leave
+	// empty otherwise. Names that could alter merge semantics do not belong
+	// here — the scrubbed environment is what keeps a merge reproducible.
+	PassthroughEnv []string
+}
+
+// Params holds the dependencies for the git Merger.
+type Params struct {
+	// CheckoutPath is the absolute path to an existing git checkout that the
+	// Merger will operate against. The Merger owns this working tree.
+	CheckoutPath string
+	// Remote is the name of the git remote to fetch from and push to
+	// (e.g. "origin").
+	Remote string
+	// Target is the destination branch ref on the remote (e.g. "main").
+	Target string
+	// DefaultStrategy resolves a step whose strategy is DEFAULT. Must be a
+	// concrete strategy (currently only REBASE).
+	DefaultStrategy mergestrategypb.Strategy
+	// Runtime is the pinned Git runtime used for every invocation.
+	Runtime GitRuntime
+	// MaxPushAttempts caps how many times a committing merge retries the full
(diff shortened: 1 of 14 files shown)


runway/extension/merger/git/git_merger.go
```diff
@@ -0,0 +1,651 @@
...
+// step, in order, returning the new commit SHAs (empty for picks that were
+// no-ops because the content is already on the target).
+func (m *gitMerger) pickStepChanges(ctx context.Context, step *runwaymq.MergeStep) ([]string, error) {
+	var picked []string
+	for _, uri := range step.GetChange().GetUris() {
+		cid, err := entitygithub.ParseChangeID(uri)
+		if err != nil {
+			return nil, fmt.Errorf("%w: invalid change URI %q: %v", merger.ErrInvalidRequest, uri, err)
+		}
+		sha, ok, err := m.cherryPickSHA(ctx, cid.HeadCommitSHA)
```
behinddwalls: say a PR has mutliple commits when we cherry-pick just the head? does it have all the changes or do we need to go down the history to cherry pick all commits and then put on the head/main?

runway/extension/merger/git/git_merger.go
```diff
@@ -0,0 +1,651 @@
...
+	}
+	return toOutputs(picked), nil
+}
+
+// pickStepChanges cherry-picks the head SHA of every URI of every change in the
+// step, in order, returning the new commit SHAs (empty for picks that were
+// no-ops because the content is already on the target).
+func (m *gitMerger) pickStepChanges(ctx context.Context, step *runwaymq.MergeStep) ([]string, error) {
+	var picked []string
+	for _, uri := range step.GetChange().GetUris() {
```
behinddwalls: have we fetched these commits? before cherry-picking? how do we know if we have them in local

runway/extension/merger/git/git_merger.go
behinddwalls: Good catch — you're right, and it was a real bug. Fixed in this PR.

A URI identifies a **pull request**, not a commit: `github://{host}/{owner}/{repo}/pull/{n}/{head_sha}`. So a 3-commit PR arrived as one URI carrying only commit 3's SHA, and `git cherry-pick <head>` applies only that commit's diff against its own parent. Two failure modes, both reproduced:

- **Spurious conflict** when the commits touch the same region — the head's patch has no context to apply against, so we reported `ErrConflict` for a PR that doesn't actually conflict.
- **Silent partial landing** when they touch different files. A PR adding `a.go` then `b.go` landed `b.go` only, exited clean, and published `SUCCEEDED`.

PR #141 in this repo is a live example: 4 commits, of which the old code would have landed only the last.

The fix replays the **range** rather than the commit: `merge-base(HEAD, head_sha)..head_sha`. The base is computed locally, so nothing needs to change on the wire. A few things fell out of it:

- A range runs through git's sequencer, so `--skip` advances to the next commit instead of ending the operation. Commits already on the target and commits that were empty to begin with are skipped; anything else aborts as a real conflict.
- An already-landed change has an empty range (its merge base is its own head) and is a no-op, which is what keeps redelivery idempotent.
- Across unrelated histories `merge-base` is empty, so the picking strategies now reject that explicitly and point at MERGE instead of building a malformed range.

`MERGE` and `PROMOTE` were never affected — both carry the change's full ancestry by construction — but they now have regression tests too.

Regression coverage is in `git_merger_test.go`: a multi-commit change expressed as **one** URI for both disjoint and overlapping edits, a partially-landed change, an empty commit inside a range, a conflict inside a range, and stacked multi-commit changes. I verified they catch the original bug — reverting just the range line fails 5 of them.

runway/extension/merger/git/git_merger.go
behinddwalls: We weren't, and it was worse than just missing a fetch. Fixed in this PR.

`resetToRemote` ran a bare `git fetch <remote>`, which uses the default refspec `+refs/heads/*:refs/remotes/origin/*`. That does **not** cover `refs/pull/N/head`, so any PR head not also pushed as a branch — the normal case for a fork — was simply absent locally. I confirmed this against a real remote.

The consequence was the bad part. `git cherry-pick <absent-sha>` prints `fatal: bad object`, which failed the `isRedundantCherryPick` check and fell through to the abort branch, so a **fetch problem was reported to SubmitQueue as `ErrConflict`** — telling the client its change conflicts when nothing was ever applied. `applyMerge` had the same shape; `promote` differed but was also wrong, returning a plain error that the consumer would retry forever.

The fix, per your other suggestion: fetch by SHA. Every referenced commit is now fetched and verified with `git cat-file -e` **before any step is applied**, for all steps, so an unusable request fails without having touched the checkout.

I verified `git fetch <remote> <sha>` works against real GitHub for an unadvertised PR head — `008f7d25`, reachable only from `refs/pull/141/head` — so `uploadpack.allowReachableSHA1InWant` is enabled there. It also gives the ancestry the range fix needs, provided the fetch isn't shallow (`--depth=1` returns the commit with no history and breaks `merge-base`). The provider's canonical ref is the fallback, and `MERGE_FETCH_REFSPECS` is there for a server that allows neither.

Classification is now split: a commit that a **reachable** remote cannot supply is a terminal `ErrInvalidRequest` (verified with `ls-remote` before deciding), while an unreachable remote stays a plain retryable error. The first is a property of the request, the second of the moment.

One thing fetching by SHA does *not* give us, which is worth being explicit about: it guarantees we apply exactly the commit the URI names, but not that the commit is still the PR's head — after a force-push the old SHA usually stays fetchable. That needs a separate check, so `MERGE_CHECK_STALENESS` (on by default) reads the PR ref via `ls-remote` and rejects a superseded URI. One ref advertisement, no object transfer.

kevinlnew (commented):
test

kevinlnew (commented):
test review comment

runway/extension/merger/git/git_merger.go
```diff
@@ -0,0 +1,796 @@
...
+		}
+		ch := step.GetChange()
+		if ch == nil || len(ch.GetUris()) == 0 {
+			return nil, fmt.Errorf("%w: step %q has no change URIs", merger.ErrInvalidRequest, step.GetStepId())
+		}
+		for _, uri := range ch.GetUris() {
+			if _, err := resolveChange(uri); err != nil {
+				return nil, err
+			}
+		}
```
kevinlnew, line 302: AI-generated nit: Every URI is resolved here for validation, then resolved again at line 439 inside pickStepChanges. Could thread the resolved []changeRef through to avoid the double parse.

kevinlnew (approved):
Nice, left a few comments

runway/extension/merger/git/git_merger.go
behinddwalls: Good call — done, and there was a third parse site once I looked.

`resolvedStep` now carries the resolved changes (`refs []changeRef`), populated in `resolveAndValidate` as it walks the URIs it was already walking. Every apply path reads `rs.refs`, so `resolveChange` is now called from exactly one place in the package.

The double parse was in fact worse than two: `resolveStepChanges` in `tryApply` was a *third* site, and both it and the apply-path parse sat inside the push-contention retry loop, so the count was roughly `1 + attempts × URIs × 2` rather than the `1 × URIs` it is now. `resolveStepChanges` no longer resolves anything and is now `stepChangeRefs`, a flatten of the already-parsed refs.

Once nothing in the loop needed to re-resolve, `ensureObjects` + `checkStale` had no reason to be in there either — the refs are fixed for the whole request, and re-running them per attempt just re-asks the remote what it already answered. They moved to the top of `applyTransforming`, before the first attempt. (`promote` had the same shape; @kevinlnew flagged it separately on #462 and it's fixed there too.)

No behavior change intended: parse errors are still terminal `ErrInvalidRequest` and still raised before anything touches the checkout — earlier now, if anything.

PR #461, opened 2026-07-29
feat(runway): git merger SQUASH_REBASE and MERGE

behinddwalls:
### Why?

The git merger landed with REBASE only. `SQUASH_REBASE` and `MERGE` are part of the wire contract SubmitQueue already publishes against, and until they apply here a request naming either is rejected as an invalid request. This adds the two remaining transforming strategies on top of the shared apply machinery.

### What?

**SQUASH_REBASE** applies each change exactly like REBASE — replaying every commit it introduces — then collapses what that change produced into a single commit. The squash unit is the change: a change of ten commits becomes one, and a step whose change carries several URIs yields one commit per URI. Those URIs are a stack of pull requests, and squashing them together would erase the per-PR boundary the stack exists to express.

Two degenerate cases produce no output rather than an empty commit. A change already present on the target creates no commits, so there is nothing to squash. A change that does create commits whose net tree matches the base would squash to an empty commit, so the intermediates are dropped. Both keep redelivery idempotent.

**MERGE** creates a `--no-ff` merge commit per change, which keeps the change's original commits reachable through second-parent history — the property that separates it from the picking strategies, which rewrite those hashes. A change already contained in HEAD is skipped rather than merged again.

**Not every failed merge is a conflict.** `applyMerge` previously reported any `git merge` failure as `ErrConflict`, which tells the client its change collides with the target even when nothing collided. Failures are now classified, and the case that matters in practice is an unrelated history.

**Importing an unrelated history.** A repository migration arrives as an ordinary change in the target repo whose branch carries the source repo's whole history. Being in the target repo is what makes the commits fetchable; it says nothing about ancestry, and git refuses to merge two graphs with no common ancestor. `AllowUnrelatedHistories` lifts that refusal for a queue that exists to perform such imports. It is off by default because the refusal is a genuine safeguard — with it always on, merging the wrong object silently produces a nonsense result instead of failing. Without the option, the refusal is now reported as an invalid request rather than a conflict.

MERGE is the only strategy that can serve a migration: it is the only one that preserves the imported commits' original hashes, and the picking strategies have no range to compute across disjoint graphs, so they reject such a change explicitly and say so.

## Test Plan

✅ `bazel test //runway/extension/merger/git:go_default_test` — passes (58s)

New coverage: SQUASH_REBASE collapsing a multi-commit change into one commit while landing all of its content, a two-URI change yielding one squashed commit per URI in application order, SQUASH_REBASE over an already-landed change producing none, MERGE creating one merge commit for a multi-commit change, MERGE skipping a change already an ancestor of the tip, and the MERGE dry-run path.

For migration specifically: importing a three-commit unrelated history and asserting every imported commit is reachable **under its original hash**, the merge commit has two parents, and the target keeps its own files; redelivery of the same request being a no-op; the import rejected as an invalid request (explicitly not a conflict) when the option is off, leaving the remote and checkout untouched; and both picking strategies rejecting an unrelated history rather than rewriting it.

## Stack
1. #459
1. #460
1. @ #461
1. #462
1. #463
1. #476
1. #477

### DIFF
--- a/runway/extension/merger/git/git_merger.go
+++ b/runway/extension/merger/git/git_merger.go
@@ -19,11 +19,15 @@
 // Strategy → git operation:
 //
 //   - REBASE:        cherry-pick each URI's head commit onto the target tip.
+//   - SQUASH_REBASE: cherry-pick the step's changes, then collapse them into a
+//     single commit (squash unit = the step).
+//   - MERGE:         create a --no-ff merge commit per URI, preserving the
+//     original commit hashes in second-parent history.
 //   - DEFAULT:       resolved to the instance's configured DefaultStrategy
 //     before any step runs.
 //
-// REBASE is the only strategy implemented so far. A step naming any other
-// strategy is rejected as merger.ErrInvalidRequest until its apply path lands.
+// PROMOTE is not implemented yet; a step naming it is rejected as
+// merger.ErrInvalidRequest until its apply path lands.
 //
 // Atomicity: for a committing merge nothing reaches the remote until the final
 // push. A step that fails to apply aborts the in-progress git operation and
@@ -77,8 +81,8 @@ const defaultMaxPushAttempts = 10
 
 // Default committer identity used when Params leaves it unset. The scrubbed
 // environment (GIT_CONFIG_NOSYSTEM, GIT_CONFIG_GLOBAL=/dev/null) leaves no
-// ambient identity, so the commit-creating REBASE strategy needs one supplied
-// explicitly.
+// ambient identity, so commit-creating strategies (REBASE/SQUASH_REBASE/MERGE)
+// need one supplied explicitly.
 const (
 	defaultCommitterName  = "SubmitQueue Runway"
 	defaultCommitterEmail = "runway@submitqueue.invalid"
@@ -112,7 +116,7 @@ type Params struct {
 	// Target is the destination branch ref on the remote (e.g. "main").
 	Target string
 	// DefaultStrategy resolves a step whose strategy is DEFAULT. Must be a
-	// concrete strategy (currently only REBASE).
+	// concrete strategy (REBASE, SQUASH_REBASE, or MERGE).
 	DefaultStrategy mergestrategypb.Strategy
 	// Runtime is the pinned Git runtime used for every invocation.
 	Runtime GitRuntime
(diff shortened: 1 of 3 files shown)


runway/extension/merger/git/git_merger.go
```diff
@@ -614,6 +778,52 @@ func (m *gitMerger) refetchTipSHA(ctx context.Context) (string, error) {
...
+	headTree, err := m.commitTreeSHA(ctx, "HEAD")
+	if err != nil {
+		return false, err
+	}
+	parentTree, err := m.commitTreeSHA(ctx, "HEAD^")
+	if err != nil {
+		return false, err
+	}
+	return headTree == parentTree, nil
+}
```
kevinlnew: AI-generated: `isEmptyHEADCommit` is dead code — defined but never called. `squashChange` compares trees via `commitTreeSHA` directly.

runway/extension/merger/git/git_merger.go
```diff
@@ -614,6 +778,52 @@ func (m *gitMerger) refetchTipSHA(ctx context.Context) (string, error) {
...
+
+// commitTreeSHA returns the tree SHA recorded in the commit object at ref.
+func (m *gitMerger) commitTreeSHA(ctx context.Context, ref string) (string, error) {
+	out, err := m.run(ctx, nil, "cat-file", "commit", ref)
```
kevinlnew: AI-generated nit: Parsing `cat-file commit` output manually to extract the tree SHA. `rev-parse ref^{tree}` does the same in one call and is less fragile.

kevinlnew (approved):
Lgtm. Left a couple nits.

runway/extension/merger/git/git_merger.go
behinddwalls: Correct — deleted. It was a leftover from an earlier draft of the squash path that compared HEAD to HEAD^; `squashChange` ended up comparing the pre-pick tree to the post-pick tree instead, which is the check that actually handles a multi-commit range, and the helper was never wired up. Nothing else referenced it.

runway/extension/merger/git/git_merger.go
behinddwalls: Agreed — switched to `git rev-parse <ref>^{tree}`.

Same one invocation, and it drops the hand-rolled parse along with the "unexpected first line" error branch that existed only because the old form could return something unparseable. `rev-parse` either resolves the tree or exits non-zero, so the remaining error path is the real one.

Covered by the existing squash tests — the empty-squash case (a change whose net tree matches the base) is decided entirely by this comparison.

PR #462, opened 2026-07-29
feat(runway): git merger PROMOTE

behinddwalls:
### Why?

`PROMOTE` is the last strategy in the wire contract without an apply path. It is also the one that does not fit the shared machinery: the transforming strategies build new commits locally and push `HEAD:target`, while PROMOTE advances the target to a commit that already exists, unchanged.

### What?

Adds the `promote` path, dispatched directly from `process` rather than through `applyTransforming`.

**Fast-forward only.** After resetting to the remote tip, promote classifies the named commit three ways. Already the tip, or contained in it — idempotent success, no push. A strict descendant of the tip — a genuine fast-forward, pushed as `<sha>:refs/heads/<target>`. Anything else has diverged and is a terminal `ErrConflict`; PROMOTE never creates a commit to reconcile the two.

Because it moves the ref to an existing commit, a change of any size arrives whole by construction — its ancestry comes with it, so PROMOTE needs none of the range machinery the picking strategies do.

**Exclusivity.** `resolveAndValidate` rejects a PROMOTE that is not the entire request — one step, one change, one URI — as `ErrInvalidRequest`. Two reasons, both structural: a pre-existing commit cannot descend from commits an earlier transforming step just produced, and the push targets an exact SHA rather than the locally-built HEAD, so there is nothing for a preceding step to contribute.

**Its own availability checks.** promote bypasses `tryApply`, so it performs the object-availability and staleness checks itself. Without them a commit the remote cannot supply makes every containment query fail with a plain error, which the consumer retries forever rather than reporting a request that can never succeed.

**Contention.** The same bounded retry as the transforming path, but the loop re-runs the classification rather than the apply: if the push is rejected the tip may have moved, and the commit that was a fast-forward a moment ago may now be contained (success) or divergent (conflict).

The push is a single atomic ref update, so PROMOTE needs no separate atomicity argument. A dry-run check performs the identical classification and returns without pushing, reporting no output.

With this the merger implements every strategy in the contract; `isConcreteStrategy` now admits all four.

## Test Plan

✅ `bazel test //runway/extension/merger/git:go_default_test` — passes (61s)

New cases: fast-forward promote, promote of a commit already contained in the tip, divergent promote rejected as a conflict, a multi-commit change promoted whole to the exact named commit, an unavailable commit reported as an invalid request rather than retried, both dry-run classifications, and the two composition rules (PROMOTE with a second step, PROMOTE with a second URI) rejected as invalid requests.

## Stack
1. #459
1. #460
1. #461
1. @ #462
1. #463
1. #476
1. #477

### DIFF
--- a/runway/extension/merger/git/git_merger.go
+++ b/runway/extension/merger/git/git_merger.go
@@ -23,14 +23,16 @@
 //     single commit (squash unit = the step).
 //   - MERGE:         create a --no-ff merge commit per URI, preserving the
 //     original commit hashes in second-parent history.
+//   - PROMOTE:       fast-forward the target to an already-existing commit with
+//     no content transform. PROMOTE must be the entire request (one step, one
+//     change, one URI) because a pre-existing commit cannot descend from
+//     commits produced by an earlier transforming step.
 //   - DEFAULT:       resolved to the instance's configured DefaultStrategy
 //     before any step runs.
 //
-// PROMOTE is not implemented yet; a step naming it is rejected as
-// merger.ErrInvalidRequest until its apply path lands.
-//
 // Atomicity: for a committing merge nothing reaches the remote until the final
-// push. A step that fails to apply aborts the in-progress git operation and
+// push (PROMOTE excepted, which is itself a single atomic fast-forward ref
+// update). A step that fails to apply aborts the in-progress git operation and
 // returns without pushing.
 //
 // Contention: if the push fails because the remote tip moved between reset and
@@ -116,7 +118,7 @@ type Params struct {
 	// Target is the destination branch ref on the remote (e.g. "main").
 	Target string
 	// DefaultStrategy resolves a step whose strategy is DEFAULT. Must be a
-	// concrete strategy (REBASE, SQUASH_REBASE, or MERGE).
+	// concrete strategy (REBASE, SQUASH_REBASE, MERGE, or PROMOTE).
 	DefaultStrategy mergestrategypb.Strategy
 	// Runtime is the pinned Git runtime used for every invocation.
 	Runtime GitRuntime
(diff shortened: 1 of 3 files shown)


runway/extension/merger/git/git_merger.go
```diff
@@ -570,6 +593,75 @@ func (m *gitMerger) applyMerge(ctx context.Context, step *runwaymq.MergeStep) ([
...
+		// PROMOTE does not go through tryApply, so it performs the same
+		// availability and freshness checks itself. Without them a commit the
+		// remote cannot supply turns every containment query into a plain
+		// error, which the consumer retries forever instead of reporting.
+		if err := m.ensureObjects(ctx, []changeRef{ref}); err != nil {
+			return nil, err
+		}
+		if err := m.checkStale(ctx, []changeRef{ref}); err != nil {
+			return nil, err
+		}
```
kevinlnew: AI-generated nit: `ensureObjects` + `checkStale` re-run on every retry attempt inside promote, but the SHA doesn't change between attempts. Consider moving them before the loop (before line 608).

kevinlnew (approved):

runway/extension/merger/git/git_merger.go
behinddwalls: Fixed — both moved above the loop.

You're right that the SHA is fixed: `promote` is validated to one step / one change / one URI, so the commit under promotion cannot change between attempts. Only the target tip moves, which is what the loop is actually retrying for.

Worth noting what the per-attempt cost was, since the two differ: `ensureObjects` after the first pass is just a local `cat-file -e`, but `checkStale` is an `ls-remote` — a network round trip per attempt, against a remote we retry precisely because it's busy.

While doing it I took the re-resolve out too: `promote` now reads `rs.refs[0]` instead of calling `resolveChange` on the URI again (see the threading change on #460).

The same hoist landed in `applyTransforming` on #460, so both paths now vet the request once, up front, rather than once per attempt.

PR #464, opened 2026-07-29
test(stovepipe): add build-slow marker to the fake build runner

behinddwalls:
### Why?

The fake build runner never reports a non-terminal status — `Status` decides purely from the build id and always returns succeeded, failed, or an error. That means no integration or e2e stack can reach `buildsignal`'s reschedule branch, so the entire poll loop is exercised only by unit tests with mocked publishers. A defect in the reschedule path is invisible above the unit layer.

### What?

Adds a `build-slow` marker. `Trigger` encodes a ready-at wall-clock instant into the build id (`fake-build-slow-<readyAtMs>-<suffix>`) and `Status` reports `running` until that instant passes, then `succeeded`. Encoding the deadline in the id preserves the fake's stateless, decide-purely-from-the-id property, so `Trigger` and `Status` can still live in different processes. The window is a `slowBuildDuration` field on the runner, defaulted in `New()`, so tests construct a runner with their own value instead of mutating shared state.

Also widens `marker()` to end a token at `/` as well as `&` and `#`. The fake SourceControl builds head URIs as `git://<queue>/HEAD`, so a marker injected through a queue name sits mid-URI and would otherwise be read as `build-slow/HEAD`. Existing markers sit at the end of a URI and are unaffected.

No pipeline behaviour changes; this only makes a previously unreachable path reachable in tests.

## Test Plan

✅ `bazel test //stovepipe/...` — covers `build-slow` reporting running then succeeded, the no-deadline fallback, and `marker()` against trailing path segments, query separators, and end-of-URI.

## Stack
1. @ #464
1. #465
1. #466
1. #467
1. #468
1. #469

### DIFF
--- a/stovepipe/extension/buildrunner/fake/fake.go
+++ b/stovepipe/extension/buildrunner/fake/fake.go
@@ -14,12 +14,14 @@
 
 // Package fake provides a buildrunner.BuildRunner whose outcome is driven by the
 // triggered head URI. With no marker every build immediately succeeds, behaving
-// as a best-case stub for local-stack/e2e wiring. Failures are injected by
+// as a best-case stub for local-stack/e2e wiring. Other behaviors are injected by
 // embedding a marker token in headURI of the form "buildrunner-fake=<token>":
 //
 //	buildrunner-fake=trigger-error -> Trigger returns a non-nil error
 //	buildrunner-fake=build-fail    -> Status reports BuildStatusFailed
 //	buildrunner-fake=build-error   -> Status returns a non-nil error
+//	buildrunner-fake=build-slow    -> Status reports BuildStatusRunning for a
+//	                                  short window after Trigger, then succeeds
 //
 // The runner is stateless: Trigger encodes the desired terminal outcome into the
 // returned BuildID, and Status decides the result purely from the BuildID it is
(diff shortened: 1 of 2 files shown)


stovepipe/extension/buildrunner/fake/fake.go
```diff
@@ -35,12 +37,20 @@ import (
...
 
 	"github.com/uber/submitqueue/stovepipe/entity"
 	"github.com/uber/submitqueue/stovepipe/extension/buildrunner"
 )
 
+// SlowBuildDurationMs is how long a build-slow build reports BuildStatusRunning
+// before succeeding. A var (not a const) so tests can shorten it; the local stack
+// always uses the default. It must be long enough for the caller's poll loop to
+// observe at least one non-terminal status.
+var SlowBuildDurationMs int64 = 3000
```
mnoah1: An exported var just to be overridden by tests is a bit of an anti-pattern.  We control the `runner` implementation below, so why not just set a default value New() and let tests set the struct field manually when needed.

mnoah1 (approved):

stovepipe/extension/buildrunner/fake/fake.go
behinddwalls: Agreed — dropped the exported var.

The window is now a `slowBuildDuration` field on `runner`, defaulted in `New()` from an unexported `defaultSlowBuildDuration` const. The tests are in `package fake`, so they just construct the runner directly rather than reaching for a setter or mutating shared state:

```go
slow := runner{slowBuildDuration: 30 * time.Second}
id, err := slow.Trigger(ctx, "", "…?buildrunner-fake=build-slow", nil)
```

Nicer than the save-and-restore dance the package var forced, and there's no cross-test leakage if one of them ever fails mid-way. Nothing outside the package needs to tune it — the local stack and e2e both use the default — so it stays unexported.

Also switched it from `int64` ms to `time.Duration` while it was moving, since it's a plain config field rather than an entity/wire value where the repo's int64-ms convention applies.

PR #465, opened 2026-07-29
fix(stovepipe): mint a distinct message id for each buildsignal re-poll

behinddwalls:
### Why?

`buildsignal` schedules its next poll by re-publishing to its own topic, and it reused the build id as the message id — byte-identical to the message `build` published to start the loop. The MySQL queue dedups on the `(topic, partition_key, id)` unique key and `InsertDelayed` swallows the collision with `ON DUPLICATE KEY UPDATE topic = topic`, returning success. So the reschedule was accepted and silently discarded.

This is deterministic, not a race. The re-publish happens before the delivery is acked, and GC only collects up to the minimum *acked* offset on idle ticks, so the colliding row is always still present.

The effect: any build that is not terminal on its first poll is never polled again. `Build.Status` freezes at `accepted`/`running`, the request never leaves `processing`, nothing is published to `record`, and the queue's `in_flight_count` slot is never released — so after `MaxConcurrent` such builds the queue stops admitting work entirely.

Nothing caught it because the fake build runner could not report a non-terminal status until the previous commit, and the unit tests matched the published message with `gomock.Any()`.

### What?

`publishBuildSignal` now mints `{buildID}/poll/{generation}`, where the generation is read off the id of the delivery being processed and incremented — so a chain runs `B` -> `B/poll/1` -> `B/poll/2` -> … The partition key stays the build id, so each build's poll loop keeps its own partition.

The generation advances deterministically rather than randomly, which matters for the case a random suffix handles badly. The next id is a pure function of the delivery, so a redelivery racing the original computes the *same* id and dedup collapses the two into one message: the build keeps a single poll chain. A random suffix would instead fork a second chain, doubling the poll rate and racing the first chain's status CAS for no benefit.

A stable id is not an option in the other direction either — that is the bug itself. Even a fixed-but-different id like `B/poll` only survives one extra tick, because the second tick's re-poll then collides with the message being processed.

## Test Plan

✅ `bazel test //stovepipe/...` — new unit test asserts successive re-polls mint distinct ids and never reuse the build id. Verified it genuinely regresses: reverting just the id line fails it with `"map[bk-1:{}]" should have 3 item(s), but has 1`.

✅ `bazel test //test/e2e/stovepipe/...` — new e2e ingests a queue carrying the `build-slow` marker and waits for the build row to reach `succeeded`, which requires more than one poll tick. The service log shows three distinct ids on the `buildsignal` topic for one build.

## Stack
1. #464
1. @ #465
1. #466
1. #467
1. #468
1. #469

### DIFF
--- a/stovepipe/controller/buildsignal/buildsignal.go
+++ b/stovepipe/controller/buildsignal/buildsignal.go
@@ -22,6 +22,8 @@ package buildsignal
 import (
 	"context"
 	"fmt"
+	"strconv"
+	"strings"
 
 	"github.com/uber-go/tally"
 	entityqueue "github.com/uber/submitqueue/platform/base/messagequeue"
@@ -151,7 +153,7 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) er
 	}
 
 	delayMs := pollDelay(effective)
-	if err := c.publishBuildSignal(ctx, build.ID, delayMs); err != nil {
+	if err := c.publishBuildSignal(ctx, build.ID, msg.ID, delayMs); err != nil {
 		return errs.NewRetryableError(fmt.Errorf("failed to reschedule poll for build %s: %w", build.ID, err))
 	}
 	c.logger.Debugw("rescheduled build status poll",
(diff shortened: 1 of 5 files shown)


stovepipe/controller/buildsignal/buildsignal.go
```diff
@@ -221,12 +223,22 @@ func (c *Controller) publishRecord(ctx context.Context, buildID, requestID strin
 // publishBuildSignal re-publishes buildID to buildsignal after delayMs,
 // partitioned by build id so each build's poll loop runs in its own
 // partition. A fresh message, not a nack — polling is not failure.
+//
+// The message id must be unique per tick because the queue dedups on
+// (topic, partition_key, id) and the delivery being processed has not been
+// acked yet, so its row is still present: reusing the build id would make
+// every re-poll collide with the message that scheduled it and be silently
+// discarded, ending the poll loop after one tick. The timestamp is for human
+// readability; the random suffix is what guarantees uniqueness, since two
```
roychying: why we need suffix here? if a redelivery races the original one, we will leave one build with two rows and two poll chains. letting dedup drop one seems fine? A build only needs one chain, and two just doubles the poll rate and races reconcile's CAS for no benefit?

mnoah1 (approved):

stovepipe/controller/buildsignal/buildsignal.go
behinddwalls: Good catch — a random suffix was the wrong primitive, and I've replaced it with a monotonic generation.

The reason it can't simply be a stable id, though, is that the collision isn't between two racing redeliveries — it's between the re-poll and **the message currently being processed**. That message hasn't been acked yet, so its row is still present and dedup drops the reschedule. That's the bug this PR fixes. A fixed-but-different id like `B/poll` only buys one extra tick: tick 1 consumes `B` and publishes `B/poll` fine, then tick 2 consumes `B/poll` and republishes `B/poll` into itself, and the chain dies there.

So the id has to differ between *consecutive* ticks — but it doesn't have to be unpredictable, which is where the random suffix overreached. It now derives the id from the delivery being processed:

```go
func nextPollMessageID(buildID, currentMsgID string) string  // B -> B/poll/1 -> B/poll/2 -> …
```

That gets both properties. The next id is a pure function of the current delivery, so a redelivery racing the original computes the *same* id and dedup collapses them — exactly the "one build, one chain" you're asking for — while consecutive ticks still differ, so the loop survives. Your point about doubled poll rate and racing reconcile's CAS is right, and this removes that case rather than tolerating it.

One residual worth naming: if a redelivery of generation N lands after N+1 was consumed, acked *and* GC'd, the insert succeeds and forks a chain. GC only runs on idle ticks up to the min acked offset, so that requires the live chain to have already finished — at which point the forked chain re-polls, sees terminal, no-ops on the outcome (already recorded), republishes to record (dedups), and acks. Converges harmlessly.

Verified in the e2e log — one build-slow build, one chain:

```
fake-build-slow-1785435376049-6cebfb08
fake-build-slow-1785435376049-6cebfb08/poll/1
fake-build-slow-1785435376049-6cebfb08/poll/2
```

Added `TestPublishBuildSignalIsIdempotentPerDelivery` pinning that replaying a delivery republishes the same id, plus a table test for the generation parsing.

PR #497, opened 2026-08-01
refactor(storage): replace batch state updates

ubettigole:
## Summary
Persist complete Batch entities through guarded updates and migrate controllers to failure-safe candidate copies.

## Test Plan
Unit tested

### DIFF
--- a/submitqueue/extension/storage/batch_store.go
+++ b/submitqueue/extension/storage/batch_store.go
@@ -31,10 +31,10 @@ type BatchStore interface {
 	// Returns ErrAlreadyExists if a batch with the same ID already exists.
 	Create(ctx context.Context, batch entity.Batch) error
 
-	// UpdateState updates the state of a batch to newState and the version to newVersion
+	// Update replaces every non-key field of a batch and writes newVersion
 	// if the current persisted version matches oldVersion. If versions do not match, returns ErrVersionMismatch.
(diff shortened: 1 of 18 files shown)


JamyDev (approved):

submitqueue/entity/batch.go
```diff
@@ -156,6 +156,13 @@ type Batch struct {
 	Version int32
 }
 
+// WithState returns a shallow copy of the batch with State replaced.
+// Slice fields continue to share their backing arrays.
+func (b Batch) WithState(state BatchState) Batch {
```
behinddwalls: wouldn't this also update the original state?

behinddwalls: I would probably also drop With and just say `State(...)` and may doc just need to say updates state.. it's not creating new batch but just updating state for the current batch

behinddwalls (approved):

submitqueue/entity/batch.go
ubettigole: >wouldn't this also update the original state?

No. Since it's a value receiver (`(b Batch)`, not `(b *Batch)`), it returns a shallow copy.

ubettigole: Decided we don't need a copy. Just updating `State` directly instead: `batch.State = newState`.

PR #498, opened 2026-08-01
refactor(storage): replace request state updates

ubettigole:
## Summary
Persist complete Request entities through guarded updates, migrate callers to candidate copies, and document full-entity replacement semantics.

## Test Plan
Unit tested

### DIFF
--- a/submitqueue/core/request/terminate.go
+++ b/submitqueue/core/request/terminate.go
@@ -114,6 +114,7 @@ func TerminateRequest(
 	// logVersion is the request version reflected in the published terminal log.
 	// It stays at the current version on the idempotent same-state path and
 	// advances to the new version only after a successful reconciling write.
+	beforeState := request.State
 	logVersion := request.Version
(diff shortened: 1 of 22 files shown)


JamyDev (approved):

submitqueue/entity/request.go
```diff
@@ -96,6 +96,13 @@ type Request struct {
 	Version int32 `json:"version"`
 }
 
+// WithState returns a shallow copy of the request with State replaced.
+// Slice fields continue to share their backing arrays.
+func (r Request) WithState(state RequestState) Request {
```
behinddwalls: same as the prev PR, we can drop With and update docs to not say shallow copy as it;s not and just update to state in curr batch

behinddwalls (approved):

submitqueue/entity/request.go
ubettigole: It was a copy, but we decided we don't need one. Using `request.State = newState` now.

PR #506, opened 2026-08-04
feat(orchestrator): re-plan the queue from the Speculator each run

behinddwalls:
### Why?

The speculate controller had no speculation in it: it advanced one batch at a time along a single hard-coded chain — every dependency assumed to pass — and nothing ever called the Speculator, the Generator, or the Allocator that had been built for it. This wires that machinery into the pipeline: a queue's paths are chosen by a swappable extension within a build budget, and each is built against only the dependencies it assumes will succeed.

### What?

Every message is a dirty signal naming a batch; the controller re-plans that batch's whole queue from a single read: read the state, cancel paths whose assumptions a finished dependency has proven wrong, ask the Speculator, filter its proposals, dispatch what survives. Nothing carries over between runs, so duplicated or reordered signals are harmless and a later run repairs whatever an earlier one left half-done. `doc.go` explains the model in plain terms — no vocabulary section, every term is defined where it is used.

The path set keeps exactly one writer — this run. The build stages record what CI did on per-build records; the run folds those into the set and alone decides each path's status. Pending paths are re-dispatched every run until their build is seen running, dispatches partition by batch so heads proceed in parallel, and cancelling paths need no dispatch at all: the poll loop reads the stop off the set and enacts it.

Speculation lands inert. The wiring layer passes a placeholder Speculator that proposes nothing, so the run executes end to end but funds no paths; composing real per-queue speculators and turning the feature on is the wiring change at the top of this stack. Batch outcomes still come from the legacy per-batch finalizer, which waits on every dependency — strictly stricter than path-aware finalization, so the system stays correct until the next commit replaces it.

## Test Plan
✅ `bazel test //submitqueue/orchestrator/controller/speculate/...` — assumption checks and proposal filtering are table-driven; run tests cover funding a first path, re-dispatching pending paths, broken-path cancellation, build results recorded onto paths, lost CAS races skipped rather than failed, and Speculator errors abandoning the run.

✅ `make fmt`, `make gazelle`


submitqueue/orchestrator/controller/speculate/dispatch.go
```diff
@@ -0,0 +1,213 @@
...
+		}
+
+		// Dispatch whenever the head has paths waiting to start, whether or not
+		// this run changed them: a pending path whose earlier dispatch was lost
+		// is re-sent until the build stage records it building or terminal.
+		// Cancelling paths need no dispatch at all — the poll loop reads the
+		// stop off the set and enacts it. Partitioned by batch, so heads
+		// dispatch in parallel while one head's dispatches stay ordered.
+		if hasActionablePaths(set) {
+			if err := c.publishBatchID(ctx, topickey.TopicKeyBuild, batch.ID, batch.ID); err != nil {
```
mnoah1: Definition below defines the last argument as `queue`, which I guess you'll need to keep for the future sharding.  Maybe we need a separate argument here which may be the queue or any other value that you want to use for partitioning.

submitqueue/orchestrator/controller/speculate/speculate.go
```diff
@@ -436,39 +465,23 @@ func (c *Controller) fetchDependencies(ctx context.Context, store storage.Storag
...
-// publish publishes a batch ID to the specified topic key, stamped with and
-// partitioned by the batch's queue.
-func (c *Controller) publish(ctx context.Context, key consumer.TopicKey, batchID string, queue string) error {
-	bid := entity.BatchID{ID: batchID, Queue: queue}
-	payload, err := bid.ToBytes()
+// publishBatchID publishes a batch ID to the topic behind key, stamped with
+// and partitioned by the batch's queue. The batch ID doubles as the message
+// ID, so the queue deduplicates repeat publishes for the same batch against
+// rows it has not collected yet.
+func (c *Controller) publishBatchID(ctx context.Context, key consumer.TopicKey, batchID string, queue string) error {
```
mnoah1: Add an additional arguments to support partitioning on a value other than queue (related to comment in dispatch.go):  `publishBatchID(ctx context.Context, key consumer.TopicKey, batchID, queue, partitionKey string) error`

submitqueue/orchestrator/controller/speculate/check.go
```diff
@@ -0,0 +1,169 @@
...
+
+import "github.com/uber/submitqueue/submitqueue/entity"
+
+// rejection is why one proposed action was dropped. The reasons are the metric
+// dimension for a misbehaving Speculator, so each names a distinct fault.
+type rejection string
+
+const (
+	// rejectUnknownAction is a zero-value or unrecognized action.
+	rejectUnknownAction rejection = "unknown_action"
```
mnoah1, line 25: nit: `_` prefix for string constants

submitqueue/orchestrator/controller/speculate/check.go
```diff
@@ -0,0 +1,169 @@
...
+		return "", false
+	}
+
+	// A build proposal for a path whose build already finished would discard a
+	// recorded result and start the same work again.
+	if stored && entry.Status.IsTerminal() {
+		return rejectPathTerminal, true
+	}
+
+	return "", false
```
mnoah1, line 120: Do we need to add handling for `rejectOverBudget` here?

mnoah1 (approved):
Left a few things but nothing major if you want to take a look later (other than the `publishBatchID` one).

submitqueue/orchestrator/controller/speculate/dispatch.go
behinddwalls: Good catch, and it was worse than a naming problem — it was a live bug, now fixed in 643005d.

The fourth argument used to be `partitionKey`, with the payload carrying no queue at all. When main started stamping the queue into the payload (#538), I collapsed the two roles into one `queue` parameter while resolving the rebase, so this call site's `batch.ID` silently became the payload's queue. The build controller resolves its storage from `bid.Queue` and then guards on `bid.Queue != batch.Queue`, so every build dispatch would have dead-lettered.

Split exactly as you suggested: `publishBatchID(ctx, key, batchID, queue, partitionKey string)`. This call is now `(ctx, TopicKeyBuild, batch.ID, queue, batch.ID)` — real queue in the payload, batch as the partition key so heads still dispatch in parallel. Added `TestRun_DispatchStampsQueueAndPartitionsByHead`, which fails against the old behaviour.

submitqueue/orchestrator/controller/speculate/speculate.go
behinddwalls: Done in 643005d, with your exact signature: `publishBatchID(ctx context.Context, key consumer.TopicKey, batchID, queue, partitionKey string) error`.

The doc comment now spells out why the two cannot be one parameter: `queue` is what the payload asserts the batch belongs to and is what the consumer resolves its storage from, so it must always be the real queue; `partitionKey` only decides what stays serialized behind what. Every call passes the queue for both except the build dispatch, which partitions by batch. The same split is carried into the finalization commit on top of this one, where `fanout` had the same shape.

submitqueue/orchestrator/controller/speculate/check.go
behinddwalls: I left these as-is, but happy to change them if you'd rather.

The `_` prefix for unexported globals is in the Uber Go style guide, but this repo doesn't currently follow it anywhere: I count 42 unexported package-level constants (`opName` in every controller, the constants here) and zero with a `_` prefix. Adopting it in this one file would make it the only place in the codebase using that style, so it seemed worse than the inconsistency it fixes.

If you want the convention adopted, I'd rather do it repo-wide in its own change than start here — say the word and I'll put one up.

submitqueue/orchestrator/controller/speculate/check.go
behinddwalls: Deliberately not, though it's a fair question and the boundary is worth stating.

Everything this filter rejects is a *coherence* fault it can prove from the run's own snapshot: a head it never read, a path whose assumptions don't match its head's dependency list, an assumption a finished dependency already disproved, a rebuild of a path whose build finished. Each one would corrupt state or waste a slot on something unmergeable, and each is checkable with no config.

Budget is *policy*, and it belongs to the Speculator — specifically the Allocator, which is constructed with the queue's budget and stops pulling candidates when the slots are full. The controller is never given that number, so a check here would mean plumbing per-queue budget config into the controller as a second source of truth alongside the extension that already owns it. Over-budget also isn't corrupting in the way the other rejections are: it costs extra builds, it doesn't produce an unmergeable path.

The honest gap: a misbehaving Allocator can overspend and nothing here stops it. If you'd like a backstop, I'd rather do it as a follow-up that passes the budget to the controller explicitly, so the duplication is deliberate and visible, than infer a cap here.

PR #508, opened 2026-08-04
feat(orchestrator): compose per-queue speculators and turn speculation on

behinddwalls:
### Why?

Everything below this commit landed inert: the orchestrator passed the speculate run a placeholder that proposes nothing, so no path was ever funded and no speculative build started. The machinery is all reviewed; nothing configures it per queue or switches it on.

### What?

This is the activation switch. Per-queue profiles gain a `Scorer` and a `Speculator`: each queue's speculator is composed from its own scorer as `standard.New(bestfirst.New(scorer), sticky.New(budget))`, with a factory adapter beside the existing ones so routing stays in the wiring layer, and `main.go` swaps the placeholder for `profiles.SpeculatorFactory()`.

The scorer profiles are the policy knobs: the baseline scores everything 0.5, test-queue buckets by lines changed, and e2e-test-queue exercises the composite scorer. Every scorer is wrapped by `scorerfake` so a change URI carrying a failure marker forces a scoring error end-to-end. The build budget is a wiring-level constant (4 concurrent builds per queue) with a TODO to move it onto `entity.QueueConfig`.

With this commit, paths are funded, builds run per path, and batches finalize from their paths — the whole stack goes live in one revertable step.

## Test Plan
✅ `bazel build //...`, `make fmt`, `make gazelle`

✅ `make e2e-test` — the full pipeline lands changes end-to-end with per-path speculation on.


mnoah1 (approved):

behinddwalls:
The e2e timeout here is not flakiness — speculation never funds a path, so no build is ever dispatched and the test waits for a batch that can never land, while the failing messages retry (that run spilled ~11MB of stdout).

Two producers build an entity without naming its queue, and both of their stores are bound to one queue and reject a write that disagrees:

- `speculate/dispatch.go` created a head's first `SpeculationPathSet` as `{Head: batch.ID}` — this is the first write of every newly funded head, so nothing is ever funded.
- `build/build.go` created the `PathBuild` link without its queue, so a path is never observed moving to `building` and gets re-dispatched on every run.

Confirmed by calling each store the way its producer does:

```
path build pathID=p1 attempt=1 queue "" does not match the store's bound queue "monorepo"
speculation path set head=monorepo/batch/1 queue "" does not match the store's bound queue "monorepo"
```

Both stores gained the guard when their tables were re-keyed to lead with the queue (#543 and #502) and the producers weren't updated to match — my miss in the #502 rebase. Unit tests use the gomock stores, which don't enforce the binding, and the storage contract suite builds its own fixtures with the queue set, so only the e2e could catch it.

Fix plus regression tests are in #551, which targets `main` since both bugs are already merged. This PR's e2e should go green once that lands.

PR #513, opened 2026-08-05
docs(rfc): explain best-first speculation generation

behinddwalls:
### Why?

The best-first generator needs a durable explanation of how it ranks paths lazily without enumerating every dependency-outcome combination.

### What?

Document the Generate and Next flow: the strict snapshot contract, preferred assumptions and flip costs, per-head streams that sort flips only when a head is first pulled, the extend/swap subset walk, the global k-way merge over each head’s current best, logarithmic scoring, tie-breaks, per-pull complexity, and rejected alternatives.

### DIFF
--- a/doc/rfc/submitqueue/speculation-generator-best-first.md
+++ b/doc/rfc/submitqueue/speculation-generator-best-first.md
@@ -1,0 +1,507 @@
+# Best-First Speculation Path Generation
+
+## Summary
+
+The default generator returns the most likely complete build path across the whole queue, one path per call.
+
+The code has one setup method, `Generate`, and one repeated method, `Next`:
+
+1. `Generate` validates the queue snapshot, then records a preferred assumption and the cost of flipping to its opposite for each unresolved direct dependency. It totals the best score for each eligible head; resolved dependencies remain fixed facts.
+2. `Generate` pushes one lightweight best-path candidate per head into a global heap. Each head also owns a stream that enumerates its remaining paths on demand; at this point the stream has done no work beyond that total.
+3. `Next` removes the highest-ranked candidate from the global heap, advances only that head's stream, inserts the head's next candidate, and constructs and returns the complete path that was removed.
+
+The global heap always contains only the current best candidate from each head, so removing its top is a k-way merge over per-head best-first streams: it yields the best path across the whole queue however the heads interleave, without ever comparing more than one candidate per head.
+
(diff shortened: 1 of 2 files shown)


doc/rfc/submitqueue/speculation-generator-best-first.md
```diff
@@ -0,0 +1,509 @@
...
+
+The ordering stays the same. `CandidatePath.RankingScore` contains this logarithmic value. It is recomputed for every run and is not compared across runs.
+
+## Completed and missing dependencies
+
+- `Succeeded` fixes an assumption to succeeds.
+- `Failed` or `Cancelled` fixes an assumption to fails.
+- `Cancelling` remains undecided because cancellation may lose a race with completion.
+- A fixed assumption stays in the returned path but contributes probability 1 and has no flip.
+- A dependency missing from the input snapshot uses a success probability of 0.95.
```
sbalabanov: how can it happen? All batches should be somehow scored?

doc/rfc/submitqueue/speculation-generator-best-first.md
```diff
@@ -0,0 +1,509 @@
...
+- A shared dependency is scored once per run.
+
+## Why the algorithm works
+
+- Every candidate in the heap represents a complete path that can be returned immediately.
+- Sorting flips cheapest-first ensures a child never outranks its parent.
```
sbalabanov: does it mean for each batch it needs to build all the combinations of flips first, isn't the computational complexity O(2^n) where n is the number of unresolved dependencies? Can it become a problem?

doc/rfc/submitqueue/speculation-generator-best-first.md
```diff
@@ -0,0 +1,509 @@
...
+```
+C [A succeeds, B succeeds]
+C [A succeeds, B fails]
+C [A fails, B succeeds]
+C [A fails, B fails]
+```
+
+Those extra A branches over-speculate. C does not directly conflict with A, so A does not create a separate way to build C; its effect already flows through B. A longer chain makes the problem much worse: if the full closure contains n ancestors, a head with one direct dependency grows from two paths to 2ⁿ paths.
+
+The transitive closure is also not part of the generator's input contract. Stored dependency lists, path validation, path identity, merge checks, and outcome checks all use direct dependencies. We reject the DAG walk because it creates unnecessary paths and would require changing those contracts.
```
sbalabanov: this paragraph is unnecessary and in general, a slop

sbalabanov (approved):

PR #518, opened 2026-08-05
feat(stovepipe): add Queue.LastGreenRequestID in Stovepipe Queue

roychying:
## Why?
The `record` stage advances the queue's last-green bookmark, and that bookmark must only ever move forward. `entity.Queue` today carries `LastGreenURI` but no record of *which* request established it, so there is nothing to compare a candidate against — a late-arriving delivery for an older request could regress the bookmark to a stale commit, which `process` would then hand to `build` as an incremental baseline.

 Adding the owning request id gives the guard something to compare, using the same ingest-order comparison (`CompareRequestID`) that `ingest` and `process` already use for coalescing.
## What?
 - `entity.Queue` gains `LastGreenRequestID`, paired with the existing `LastGreenURI`.
  - The MySQL `queueStore` threads it through `Create`, `Get`, and `Update`.
  - `schema/queue.sql` gains `last_green_request_id VARCHAR(255) NOT NULL DEFAULT ''`, so existing rows read back as "bookmark never set" and the first green outcome adopts unconditionally.

## Test Plan
make gazelle, make build, make test - all clean

### DIFF
--- a/stovepipe/extension/storage/mysql/schema/queue.sql
+++ b/stovepipe/extension/storage/mysql/schema/queue.sql
@@ -1,10 +1,11 @@
 -- queue holds per-queue coordination state for the validation pipeline: the last-green
 -- bookmark, in-flight gate count, and latest-request id pointer.
 CREATE TABLE IF NOT EXISTS queue (
-    name               VARCHAR(255) NOT NULL,
-    last_green_uri     VARCHAR(255) NOT NULL DEFAULT '',
-    in_flight_count    INT          NOT NULL DEFAULT 0,
-    latest_request_id  VARCHAR(255) NOT NULL DEFAULT '',
-    version            INT          NOT NULL,
+    name                   VARCHAR(255) NOT NULL,
+    last_green_uri         VARCHAR(255) NOT NULL DEFAULT '',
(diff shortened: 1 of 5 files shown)


stovepipe/extension/storage/mysql/schema/queue.sql
```diff
@@ -1,10 +1,11 @@
...
 -- bookmark, in-flight gate count, and latest-request id pointer.
 CREATE TABLE IF NOT EXISTS queue (
-    name               VARCHAR(255) NOT NULL,
-    last_green_uri     VARCHAR(255) NOT NULL DEFAULT '',
-    in_flight_count    INT          NOT NULL DEFAULT 0,
-    latest_request_id  VARCHAR(255) NOT NULL DEFAULT '',
-    version            INT          NOT NULL,
+    name                   VARCHAR(255) NOT NULL,
+    last_green_uri         VARCHAR(255) NOT NULL DEFAULT '',
+    last_green_request_id  VARCHAR(255) NOT NULL DEFAULT '',
```
mnoah1: Add new columns at the end for easier migration compatibility

mnoah1 (approved):

PR #523, opened 2026-08-06
feat(storage): add QueueBatchStateStore and shared batch state helpers

behinddwalls:
### Why?

`BatchStore.GetByQueueAndStates` is the storage contract's only query-by-attribute, backed by `idx_queue_state` — the only secondary index in the entire schema. The storage README's key-value contract flags exactly this shape: a genuinely needed reverse lookup should be a first-class mapping store keyed by the lookup attribute, so any backend (SQL, DynamoDB, Bigtable) can serve it as a primary-key read. This PR adds that store and the shared helpers controllers will migrate onto; `speculate/run.go` already carries a TODO for this replacement.

### What?

- `entity.QueueBatchState` — an advisory membership record filing an in-queue batch under a (queue, state) bucket, existing from batch creation until the batch exits the queue (through terminal states, hence not "active"). Also adds `entity.AllBatchStates()` for the future conclude-time sweep.
- `storage.QueueBatchStateStore` — `List(queue, state)` / `Put` / `Delete`, all idempotent. The MySQL impl is backed by a new `queue_batch_state` table whose PK (queue, state, batch_id) *is* the lookup: listing a state bucket is a PK-prefix scan, no secondary index.
- `submitqueue/core/batch` — the shared protocol primitives: `Transition` (batch CAS, then Put the new-bucket record before Deleting the old one, so a batch always has at least one record), `EnsureRecord` (idempotent repair for redelivery skip branches), and `ListByStates` (scan requested buckets, dedupe, hydrate by key with bounded concurrency, classify by the hydrated authoritative state — a stale record can misplace a batch but never misreport it).
- Contract-suite coverage for the new store and a README pointer beside the `ChangeRecord` mapping-store example.

Scoped to the foundation only: the store has no callers yet, so there is no runtime behavior change. Follow-ups migrate the six batch-transition sites and the two `GetByQueueAndStates` readers onto the helpers, add conclude-time record deletion, and then remove `GetByQueueAndStates` + `idx_queue_state`.

## Test Plan

- ✅ `bazel test //submitqueue/entity:all //submitqueue/core/batch:all //submitqueue/extension/storage/...`
- ✅ `bazel test //test/integration/submitqueue/extension/storage/mysql:go_default_test` (real MySQL; includes the new `TestStorage_QueueBatchStateRecordLifecycle` contract case)
- ✅ `make tidy` / `make gazelle` / `make fmt` / `make mocks` — all stable (re-running produces no diffs)
- Note: `make test` failures in `orchestrator/controller/batch` and `controller/cancel` are pre-existing — they reproduce at clean HEAD db51df4d in a fresh worktree.


submitqueue/core/batch/list.go
```diff
@@ -0,0 +1,89 @@
...
+			seen[record.BatchID] = true
+			ids = append(ids, record.BatchID)
+		}
+	}
+
+	hydrated := make([]entity.Batch, len(ids))
+	g, gctx := errgroup.WithContext(ctx)
+	g.SetLimit(hydrateConcurrency)
+	for i, id := range ids {
+		g.Go(func() error {
```
ubettigole, line 69:
```suggestion
	for i, id := range ids {
	    i, id := i, id
		g.Go(func() error {
```
This could cause a bug in Go v1.21 and earlier. Go 1.22 was released in early 2024, so we'd probably be fine at Uber, but given this is OSS, we may want to make sure it works for earlier versions. In those earlier versions, for-loops reused the memory addresses of the loop variables (`i` and `id`), so by the time we use `i` on line 74 (after a DB query), the memory address very well could've changed, and you'd insert into the wrong index. Creating local variables inside the loop with the same values solves the problem.

ubettigole (approved):
Looks good, just one small comment

submitqueue/core/batch/list.go
behinddwalls: Verified — safe as written, for a subtle reason: whether a `for` loop gets per-iteration variables is decided by the module's language version (the `go` directive in go.mod), not by the toolchain that happens to compile it, and this module declares `go 1.24.5`.

Two consequences:
- Every toolchain that can build this module compiles the loop with per-iteration `i`/`id` — the 1.22 loopvar change applies to any file in a module whose `go` directive is ≥1.22.
- Toolchains predating the change cannot produce the buggy build: 1.21–1.23 either auto-download a 1.24.5 toolchain (`GOTOOLCHAIN=auto`, the default) or refuse with `requires go >= 1.24.5`, and ≤1.20 fails even parsing the three-part `go 1.24.5` directive.

So OSS consumers building from source are covered by the same gate, and adding `i, id := i, id` would be a redundant copy that `copyloopvar`-style linters flag for removal on ≥1.22 modules. If we ever lower the `go` directive below 1.22 the copy would become necessary — but that would trip the same review on every `g.Go` closure in the repo, so I'd rather keep the directive as the single guard.

PR #533, opened 2026-08-06
fix(messagequeue): release leases on drained partitions

behinddwalls:
### Why?

A lease is held forever once acquired: `ReleaseLease` is only called on shutdown and fair-share shedding, renewal is unconditional, and workers are reconciled from the lease set — so when a partition drains (every message acked and garbage-collected, hence absent from discovery), its holder keeps a goroutine polling an empty partition at ~4-5 queries per poll tick, permanently, plus a lease row and a `queue_offsets` row that nothing ever deletes. On topics with bounded partition keys (queue names) this is harmless stickiness; on topics with unbounded short-lived keys it is a leak proportional to cumulative traffic: buildsignal partitions per batch ID, and the gateway log and cancel topics partition per request ID — every request ever processed would leave a ghost worker behind. The crash variant is worse: a stale lease on a drained partition is never even stealable, because acquisition only probes discovered partitions.

### What?

- Idle-lease release: the discovery tick tracks, per owned partition absent from discovery, when it was first observed drained (a partition with any in-flight, postponed, or unacked message still has stored rows and is always discovered — only fully-consumed partitions qualify). After a grace of 2x `LeaseDurationMs` (60s at defaults) the subscriber deletes its consumer group's offsets row (while still holding the lease, so no concurrent initialization is possible), releases the lease, and reconciliation stops the worker. If a message arrives later, the partition reappears in discovery and is reacquired like any new partition — `Initialize` recreates the offsets row, and since drained meant zero stored rows there is nothing to replay.
- Stale-lease purge: the lease tick deletes lease rows not renewed within 10x `LeaseDurationMs`, covering holders that crashed while owning a drained partition. Deleting a stale row is equivalent to expiry — a concurrent renewal refreshes the row and the age predicate skips it.
- Rebalance integration tests pin `Retry.MaxAttempts` high: they publish messages they never ack, and dead-lettering mid-test would now drain the partitions and dissolve the lease distribution their assertions wait on.

## Test Plan

- ✅ New integration test `TestIdleLeaseRelease` covers the full lifecycle against real MySQL: consume, wait out GC + grace, assert the lease and offsets rows are gone, then republish to the same partition key and assert delivery resumes through normal discovery.

### DIFF
--- a/platform/extension/messagequeue/mysql/subscriber.go
+++ b/platform/extension/messagequeue/mysql/subscriber.go
@@ -61,6 +61,23 @@ const (
 	// Purging a live-but-stalled subscriber's row is harmless: its next
 	// heartbeat re-inserts it.
 	heartbeatPurgeAfterLeaseDurations = 10
+
+	// idleLeaseReleaseAfterLeaseDurations sets how long an owned partition
+	// must stay drained (no stored messages, hence absent from discovery)
+	// before its lease is released, as a multiple of LeaseDurationMs (2x =
+	// 60s at defaults). Long enough to keep ownership sticky across brief
+	// quiet spells on stable partition keys; short enough that short-lived
+	// keys (one batch, one request) don't hold leases, offset rows, and
+	// polling workers forever. A released partition rejoins through normal
+	// discovery as soon as a message arrives.
+	idleLeaseReleaseAfterLeaseDurations = 2
+
+	// leasePurgeAfterLeaseDurations sets the age threshold for purging
+	// abandoned lease rows, as a multiple of LeaseDurationMs. Covers holders
+	// that crashed while owning a drained partition: acquisition only probes
+	// discovered partitions, so nothing would ever steal (and thereby
+	// refresh or remove) a stale lease on a partition with no messages.
+	leasePurgeAfterLeaseDurations = 10
 )
 
 // HookSignal identifies the type of subscriber lifecycle event.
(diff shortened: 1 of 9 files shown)


platform/extension/messagequeue/mysql/subscriber.go
```diff
@@ -650,10 +682,103 @@ func (s *subscriber) discoverAndReconcileWorkers(ctx context.Context, sub *subsc
...
+				// Retried next tick — the lease is still held, so the
+				// partition stays tracked as drained.
+				s.logger.Errorw("delete offsets for drained partition failed",
+					"topic", sub.topic,
+					"partition_key", pk,
+					"error", err,
+				)
+				continue
+			}
+			if err := s.leaseStore.ReleaseLease(ctx, sub.topic, pk, cfg.SubscriberName, cfg.ConsumerGroup); err != nil {
```
roychying, line 713: nit: shall we stop the worker immediately after each release here? currently it happens in `reconcilePartitionWorkers` after the whole batch processed.

I saw `rebalance` does kill the worker after each release, with an explicit comment about preventing duplicate processing.

roychying (approved):

platform/extension/messagequeue/mysql/subscriber.go
behinddwalls: Done — worker is now stopped immediately after each release, mirroring the shed path in rebalance. One nuance worth noting: on this path the partition is drained (zero stored rows — that's what makes it eligible), so unlike the shed path there's normally nothing left to double-process. The case the immediate stop does cover is the resurrection race: a message published in the window right before the release makes the partition acquirable by another subscriber immediately, and the old worker shouldn't poll alongside the new owner. Added a comment in the code to that effect.

PR #534, opened 2026-08-06
feat(orchestrator): adopt QueueBatchStateStore for batch state reads and writes

behinddwalls:
### Why?

`BatchStore.GetByQueueAndStates` was the storage contract's last query-by-attribute, backed by the schema's only secondary index (`idx_queue_state`). The queue's per-state membership records (`QueueBatchStateStore`) exist to replace it with pure by-key reads, but had no callers yet.

### What?

Every batch state write now goes through the shared transition protocol in `submitqueue/core/batch`: `Transition` in the batch (Creating→Created), speculate (Speculating/Merging/Failed/Cancelled), mergesignal (terminal), cancel (Cancelling), and DLQ (Failed) controllers, with `EnsureRecord` repairing records after batch creation and on idempotent redelivery skip branches. The batch controller's dependency read goes through `core/batch.ListByStates` over the record buckets. `GetByQueueAndStates` is deleted from the contract, MySQL implementation, and mock, and `idx_queue_state` is dropped from the batch schema, so the batch table is pure get/put-by-key. Also repairs three stale controller tests that no longer matched the current control flow (they fail at the parent commit as well).

## Test Plan

✅ `go test ./...` (only pre-existing runway git-merger environment failures remain; they fail at the parent commit too) ✅ `make fmt` ✅ `make lint` ✅ `make gazelle` ✅ `make mocks` ✅ storage + orchestrator integration suites via `make integration-test` targets.

## Stack
1. @ #534
1. #527
1. #535


behinddwalls:
Duplicate — this branch already landed on main via #525 (squash a08ae646); arh re-opened it against a stale local stack. The follow-up work continues in #527 and #535.

PR #544, opened 2026-08-07
feat(runway): resolve merge targets per queue and provision their checkouts

behinddwalls:
### Why?

Three things stopped Runway from performing a real merge. It resolved one merger for the whole process, so a deployment could serve only one repository. Nothing created the checkout the git merger requires, which it declares must already exist. And the container had no `git` binary, while the merger demanded three absolute runtime paths — so enabling it failed at startup with `git runtime executable is required` rather than working.

### What?

Merge targets come from `MERGE_CONFIG_PATH`, a YAML file with a `defaults` block and per-queue overrides. Each git target is provisioned at startup — repository initialised, remote configured, credential written, target branch checked out — idempotently, so a restart against a persisted volume costs nothing and a rotated token takes effect. The existing `MERGE_*` environment variables still configure a single target when no file is given.

Two queues naming one checkout resolve to the *same* merger instance, which is what serializes them: a git merger locks the working tree it owns, and two instances over one tree would reset it out from under each other mid-merge. That instance is built from whichever queue is reached first, so queues sharing a checkout have to agree on every field it is built from — not merely on the target. Any disagreement is rejected at startup, naming the field, because the alternative is a queue that asked for `SQUASH_REBASE` quietly merging with `REBASE` and reporting nothing. The comparison is structural rather than a list of fields to remember, so a field added to the merger later is covered on its own.

The credential never enters the remote URL. The merger folds git's stderr into the errors it returns, so a URL-embedded token would be reprinted into logs and dead-letter payloads by any failed fetch. It is written as an HTTP `Authorization` header into a `0600` config fragment the repository includes — which keeps it off the command line too. SSH needs no code at all: the merger already passes `SSH_AUTH_SOCK` and `GIT_SSH_COMMAND` through its scrubbed environment.

The git runtime is derived from the installed git when unset — executable from `PATH`, exec path from `git --exec-path`, templates from the matching install prefix — so the pinning safeguard stops being a boot-time trap. `git` is installed in the image.

## Test Plan

✅ `bazel test //service/runway/server:go_default_test` — provisioning runs against real git (pinned, as the merger does): clone into an empty directory, idempotent re-run, drifted remote URL corrected, non-default target branch, credential written `0600` and absent from the remote URL, stale fragment removed, unset token rejected, and git-runtime derivation with and without the environment overrides.

✅ Sharing a checkout is now judged on the whole merger configuration, not just the target: two queues that agree everywhere still share one instance, two that disagree in `defaultStrategy`, `updateHeadBranch`, `checkStaleness`, `maxPushAttempts`, the committer identity or the refspecs are rejected at load, and a queue spelling a default explicitly still counts as agreeing.

### DIFF
--- a/service/runway/server/main.go
+++ b/service/runway/server/main.go
@@ -170,7 +170,7 @@ func run() error {
 		gate,
 	)
 
-	mergerFactory, err := newMergerFactory(logger, scope.SubScope("merger"))
+	mergerFactory, err := newMergerFactory(ctx, logger, scope.SubScope("merger"))
 	if err != nil {
 		return fmt.Errorf("failed to create merger factory: %w", err)
 	}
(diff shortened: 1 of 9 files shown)


service/runway/server/main.go
```diff
@@ -304,76 +304,204 @@ func run() error {
...
+	// held here so ids stay unique across every queue.
+	seq *atomic.Uint64
+}
+
+func (b *mergerBuilder) build(cfg mergerConfig, where string) (merger.Factory, error) {
+	if cfg.Type != mergerTypeGit {
+		return &noopMergerFactory{seq: b.seq}, nil
+	}
+
+	if existing, ok := b.byTarget[cfg.CheckoutPath]; ok {
```
mnoah1, line 465: The `byTarget` cache is keyed on `CheckoutPath` alone, but `sameTarget` only compares `RemoteURL`, `Remote` and `Target`. So two queues sharing a checkout but declaring different `defaultStrategy`, `updateHeadBranch`, `checkStaleness`, `maxPushAttempts` or committer identity pass validation, and the second silently gets the first one's values. A queue asking for `SQUASH_REBASE` quietly merging with `REBASE` is a wrong-history outcome that no error surfaces. Either widen `sameTarget` to full-config equality, or reject two entries that share a checkout and differ in any field the merger consumes.

mnoah1 (approved):
Approving - comment below seems like an edge case we may want to consider.

service/runway/server/main.go
behinddwalls: Confirmed, and it was worse than an edge case — thanks for catching it.

I reproduced it before fixing: two queues sharing `/var/checkouts/r`, one `SQUASH_REBASE` with `updateHeadBranch: true`, the other `REBASE` with it false. `loadMergeConfig` returned no error, and the second queue would have merged with the first's strategy. The two halves you identified are exactly right — `sameTarget` compared 3 fields while `NewMerger` is built from 11.

Fixed by widening the check, but reflectively rather than by listing fields. The old comparison was correct when the merger consumed three fields and quietly stopped being correct as it grew; enumerating them again would just reset that clock. `firstDifference` now walks the struct, so a field added to `mergerConfig` later is covered without anyone remembering. It reports the first disagreement by its YAML name, since "these queues differ somewhere" isn't actionable across fifteen fields:

```
queue "b" reuses checkout "/var/checkouts/r" but differs from an earlier queue in
defaultStrategy (SQUASH_REBASE vs REBASE); queues sharing a checkout share one
merger instance, so give each configuration its own checkout
```

Two details worth flagging:

- Comparison happens after `normalizeAndValidate`, so a queue omitting a defaulted field still agrees with one stating it explicitly. I also collapse an empty `fetchRefspecs` to nil so `[]` and omitted don't read as a difference — otherwise the stricter check would reject configurations that mean the same thing.
- `DeepEqual` on `*bool` compares pointed-to values, so two queues that both say `checkStaleness: true` still share.

I left the `byTarget` cache keyed on `CheckoutPath` and documented why that's now safe: validation is the gate, and both config paths (`MERGE_CONFIG_PATH` and the `MERGE_*` environment) run through it before the builder sees anything. The comment says so, so the next person to widen what may share a checkout knows to widen the check with it.

Tests: the reject table covers `defaultStrategy`, `updateHeadBranch`, `checkStaleness`, `maxPushAttempts`, committer identity and refspecs; a separate test pins that equivalent spellings still share. `TestMergerConfig_FirstDifference` asserts the returned field name directly rather than matching on the error string.

PR #547, opened 2026-08-07
feat(gateway): land and status subcommands for the client

behinddwalls:
### Why?

The gateway client only knew how to ping. Submitting a change meant hand-assembling protobuf with grpcurl, including a 40-character commit SHA copied out of a pull request — which is both tedious and easy to get subtly wrong.

### What?

`ping | land | status` subcommands. `land` takes repeated `-uri` flags, whose order is significant: several changes in one request are a stack, applied in the order given.

`-pr <url>` resolves a pull request's head commit and mints the change URI, so nobody types a SHA. It dispatches on the change URL the same way the merger and the change provider dispatch on the URI scheme, rather than assuming one provider; GitHub is implemented, and adding another is a case beside it. Reading a public repository needs no token, and a 404 without one says so rather than reporting the pull request as missing.

## Test Plan
✅ `go build ./service/submitqueue/gateway/client` and `go vet`.

✅ Manually: `land -queue q` reports the missing-change error; `land -pr https://github.com/uber/cadence/pull/1` resolved against the live GitHub API and printed `github://github.com/uber/cadence/pull/1/65c322d4...` before failing to dial a gateway, which is the expected end of that path with no stack running.


kevinlnew:
The diff includes a compiled binary `client` at the repo root:

```
diff --git a/client b/client
new file mode 100755
index 00000000..5db01cb2
Binary files /dev/null and b/client differ
```

Looks accidental — should be removed and probably added to `.gitignore`.

mnoah1 (dismissed):
Is the committed client (18mb) intentional here?

behinddwalls:
> The diff includes a compiled binary `client` at the repo root:
> 
> ```
> diff --git a/client b/client
> new file mode 100755
> index 00000000..5db01cb2
> Binary files /dev/null and b/client differ
> ```
> 
> Looks accidental — should be removed and probably added to `.gitignore`.

Yes..let me remove it..i guess claude just sent it anyway

behinddwalls:
Removed in `d39e85c3` — thanks both, and sorry for the noise.

**Cause:** this PR's own test plan ran `go build ./service/submitqueue/gateway/client`. That writes an executable named after the package directory into the *current* directory — run from the repo root it drops a 19,410,370-byte `client` there, which then got swept into the commit. The repo builds with Bazel (`bazel run //service/submitqueue/gateway/client:gateway`), so nothing in the normal workflow produces this file; the `go build` was the mistake.

**Fix:** binary deleted and a `.gitignore` entry added. One detail worth flagging — the entry is root-anchored as `/client`, not a bare `client`. A bare pattern would also match the `client/` **source** packages (`service/submitqueue/gateway/client/`, `service/submitqueue/orchestrator/client/`, and `submitqueue/client/` incoming in #570) and silently stop tracking them.

**No history rewrite:** `main` squash-merges, so the add+delete collapses to nothing and the blob never reaches `main`. It already exists in this PR's refs on GitHub regardless, so rewriting the branch would not remove it from there either — it would only force a restack of the branches above this one.

behinddwalls:
Follow-up in #587, stacked on this branch: adds `/server` beside `/client` (every `service/*/server` package builds to a root `server` the same way this one built a root `client`), and adds `make build-submitqueue-gateway-client` / `-server` writing to `bin/client` and `bin/server`.

That gives a sanctioned way to produce a standalone binary. `bin/` was already ignored but had no producer — the only reference to it anywhere was `rm -rf bin/` in `clean`.

Keeping this PR to just the removal so it stays easy to review.

kevinlnew (approved):

PR #552, opened 2026-08-07
feat(stovepipe): add the ValidationFact entity and its create-only queue-scoped store

roychying:
## Why?
`record` currently derives greenness straight from `request.State` and writes nothing durable beyond the queue's last-green bookmark. The record RFC (#394) calls for an immutable **ValidationFact** per validated commit — "how broken was this scope at this URI?" — with every decision after the fact is written reading the fact rather than the request.

This PR adds the entity, the store contract, and the MySQL implementation.

## What?
- **`entity.ValidationFact`** — identity `(queue, uri, project)`; carries `Degree`, `RequestID`, `CreatedAt`. Plus `DegreeGreen`/`DegreeBroken` constants and `IsGreen()`, so the float comparison lives in one place instead of `Degree == 0` at call sites.
- **`ValidationFactStore`** — `Create` (returns `ErrAlreadyExists`) and `Get` (returns `ErrNotFound`). No `Update`: facts are create-only and first-writer-wins, so the first fact for an identity is the permanent answer.
- **MySQL impl + `validation_fact.sql`**, wired into the queue-scoped `Storage` aggregate.


stovepipe/entity/validation_fact.go
```diff
@@ -0,0 +1,58 @@
...
+	// URI is the commit the fact describes, in the queue's VCS-agnostic locator
+	// form. Unique per (queue, URI, project).
+	URI string `json:"uri"`
+	// Project is the scope the degree applies to. Empty means the whole
+	// repository; named projects narrow the fact to part of it.
+	Project string `json:"project"`
+	// Degree is how broken the scope was, on the closed interval
+	// [DegreeGreen, DegreeBroken]. Intermediate values describe partial
+	// breakage.
+	Degree float64 `json:"degree"`
```
behinddwalls, line 47: how can this be in between for projects ever? should it be always 0 or 1

roychying: In #394 Record RFC: _Intermediate degrees are reserved for project analysis and deferred with the project mapping contract; Phase 1 does not manufacture fractional values._

if this is true, I believe in the end we still need intermediate degrees so make it as float directly..

behinddwalls: sg

behinddwalls (approved):

PR #565, opened 2026-08-11
feat(stovepipe): reconcile dead-lettered build signals and free the slot

mnoah1:
## What?
Add buildsignal dlq controller implementation

## Why?
A buildsignal message that dead-letters ends the only poll chain watching a build that is still running, and its request keeps holding one of the queue's in_flight_count slots. With no reconciler on that topic the count stays high for good, so the queue loses a slot per incident until it can no longer admit work.

Add the buildsignal DLQ controller, which maps the dead-lettered build back to its request, marks the request failed, and releases the slot, and register it plus its topic in the reference server.

Also document in the RFC that the stage's "classifier decides" disposition for Status failures only works if the BuildRunner backend classifies its own transport and HTTP errors. Unclassified, they take the non-retryable default, which is what sends a poll message to the DLQ on the first proxy blip.

## Test Plan
- Integrate in go-code's stovepipe service
- Monitor for cases of buildsignal remaining stuck on a poll error

### DIFF
--- a/stovepipe/controller/dlq/buildsignal.go
+++ b/stovepipe/controller/dlq/buildsignal.go
@@ -138,6 +138,10 @@ func (c *BuildSignalController) Process(ctx context.Context, delivery consumer.D
 		return nil
 	}
 
+	// Every request reachable from a build row is either still processing, and holding
+	// the slot failRequest releases, or already terminal, and past releasing it: build
+	// triggers only once process has written the strategy, which lands in the same CAS
+	// as accepted→processing, and processing exits only to a terminal outcome.
 	if err := failRequest(ctx, store, c.logger, build.RequestID); err != nil {
 		metrics.NamedCounter(c.metricsScope, _buildSignalOpName, "reconcile_errors", 1)
 		return err
--- a/stovepipe/controller/dlq/buildsignal.go
+++ b/stovepipe/controller/dlq/buildsignal.go
@@ -1,0 +1,163 @@
+// Copyright (c) 2025 Uber Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


stovepipe/controller/dlq/buildsignal.go
```diff
@@ -0,0 +1,163 @@
...
+
+	if build.RequestID == "" {
+		// Defensive: a build with no request has nothing to reconcile and no slot
+		// to release. Ack it so the DLQ does not grow forever.
+		c.logger.Errorw("dlq reconcile: build has empty request id, skipping", "build_id", sig.Id)
+		metrics.NamedCounter(c.metricsScope, _buildSignalOpName, "build_missing_request", 1)
+		return nil
+	}
+
+	if err := failRequest(ctx, store, c.logger, build.RequestID); err != nil {
```
roychying, line 145: would need to update `failRequest` since it only frees the slot for request in `Processing` state?

mnoah1: Checked this and I don't think it needs to widen — `Processing` is the only non-terminal state that can own a slot, so the condition isn't narrower than the set of slot holders:

- `process` claims the slot and then CAS-marks `accepted`→`processing`, releasing its own claim if that CAS doesn't land (`admitLatestHead`), so an `accepted` request holds nothing.
- The only exits from `processing` are the terminal outcomes, and those release the slot themselves (`buildsignal.finishRequest`).

Widening it to `accepted` would decrement for the much more common request that never claimed a slot, which over-admits against `MaxConcurrent` — the opposite failure of the one this PR fixes.

On this path it's tighter still: a build row only exists once `build` triggered, and `build` only triggers when the strategy is visible, which lands in the same CAS as `accepted`→`processing`. So the request behind a dead-lettered signal is either still `processing` (slot held, released here) or already terminal (slot already released, skipped).

The one case that escapes both this reconciler and `process`'s compensation is a hard crash between the two admit writes: that leaves an `accepted` request holding a slot, indistinguishable from one that never claimed it without per-request slot ownership on the row. Happy to file that as a follow-up if you'd like it tracked.

085d43c writes the invariant down where `failRequest` gates on the state and at the call site here, and pins it on the `accepted` test case (which passes no queue expectations, so a widening breaks it).

roychying: sounds good

behinddwalls (approved):

PR #566, opened 2026-08-11
feat(errs): http error classifier

mnoah1:
## What

platform/errs treats an unclassified error as non-retryable, so every HTTP failure an extension returned took that default. A 502 from a proxy in front of a build API was indistinguishable from "this request is invalid": the message dead-lettered on its first attempt instead of being retried.

Add `platform/http.StatusError`, a typed error that keeps the status code in the chain, and `platform/errs/http`, the classifier that reads it. Server-state codes (500, 502, 503, 504, other 5xx, 429, 408) are retryable dependency errors; verdicts on the request (4xx, 3xx, the permanently broken 501 and 505, and a code that was never a response) are not.

`*url.Error` covers transport failures, including an expired deadline — that means the remote end did not answer in time, so it is attributed to the dependency. Only `context.Canceled` is declined, so a shutdown stays out of a backend's dependency metrics and `platform/errs/generic` claims that node. Nothing in the classifier depends on `platform/errs/mysql` being wired to reach a verdict; a test pins the `generic` + `http` wiring a service with no MySQL dependency would use.

The classifier must still be listed before `platform/errs/mysql`, whose `net.Error` rule matches `*url.Error` and would otherwise claim HTTP transport failures. Documented in the errs README next to the wiring example.

`StatusError.Error` renders at most 1 KiB of the response body. The rendered string lands in the queue's dead-letter record via `Reject(ctx, err.Error())`, and `last_error` is a finite column, so an error page from a chatty gateway must not be able to fail that write and leave the message stuck rather than dead-lettered. `Body` itself is kept whole for callers that want to inspect it.

Convert both build-runner clients — Buildkite and GitHub Actions — to return the typed error, and wire the classifier into the stovepipe server, which runs both. Client tests now assert the code is reachable with `errors.As`, so a refactor back to `fmt.Errorf` cannot pass silently. The submitqueue-side clients (`changeprovider/github`, `phabricator/conduit`, `mergechecker/github`) are unchanged: no service wires `httperrs` on those paths, so nothing would read the code.

## Behavior change worth calling out: retried creates can duplicate a build

Creating a build is not idempotent, and a rejected create is now reported with its status code like any other, so a retried 502 can start a second build when the first was already accepted. Stovepipe's build controller has no guard against that — `request.State.IsTerminal()` is false at the point `Trigger` runs, and the `ErrAlreadyExists` tolerance keys on a build ID the retry has not minted yet.

This exposure is not new: a transport failure on the same call is already retryable today through the mysql classifier's `net.Error` rule, so a connection reset mid-write can already duplicate a build. This change widens it from transport failures to 5xx responses. The trade is deliberate — a request that gives up on a proxy blip strands its queue slot, which is worse — and both clients' create paths now document it. Making create idempotent is follow-up work and is the next branch.

## Why

- Correctly classify various types of HTTP errors as retryable

## Test Plan

- Deploy with corresponding changes internally
- Monitor DLQ volume in the message queue — retryable failures should stop landing there
- Watch build creation counts against request counts for the duplicate-build case above, which DLQ volume will not show

### DIFF
--- a/platform/errs/http/http.go
+++ b/platform/errs/http/http.go
@@ -1,0 +1,136 @@
+// Copyright (c) 2025 Uber Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package http provides an errs.Classifier for failures returned by HTTP
+// clients: a rejected status code (platform/http.StatusError) and a transport
+// failure (*url.Error, which is what http.Client.Do returns).
+//
+// Wire it into any service whose extensions call an HTTP API — build runners,
+// CI gateways, webhook senders. Without it every status code looks the same to
+// the pipeline: unclassified, and therefore non-retryable, so a 502 from a proxy
+// dead-letters the message on its first attempt rather than being retried.
+//
+// Order matters when wiring this alongside platform/errs/mysql. The MySQL
+// classifier treats any net.Error as retryable infra, and *url.Error satisfies
+// net.Error, so it will claim HTTP transport failures if it runs first. List
+// this classifier before it to keep those failures attributed to the dependency
+// they came from:
+//
+//	errs.NewClassifierProcessor(
+//	    genericerrs.Classifier,
+//	    httperrs.Classifier,
+//	    mysqlerrs.Classifier,
+//	)
+package http
+
+import (
+	"context"
+	nethttp "net/http"
+	"net/url"
+
+	"github.com/uber/submitqueue/platform/errs"
+	phttp "github.com/uber/submitqueue/platform/http"
+)
+
+// Classifier implements errs.Classifier for HTTP client failures. It recognises:
+//
+//   - *phttp.StatusError — dispatches on the status code. Codes that describe a
+//     server-side or overload condition (500, 502, 503, 504, other unassigned
+//     5xx, 429, 408) are retryable dependency errors. Codes that describe a
+//     verdict on the request itself (4xx, 3xx, and the permanently broken 501
+//     and 505) are non-retryable dependency errors.
+//   - *url.Error — the wrapper http.Client.Do puts around connection resets, DNS
+//     failures, TLS errors and timeouts. A retryable dependency error, except
+//     for our own context cancellation (see Classify).
+//
+// Everything else returns errs.Unknown so the classifier-processor walk can keep
+// looking down the unwrap chain.
+//
+// The classifier never returns errs.User. A 400 or 403 says the request was
+// rejected, not that a person did something wrong; only the controller knows
+// whether the request was built from user input. Controllers express that by
+// wrapping with errs.NewUserError, which short-circuits pass 1 of the
+// classifier-processor before this classifier is consulted.
+//
+// The classifier is stateless; this package-level singleton is the canonical
+// handle. Pass it as one of the variadic classifiers to
+// errs.NewClassifierProcessor; the resulting processor is what gets handed to
+// consumer.New.
+var Classifier errs.Classifier = classifier{}
+
+type classifier struct{}
+
+// Classify inspects a single node. Per the errs.Classifier contract, this must
+// not call errors.Is / errors.As — the classifier-processor owns the chain walk.
+func (classifier) Classify(err error) errs.Verdict {
+	if se, ok := err.(*phttp.StatusError); ok {
+		return classifyStatusCode(se.StatusCode)
+	}
+
+	if ue, ok := err.(*url.Error); ok {
+		// A cancelled or expired context means we gave up, not that the remote
+		// end misbehaved: the deadline belongs to this process. Report Unknown
+		// so the walk continues into the cause, where the generic classifier
+		// picks up context.Canceled as plain retryable infra. That keeps
+		// shutdown noise out of the dependency-error metrics for this backend.
+		// Reading the Err field is not a chain walk, and declining here leaves
+		// the framework's walk free to reach that node.
+		//
+		// This covers the shape http.Client.Do produces for context failures. A
+		// context error buried under further transport wrapping falls through to
+		// the retryable dependency verdict below, which retries the same way and
+		// differs only in the dependency tag.
+		if ue.Err == context.Canceled || ue.Err == context.DeadlineExceeded {
+			return errs.Unknown
+		}
+		// Anything else at this layer is a failure to complete an exchange with
+		// the remote end — reset connection, refused dial, DNS, TLS, client
+		// timeout. None of those say the request was invalid, so another attempt
+		// is worth making.
+		return errs.InfraDependencyRetryable
+	}
+
+	return errs.Unknown
+}
+
+// classifyStatusCode maps an HTTP status code to a Verdict.
+//
+// The split is whether the code describes the state of the server or a verdict
+// on the request. Server-state codes change on their own, so the same request
+// can succeed later. Request-verdict codes will not: replaying the request
+// reproduces the answer, and the message should dead-letter for a human or a
+// reconciler instead of retrying until it runs out of attempts.
+func classifyStatusCode(code int) errs.Verdict {
+	switch code {
+	case nethttp.StatusRequestTimeout, // 408 — the server stopped waiting; sending it again is reasonable.
+		nethttp.StatusTooManyRequests: // 429 — over a rate limit that resets with time.
+		return errs.InfraDependencyRetryable
+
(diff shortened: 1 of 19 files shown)


platform/extension/buildrunner/buildkite/client.go
```diff
@@ -122,7 +122,9 @@ func (c *Client) do(ctx context.Context, method, rawURL string, body []byte, out
 		return ErrNotFound
 	}
 	if status < 200 || status >= 300 {
-		return fmt.Errorf("API returned status %d: %s", status, respBody)
+		// Typed rather than formatted so platform/errs/http can read the code
+		// and tell a transient 502 from a permanent 400.
+		return phttp.NewStatusError(status, respBody)
```
behinddwalls, line 132: **5xx on a non-idempotent POST now retries — this can duplicate builds.**

`CreateBuild` (POST `/builds`) goes through this branch, so a 502 here is now `InfraDependencyRetryable` instead of dead-lettering. A proxy 502 frequently arrives *after* the upstream accepted the write, and the only redelivery guard in the build controller is `request.State.IsTerminal()` (`stovepipe/controller/build/build.go:114`), which is false at the point `Trigger` is called (`:134`).

So the retry path is: a second Buildkite build is created, and the first build's ID was never persisted — it runs to completion unobserved, with no `buildsignal` poller attached. The `Create(...)` + `errors.Is(err, storage.ErrAlreadyExists)` tolerance just below doesn't help, because the retry mints a *new* build ID and there's nothing to dedup against.

That may still be the right trade (a stranded request is arguably worse than a duplicate build), but it is a behavior change the description doesn't call out, and the test plan ("monitor DLQ volume") won't catch it — duplicate builds show up as CI spend, not DLQ volume. Worth either stating the decision explicitly, or scoping it: keep the typed error on the GET/cancel paths and leave `CreateBuild`'s rejection untyped (or wrap it non-retryable at the controller).

platform/errs/http/http.go
```diff
@@ -0,0 +1,136 @@
...
+		// picks up context.Canceled as plain retryable infra. That keeps
+		// shutdown noise out of the dependency-error metrics for this backend.
+		// Reading the Err field is not a chain walk, and declining here leaves
+		// the framework's walk free to reach that node.
+		//
+		// This covers the shape http.Client.Do produces for context failures. A
+		// context error buried under further transport wrapping falls through to
+		// the retryable dependency verdict below, which retries the same way and
+		// differs only in the dependency tag.
+		if ue.Err == context.Canceled || ue.Err == context.DeadlineExceeded {
```
behinddwalls: **The hand-off works for `Canceled`, but not for `DeadlineExceeded` — and it's load-bearing on `mysqlerrs` being wired.**

The comment says declining lets the generic classifier pick the node up as plain retryable infra. That's true for `context.Canceled`. It isn't for `context.DeadlineExceeded`: `platform/errs/generic` only matches `Canceled`, and its own test asserts `Unknown` for the deadline case (`platform/errs/generic/generic_test.go:40`).

What actually claims that node today is `mysqlerrs` — `context.DeadlineExceeded` satisfies `net.Error` (it has `Timeout()`/`Temporary()`; I confirmed this), so the mysql `net.Error` rule returns `InfraRetryable`. The deadline case is therefore resolved by exactly the cross-extension misattribution the ordering rationale in this file's package doc exists to prevent.

The sharper problem is the future service the README invites: wire `generic` + `http` without `mysql`, and `&url.Error{Err: context.DeadlineExceeded}` falls through every classifier → unclassified → non-retryable → dead-letters on the first attempt. That's the exact failure this PR is fixing.

Neither test catches it — `TestClassifier_TransportFailures` asserts `Unknown` at the node level (passes either way), and `TestClassifier_AppliedViaProcessor` only exercises `Canceled`.

Two ways out: add `context.DeadlineExceeded` to `genericerrs` next to `Canceled` (most consistent with its stated reasoning — a deadline we set is also "we gave up"), or drop the special case for it here and let it be `InfraDependencyRetryable`. Either way, a processor-level test with only `generic` + `http` wired would pin it.

(Separately, and correctly anticipated by the comment below: with `http.Client{Timeout:…}` the inner error is an internal `*httpError`, not `context.DeadlineExceeded`, so that shape falls through to the retryable dependency verdict. No change needed there.)

service/stovepipe/server/main.go
```diff
@@ -264,6 +265,7 @@ func run() error {
 	primaryConsumer := consumer.New(logger.Sugar(), scope.SubScope("consumer"), registry,
 		errs.NewClassifierProcessor(
 			genericerrs.Classifier,
+			httperrs.Classifier,
```
behinddwalls, line 268: **Same service, half the coverage: the GitHub Actions client is still untyped.**

Stovepipe — the one service being wired here — has both build runners (`stovepipe/extension/buildrunner/githubactions/`), and the platform GH Actions client still formats status codes into strings: `platform/extension/buildrunner/githubactions/client.go:149` and `:188`. Those are the same two call shapes just converted in the Buildkite client.

Leaving the submitqueue-side clients (`changeprovider/github`, `phabricator/conduit`, `mergechecker/github`) is defensible — neither the gateway nor the orchestrator wires `httperrs`, so nothing there would read the code. But within stovepipe the split means a 502 is retried or dead-lettered depending only on which runner the queue happens to use. Two lines to close it.

platform/extension/buildrunner/buildkite/client.go
```diff
@@ -105,7 +105,7 @@ func (c *Client) CancelBuild(ctx context.Context, number int) error {
 		// Already terminal — no-op per BuildRunner.Cancel contract.
 		return nil
 	default:
-		return fmt.Errorf("unexpected status %d from cancel: %s", status, respBody)
+		return fmt.Errorf("cancel build: %w", phttp.NewStatusError(status, respBody))
```
behinddwalls, line 113: **Nothing tests that the client actually returns the typed error.**

The whole chain hinges on this branch and `do` returning `*phttp.StatusError`, but the existing client tests only `require.Error` (`client_test.go:88,112,153,171,222`) — they passed before this change and pass after, so a future refactor back to `fmt.Errorf` would be silent.

A one-line `errors.As(err, &se)` plus a status-code assertion in the cancel-default case and the non-2xx case in `do` would pin the linchpin behavior.

platform/errs/http/http.go
```diff
@@ -0,0 +1,136 @@
...
+	// and 504, the shapes a proxy or overloaded backend produces, plus any
+	// unassigned or vendor-specific 5xx, which follow the same convention.
+	if code >= nethttp.StatusInternalServerError {
+		return errs.InfraDependencyRetryable
+	}
+
+	// 4xx other than the two above, 3xx the client was not configured to follow,
+	// and anything else a caller chose to reject: a verdict on this request. Not
+	// retryable.
+	return errs.InfraDependency
```
behinddwalls, line 124: Nit: this also catches codes below 400, including `0`. `SendRequest` only yields status `0` alongside a non-nil error, so no caller in-tree should construct that — but since the fall-through is "non-retryable, don't try again", an explicit note (or a separate branch for an unexpected/invalid code) would make it clear that's a deliberate default rather than an unconsidered case.

platform/http/status.go
```diff
@@ -0,0 +1,57 @@
...
+//
+//	if status < 200 || status >= 300 {
+//	    return phttp.NewStatusError(status, respBody)
+//	}
+type StatusError struct {
+	// StatusCode is the HTTP status code from the response.
+	StatusCode int
+	// Body is the response body as read from the wire, or empty when the
+	// caller had no body to attach.
+	Body string
```
behinddwalls, line 46: Nit, no regression: `Body` is unbounded, so a 5xx proxy HTML page ends up in `Error()` verbatim. That's identical to the string this replaces, but the error now also flows into the structured DLQ failure record — worth a glance at whether that field has a length limit, otherwise a chatty gateway could turn a retryable failure into a failed insert.

behinddwalls (commented):
Solid, well-documented change. The per-node `Classifier` contract is respected (direct type assertions, no `errors.As`), the ordering hazard vs. `mysqlerrs` is real and correctly documented — I confirmed `*url.Error` does satisfy `net.Error`, so `mysqlerrs` would claim it if listed first — and the tests cover both the node-level and processor-level shapes.

Four things worth resolving, inline, roughly in order of consequence:

1. **Duplicate builds** — 5xx on `CreateBuild` (a non-idempotent POST) is now retryable, and the build controller has no guard that would stop a retry from starting a second Buildkite build.
2. **`context.DeadlineExceeded`** — the deferral to "the generic classifier" doesn't land: generic returns `Unknown` for it, and only `mysqlerrs`' `net.Error` rule catches it. A service wiring `generic` + `http` without `mysql` dead-letters client timeouts.
3. **GH Actions client unconverted** — same service, both runners, only one gets the benefit.
4. **No test** that the Buildkite client returns `*StatusError`.

Plus two nits.

platform/extension/buildrunner/buildkite/client.go
mnoah1: resolved in client_test.go.

service/stovepipe/server/main.go
mnoah1: Updated

platform/extension/buildrunner/buildkite/client.go
mnoah1: Updated - implementers are responsible for specifying any non-retryable ranges and wrapping there.

behinddwalls (approved):

PR #567, opened 2026-08-11
[draft] "sampler" build runner to sample between multiple build runners

mnoah1:


behinddwalls:
we can probably close it in favor of shadow/simulation framework

PR #570, opened 2026-08-11
feat(client): a SubmitQueue client library, with list and watch

behinddwalls:
### Why?

Seeing what a queue was doing meant running the demo tool, which creates pull requests as a side effect. The live table it draws is the good part, and it was trapped inside a tool whose job is generating traffic. Meanwhile the gateway has exposed a paged `List` RPC that no client has ever called, and the CLI's `status` reads one request at a time by id — so there was no way to ask what a whole queue was doing without adding to it.

Underneath that sat a duplication problem heading somewhere worse. Three binaries dialled the gateway for themselves, `parseStrategy` existed twice verbatim, and the demo was growing a second copy of everything the CLI would eventually need. Extracting only the table would have treated the symptom.

### What?

`submitqueue/client` is now the client for the domain: dialling, the calls made against a gateway, and the terminal view of a queue. Both binaries become thin over it — the CLI is flag parsing, and the demo is GitHub scaffolding plus calls into the library. The demo's `main.go` drops from 1069 lines to 369 with no change in what it does, and its GitHub REST helpers move to their own file, since they are not SubmitQueue client code.

**`list` and `watch`.** `list` draws a queue's recent requests once, following continuation tokens so a caller gets the answer rather than a cursor. `watch` seeds its rows from the same listing, or from named ids, and then follows them with the tracker that already existed. Its set is fixed when it starts: a watch that grew as its queue did would never finish, and finishing is what makes it usable from a script — it exits non-zero if anything settles anywhere other than `landed`, the contract the demo used to own alone.

**Addressing.** `-addr` is unchanged and passed to the dialler untouched, so `dns:///host:port` and `unix:///path.sock` work alongside a plain `host:port`. Transport security is a separate `-tls` flag rather than part of the address, because gRPC keeps target resolution and credentials apart — there is no scheme meaning "use TLS", and inventing one would only mislead. The demo's odd `-gateway` flag is renamed to `-addr` to match the three real clients.

**The changes column.** It was the one part of the table tied to having created the pull requests. A row now carries cells of text and an optional URL, supplied by the caller: the demo passes pull request numbers linked to their pages, and a client watching a queue it did not create passes the change URIs the gateway reports. The hyperlink and width handling is shared, including that padding counts on-screen width — a hyperlink is mostly escape bytes occupying no columns.

**Credentials.** The client can present a bearer token. `TokenEnv` names the variable holding it rather than carrying the token, so a secret never reaches a command line, and an unset variable sends nothing rather than failing. Nothing in this repository checks it — the gateway admits every caller — so it is there for a gateway reached through something that does: a proxy, a sidecar, an ingress terminating auth ahead of the service. gRPC refuses per-RPC credentials on an insecure connection unless they declare they do not need transport security, which is why the credential declares it and `-tls` stays a separate choice.

## Test Plan

✅ `bazel test //...` — all 112 targets pass, including the Docker-backed integration and end-to-end suites.

✅ The extraction is behaviour-preserving by construction: all 23 view tests moved to `submitqueue/client` and pass unchanged there. None was lost — the demo had 28, and 23 plus the 5 file-layout tests that stayed accounts for all of them.

✅ `TestCredentialsReachTheServerOverPlaintext` runs a real gRPC server on a loopback port and asserts the token arrives as `Bearer …`. This is the case that silently breaks otherwise: gRPC refuses per-RPC credentials on an insecure connection unless they declare they do not need transport security, so a token that works over TLS can simply never be sent without it.

✅ `List` paging: pages are followed to the end, a limit cuts across pages without over-fetching, an empty page ends the walk so a server that never stops handing out tokens cannot spin, and `-since` becomes a receipt-time bound while its absence leaves the window open.

Not driven by hand: `list` and `watch` have not been pointed at a running gateway, so the end-to-end shape of the rendered table against real data is unverified. Their paging, settle and verdict logic is covered hermetically above.

## Also in this PR

- **feat(client): wrap the stage at the terminal's width, never truncate**
- **feat(demo): open independent pull requests in parallel**
- **fix(demo): let demo-queue inherit the fake build runner**

<sub>Each commit above carries its own rationale and test plan in its message.</sub>


submitqueue/client/query.go
```diff
@@ -0,0 +1,140 @@
...
+//
+// Paging is followed here rather than exposed, because a caller asking what a
+// queue is doing wants the answer, not a cursor. A caller that needs the cursor
+// can reach the generated client through Gateway.
+func (c *Client) List(ctx context.Context, q ListQuery) ([]*pb.RequestSummary, error) {
+	if q.Queue == "" {
+		return nil, fmt.Errorf("queue must not be empty")
+	}
+
+	req := &pb.ListRequest{Queue: q.Queue, PageSize: int32(q.PageSize)}
```
mnoah1: `ReceivedBeforeMs` is never set, but the gateway's list controller rejects any request where `received_at_or_after_ms >= received_before_ms` (`list.go:85`). With `Since` unset you send `(0, 0)`; with `-since 1h` you send `(now-1h, 0)` — both invalid, so every `list` and `watch` call fails with InvalidArgument against a real gateway. Every other caller in the repo sets both bounds. Set `req.ReceivedBeforeMs = time.Now().UnixMilli()` once before the loop and keep it fixed across pages, since the continuation token pins both bounds.

mnoah1 (approved):

submitqueue/client/query.go
behinddwalls: Good catch — fixed in `06080b1a`, and you were right on both halves.

Confirmed the failure: `list.go:85` rejects `received_at_or_after_ms >= received_before_ms`, so `(0, 0)` and `(now-1h, 0)` both fail. It also reaches further than `list` — the `watch` subcommand goes through `watchRows` → `List` (`gateway/client/main.go:313`) unless you name requests with `-sqid`, so default `watch` was broken too. `Tracker.Poll` itself only calls `GetRequestHistoryByID`, so the polling loop was fine; it was the initial row discovery that failed.

Both bounds now come from a single `now` taken before the first page, exactly as you suggested. Worth recording *why* fixing it up front is required rather than merely tidy: `list.go:111` validates that the continuation token's bounds still match the request, so recomputing either bound per page would be rejected from the second page on.

Added two tests: one asserts the upper bound lands at the call time and that `received_at_or_after_ms < received_before_ms` holds, and one asserts both bounds are byte-identical across three pages. Verified the first genuinely catches the bug — reverting the fix fails it with `"0" is not less than "0"`, the same symptom a real gateway would report.

PR #571, opened 2026-08-11
feat(periodic metrics): emit periodic metrics on latest green

mnoah1:
## Summary
- Create a Reporter interface that can be called to take a snapshot of metrics at a current point in time
- This can be used for things that need to be reported periodically, like the age of the latest green for a queue, and other future aspects of queue health
- Wire into the record step so we have one call site. We can also schedule this as a periodic task in our internal service

## Test Plan
- Add in our service, and query on the emitted metrics


stovepipe/extension/observability/lastgreen/lastgreen.go
```diff
@@ -0,0 +1,73 @@
...
+
+type reporter struct {
+	scope         tally.Scope
+	stores        storage.Factory
+	sourceControl sourcecontrol.Factory
+}
+
+var _ observability.Reporter = (*reporter)(nil)
+
+// New creates a Reporter for a queue's last-known-green age.
```
behinddwalls, line 22: can this be done in record step itself? i am just wondering if it deserves to be it's own extension or not?

mnoah1: I am thinking that we can also schedule it separately on a periodic schedule (similar to ingest).  Basically it can be called independently from wherever, and produces a real time snapshot of the metric.

mnoah1: Actually the current PR is a duplicate - please see https://github.com/uber/submitqueue/pull/572

mnoah1:
Duplicate

PR #572, opened 2026-08-11
feat(stovepipe): report last-green freshness

mnoah1:
## Summary
Support periodic observations of queue health via a Reporter interface.  This can be called as part of other tasks that we know will alter queue state (wired here into the record step), and also triggered separately on a periodic schedule.

SourceControl will provide one extra method to provide the timestamp of the latest green commit - this will give us a real time gauge of how much time has elapsed since the last commit was marked green.

Since this is a real time observation of current state, and not a collection of observations, a histogram does not make sense for this use case - updated docs to clarify when to use gauge vs. histogram.

## Test Plan
- Deploy with this change (and corresponding source control changes), add query for this change


stovepipe/extension/observability/observability.go
```diff
@@ -0,0 +1,54 @@
...
+// the current health of a queue. A Reporter answers "how is this queue doing
+// right now?" by sampling state the pipeline already persists and emitting it
+// as metrics, so operators can see queue health without querying the store.
+//
+// A Reporter is bound to a single queue at construction by its Factory, so its
+// methods take no queue argument. Reporting is deliberately best-effort: Report
+// returns nothing, because an observation that cannot be made must never change
+// what the pipeline records or decides. Implementations emit their own error
+// metrics instead, which keeps every call site a one-line defer.
+package observability
```
behinddwalls: i think the main question for this is, does it need to be a extension? i think we need think about it... and why do we think it cannot be done in controller itself

behinddwalls (changes requested):

stovepipe/extension/observability/observability.go
mnoah1: Ok - yeah thinking about it a bit more, it's more like it's own independent controller than an extension.  The main difference is, I think it should exist outside the existing workflow and fire periodically on its own schedule independently of all other controllers.

I am thinking the best fit could be to create a periodicmetrics controller, and drive it via its own topic.  The topic could be populated on a periodic basis by our internal service, so the controller fires regularly on its own schedule.

mnoah1: Updated

mnoah1:
Closing - aligned on polling for this internally.

PR #574, opened 2026-08-12
fix(messagequeue): intent-scoped message IDs so wake-ups stop vanishing

### LINKED ISSUE #352: messagequeue: entity-ID message IDs + publish dedup silently drop wake-up events; use intent-scoped message IDs
## Problem

The MySQL message queue dedups publishes on the unique key `(topic, partition_key, id)`, and every pipeline controller reuses the **entity ID** (batch ID, build ID, request ID) as the message ID. Because consumed rows are only deleted by GC (which runs on idle poll ticks), a re-publish for the same entity on the same topic is **silently swallowed** whenever an earlier row — even an already-consumed one — has not been GC'd yet. The publisher gets a success, no new row is inserted, and the consumer never sees the event.

This turns one-shot wake-up publishes into best-effort delivery, gated by GC timing.

## Mechanics

- `platform/extension/messagequeue/mysql/message_store.go` — `InsertDelayed` uses `INSERT ... ON DUPLICATE KEY UPDATE topic = topic`: a repeated publish for the same `(topic, partition_key, id)` is deliberately treated as an idempotent success ("lets callers safely retry publishes"). That is the right behavior for **retries of the same event**, but it also fires for **new events about the same entity**.
- Consumption is offset-based; acked rows stay in the table until GC.
- `platform/extension/messagequeue/mysql/subscriber.go` — GC only runs every Nth **idle** poll tick. A busy consumer defers GC, which widens the swallow window exactly when traffic is high.

## Where it bites

Every controller publishes with `entityqueue.NewMessage(<entityID>, ...)`:

- `score` → speculate (msg ID = batch ID)
- `buildsignal` → speculate (msg ID = batch ID) and its own poll self-reschedule (msg ID = build ID)
- `mergesignal` → conclude + speculate fan-out (msg ID = batch ID)
- `cancel` → speculate nudge (msg ID = batch ID)
- `speculate` → dependent wake-up fan-out (msg ID = dependent batch ID), build/merge/conclude publishes (msg ID = batch ID)

Concrete loss scenario: batch B's speculate message from the score stage is consumed; the row awaits GC. Dependency A then merges and the terminal fan-out publishes "speculate B" to wake it — same key `(speculate, queue, B)` → silent no-op → B is never woken and stalls until some unrelated event pings it. Poll loops (buildsignal) survive because they re-publish every few seconds until terminal, but one-shot nudges (mergesignal fan-out, cancel nudge, dependent wake-up, DLQ re-trigger tooling) have no retry.

We have hit this before operationally: re-trigger publishes had to mint distinct message IDs to land at all.

## Proposal: intent-scoped message IDs

The message ID should identify the **intent to process an event occurrence**, not the entity it is about. The entity reference already travels in the payload; consumers correlate via payload, not message ID, so this is a publisher-side convention change.

Sketch:

- Wake B because A reached a terminal state → `batch/B/wake/batch/A`
- Advance B after its build finished → `batch/B/buildsignal/build/X`
- Poll tick N for build X → `build/X/poll/N`
- Stage handoff (score → speculate) → `batch/B/scored`

Properties: a retry of the **same** intent still dedups (idempotent publish is preserved — same cause, same ID), while a **new** intent about the same entity can never collide with an older row. IDs stay deterministic (derived from cause), so redelivery/crash-retry paths remain idempotent without random UUIDs.

## Suggested scope

1. Define the ID convention in `platform/base/messagequeue` docs (and decide whether `NewMessage` should take an explicit intent/cause argument to make the convention hard to miss).
2. Audit every `NewMessage` call site across `submitqueue/`, `runway/`, and tooling; migrate to intent-scoped IDs.
3. Check consumers/DLQ/log correlation for any place that assumes message ID == entity ID.
4. Consider a lint or contract test that rejects bare entity IDs as message IDs on pipeline topics.

## Notes

- Pre-existing and systemic; not introduced by the speculation-tree work — that review just made it visible (the new dependent wake-up fan-out inherits the same window as mergesignal/cancel).
- Orthogonal mitigation (GC cadence / delete-on-ack) would shrink the window but not close it; the ID convention closes it.

behinddwalls:
### Why?

The MySQL queue deduplicates publishes on `(topic, partition_key, id)` via `INSERT ... ON DUPLICATE KEY UPDATE topic = topic`, and rows are removed only by `GarbageCollect` — which runs from `subscriber.go` on idle ticks only, with `gcCounter` reset to `0` by any tick that delivered a message. On a busy partition GC never runs, so the dedup horizon is unbounded exactly when traffic is high. A publish that collides is reported as a success, writes nothing, and has no error to retry and no row to deliver.

Controllers reusing a bare entity ID as the message ID therefore lose their *second, unrelated* publish about that entity. Concretely: `batch` announces a new batch to speculate under the batch ID; when that batch later merges, `mergesignal.fanout`'s "wake the dependents" publish reuses the same ID and is dropped against the announcement. `fanout` returns nil and the delivery is acked. Other speculate publishers already mint distinct IDs, so another batch's build signal usually re-plans the queue and hides this — the stall shows at the tail, when the merged batch is the last in flight and nothing else pings speculate.

Fixing only that call site would leave the shape in place. `submitqueue/core/publish` documented the hazard but was domain-scoped, so `runway/` and `stovepipe/` published bare entity IDs with no shared guidance, and `platform/base/messagequeue` documented none of it.

### What?

`submitqueue/core/publish` moves to `platform/publish` — it already imported only `platform/base/messagequeue` and `platform/consumer`, so this is a relocation, and it lets every domain share one helper instead of hand-rolling the resolve-registry-and-publish block three more times.

The new `publish.IntentID(entityID, cause...)` names the occasion to publish rather than the entity published about. A retry of the same cause dedups, which is what keeps redelivery safe; a new cause about the same entity can never be swallowed.

Deterministic IDs go where a duplicate is harmful and the cause is nameable:

| Publish | ID |
| --- | --- |
| merge result → speculate | `{batch}/merged` (**the bug**) |
| merge result → conclude | `{batch}/conclude/merged` |
| build poll → speculate | `{batch}/build-signal/{build}/{status}` |
| speculate → merge dispatch | `{batch}/merge-dispatch` |
| speculate → conclude on terminal | `{batch}/conclude/speculate` |
| one-shot hand-offs | bare `IntentID(id)` |

`buildsignal` publishes to speculate on *every* poll, so keying on `{batch}/build-signal/{build}` would have deduplicated the terminal wake-up into the first poll's. Including the observed status lands every transition and collapses only the polls that saw nothing new — a reduction in speculate churn versus the previous `UniqueID`.

`publish.UniqueID` is kept, and deliberately left in place, for repeat-until-effective nudges whose provoking condition is that nothing recorded the last one — speculate's build dispatch, its self-heal fan-out and `recoverable`, cancel's nudge, the DLQ re-trigger. Those have no stable cause to name, and a deterministic ID would dedup the re-send against the message that went missing. Each now says so at the call site.

`runway/controller/dlq` gets its own cause: it answered on the same topic under the same correlation ID as the live handler, so a dead-lettered request could have its terminal failure deduplicated against an answer already sent.

`tool/linter/messageid` makes the helper the only door. Judging whether an ID expression is well chosen is not decidable by reading it, so the linter enforces the structural rule instead: `NewMessage` may only be called from `platform/publish` and the queue backends. It found seven production call sites the manual audit missed.

## Test Plan
- ✅ `bazel test //...` — 99 unit tests
- ✅ `bazel test //test/integration/...` — 8 suites
- ✅ `bazel test //test/e2e/...` — 3 suites
- ✅ `make lint-license`, `make lint-message-id`, `make lint-queue-shard`; `make fmt` idempotent; `make tidy` clean
- New `TestProcess_FanoutDoesNotCollideWithTheBatchAnnouncement` was confirmed to fail against the old code — reverting the one expression reproduces the drop.
- New `TestDedupOutlivesConsumption` pins the underlying behaviour against real MySQL: a consumed and acked message still deduplicates a later publish under the same ID, and naming the cause gets it through.
- Not covered: no e2e exercises a dependent chain, so the merge→dependent-wake path is not verified end to end. Building that fixture is follow-up work.

Out of scope, deliberately: GC never running on a busy partition. It widens this window and also lets `queue_messages` grow without bound, but it is orthogonal to the ID convention and wants its own review.

## Issue

Closes #352

## Stack
1. @ #574
1. #576


submitqueue/orchestrator/controller/speculate/speculate.go
```diff
@@ -182,35 +182,44 @@ func (c *Controller) admit(ctx context.Context, store storage.Storage, batch ent
...
+// swallow the repeat that finally lands.
 //
 // queue and partitionKey are separate because they answer different questions.
 // queue is what the payload asserts the batch belongs to, and the consumer
 // resolves its storage from it, so it must always be the real queue. The
 // partition key only decides what stays serialized behind what: most publishes
 // want the queue, so a queue's batches are processed in order, but the build
 // dispatch partitions by batch so heads dispatch in parallel.
-func (c *Controller) publishBatchID(ctx context.Context, key consumer.TopicKey, batchID, queue, partitionKey string) error {
+func (c *Controller) publishBatchID(ctx context.Context, key consumer.TopicKey, msgID, batchID, queue, partitionKey string) error {
```
mnoah1, line 217: 4 strings in a row for these args - we may want to consider adding types for more of these as strings can easily get swapped.

mnoah1 (approved):

submitqueue/orchestrator/controller/speculate/speculate.go
behinddwalls: ACK will take care of them in later PRs

PR #580, opened 2026-08-12
feat(stovepipe): Promote green commits to a ref

prathshenoy:
## Summary
**What**:
- When a validation establishes a commit as green, the record stage now asks source control to advance the queue's promotion ref to that commit, so downstream systems can fetch the latest known-good commit by branch name. Which ref a queue promotes to, and whether it promotes at all, is source-control configuration resolved from the queue name.
- Promotion happens only while the request still holds the queue's last-green bookmark, and a commit that a rewritten history dropped from the branch is counted and skipped rather than retried.

**Why**:
- Consumers that only need "where is this branch last green" had to query Stovepipe and understand requests, commit locators and greenness degrees; resolving a branch name lets a consumer get the same answer with a plain fetch.
- Reusing the bookmark's existing ordering guard keeps the ref forward-only without new state, and skipping commits that left the branch avoids dead-lettering work whose greenness is already durable.

## Test Plan
- [x] Add unit tests.

## Revert Plan
- Revert this PR. The only residue is the ref sitting at whatever commit it last landed on.

## Issues
- [CODEM-435](https://linear.app/uber/issue/CODEM-435/promote-green-commits-to-a-ref)


behinddwalls (approved):

mnoah1 (approved):
Looks good - before merging, go ahead and implement the Promote endpoint on our internal source control extension, so the version bump job succeeds.

PR #587, opened 2026-08-13
build(lint): fail on tracked binaries, and build the gateway binaries into bin/

behinddwalls:
### Why?

An 18MB `client` binary reached #547 and was caught only by review — nothing in CI guards against a committed binary.

`go build ./service/submitqueue/gateway/client` names its output after the package directory and writes it to the *current* directory, so run from the repo root it drops a `client` there, where a broad `git add` sweeps it up. Every `service/*/server` package produces a root `server` the same way.

The obvious fix — ignoring `/client` — is the wrong one. It would stop the file being committed but equally stop git mentioning it at all, so the mistake becomes invisible and the artifact quietly accumulates. It is also name-specific: it covers 8 of the 12 `main` packages and none of the other binaries anyone might add. Worse, an ignored path is never tracked, so it would hide the offender from the check below rather than complement it.

### What?

`tool/linter/binaryfile` fails when any tracked file is binary. It follows the existing `licenseheader` / `messageid` / `queueshard` pattern and is wired in as `make lint-binary`, so `make lint` and the CI gate already cover it. Detection is git's own heuristic — a NUL byte in the leading 8000 bytes. The repository has no tracked binaries today, so the check starts clean and the allowlist is empty.

`make build-submitqueue-gateway-client` and `make build-submitqueue-gateway-server` build with Bazel for the host platform into `bin/client` and `bin/server`, giving a sanctioned way to produce a standalone binary. `bin/` was already ignored but had no producer — the only reference to it anywhere in the Makefile, `tool/`, or `.github/` was `rm -rf bin/` in `clean`.

No `.gitignore` change is needed: `bazel-bin/`, `bin/`, and `.docker-bin/` already cover every Bazel output, and the whole point of the linter is to surface a stray binary rather than hide it.

## Test Plan

✅ `bazel test //tool/linter/binaryfile:go_default_test` passes.

✅ `make lint-binary` on a clean tree reports `All 884 tracked files are text.`

✅ Reproduced #547 end to end: `go build ./service/submitqueue/gateway/client` from the repo root drops a 17,539,170-byte `client`; `git status` shows it as `?? client` rather than hiding it; after `git add client` the linter fails, naming the file and its size and pointing at Bazel.

✅ Both build targets produce Mach-O arm64 executables at `bin/client` and `bin/server`, are idempotent over Bazel's read-only output, and leave `git status` clean.

✅ `make gazelle` generates the linter's BUILD.bazel with no further drift, and `make help` lists all three new targets.


kevinlnew (approved):
Nice, thanks

PR #589, opened 2026-08-14
feat(conflict): Init Tango-backed target-overlap conflict analyzer

manjari25:
## Summary

- Adds a `conflict.Analyzer` implementation that detects conflicts between
  batches based on build-target overlap, resolved through a `TargetResolver`
  interface whose production implementation will call Tango's
  `GetChangedTargets` in a follow-up PR.
- The analyzer is the same shape as `pathoverlap` — takes identity-level
  batches, resolves targets internally, compares sets locally — but operates
  at build-target granularity instead of file paths.
- YARPC adapter and profile wiring follow in subsequent PRs.

## Test plan

- [x] Table-driven tests covering overlap, no overlap, ordering, empty candidate targets
- [x] Empty in-flight short-circuits with no conflicts
- [x] Resolver error propagation
- [x] `make test`, `make lint`, `make fmt`, `make check-gazelle` pass


submitqueue/extension/conflict/tango/tango.go
```diff
@@ -0,0 +1,112 @@
...
+	"fmt"
+
+	"github.com/uber/submitqueue/submitqueue/entity"
+	"github.com/uber/submitqueue/submitqueue/extension/conflict"
+)
+
+// TargetResolver resolves the set of build targets a batch affects. The
+// production implementation translates the batch's changes into a Tango
+// GetChangedTargets call; tests supply a fake.
+type TargetResolver interface {
```
behinddwalls, line 48: should this be an internal interface?

behinddwalls: Discussed offline,

we can restructure a bit to do something like this,

```
/submitqueue/extension/dependency/
                                 /conflict/conflict.go
                                 /conflict/all/*.go
                                 /conflict/none/*.go
                                 /conflict/fake/*.go
                                 /resolver/resolver.go
                                 /resolver/<impl>/<impl>.go
```

manjari25: Set up the tango resolver and analyzer impl in this PR. Will move over the existing ones in the next PR.

submitqueue/extension/dependency/conflict/tango/tango.go
```diff
@@ -0,0 +1,106 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package tango provides a conflict.Analyzer that reports a conflict between
+// two batches when their changed build targets overlap. The targets a batch
+// affects are resolved through an injected resolver.TargetResolver, whose
+// production implementation calls the Tango service.
+package tango
```
behinddwalls, line 21: do we want to call tango? i think it's a generic targetanalyzer based?

manjari25: Good point. Making the change. Will call it targetanalyzer.

submitqueue/extension/dependency/resolver/targetresolver.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+// Target is a build target a batch affects.
+type Target struct {
+	// Name identifies the target (e.g. "//service/foo:lib").
+	Name string
+	// Attributes carries backend-specific metadata the analyzer does not
+	// interpret today. Future consumers (e.g. conflict relaxation) can read
+	// keys like "distance" or "rule_type" without an interface change.
+	Attributes map[string]string
```
sbalabanov: It might be inefficient for memory / GC but tempting to use for adhoc extensions.
Let's say we always have "distance" property for each target.
This means that each target structure will have a map and each map will have the same key "distance" (8 bytes * number_of_targets right away) and each distance value is a string to be converted to int.

Do we really need to abstract out target-based conflict analysis?

The option is to have a conflict analyzer that is specific to Tango API. If another build target based analyzer needs to be implemented, it will just become a separate extension with its own domain model.

sbalabanov (changes requested):

submitqueue/extension/dependency/resolver/targetresolver.go
manjari25: >  to have a conflict analyzer that is specific to Tango API

That's what I originally had - I made the changes to reflect this. I think we still need a lightweight interface that wraps the tango RPC because generated Go types aren't interchangeable due to different import paths.

sbalabanov: Interface is fine but it likely wants to be specific and type-safe.
@behinddwalls objections?

behinddwalls: Yes..we want to have specific types defined for this which are specific to tango. Plus, I am unsure how we enable conflict relaxation on top if we want to use any target related information. So another question I have on is the conflict interface as well.. wether that's right abstraction or not. Or should SQ define how to accept the target information and then run conflicts separately and be aware of target information so that we can store and use across different phases

PR #590, opened 2026-08-14
fix(batch): split dependency analysis out of batch creation

behinddwalls:
### Why?

A batch that failed between `Transition(→Created)` and its speculate announcement was orphaned in `created` forever. `DependencyBatchStates()` includes `Created`, so every batch created afterwards took the orphan as a dependency, and a batch leaves `created` only when a speculate message names it. Nothing swept for unannounced batches, so the queue wedged behind it with no self-recovery. Observed on a 100-PR run: 65 landed, then zero for 50+ minutes at 50–150 builds/minute, with all 35 remaining requests stuck behind one batch.

`created` was carrying two facts at once, and dependency eligibility needs both — that the reverse indexes are complete, and that the batch is certain to be admitted. It became eligible on the first and only reliably advanced after the second.

Dropping `Created` from `DependencyBatchStates()` is not the fix: the announce→admit window is real (~121 ms healthy, past 60 s under load), and a conflicting batch created inside it must still serialize behind.

The reason a retry could not repair it is that the batch stage's message names a *request*, and a request says nothing about which batch a previous attempt made. Every retry minted a new batch ID and abandoned the old one — 175 batches for 170 requests on that run.

### What?

Batch creation now stops at `creating`: mint an ID, write the batch, hand the batch ID to a new dependency-analysis stage. That stage owns everything that makes a batch real — the enrolment decision, the conflict analysis, the reverse indexes, the association, the request claim, the promotion to `created`, and the announcement.

Because every message after creation names the batch, a retry resumes that same batch. Each remaining failure leaves it either in `creating` — in neither `ActiveBatchStates()` nor `DependencyBatchStates()`, so unreachable — or in `created` with a retry pending.

**The new topic must stay partitioned by queue.** Analysis reads the queue's dependency-eligible batches, so two batches of one queue analyzed concurrently would each see the other in `creating`, neither would serialize behind the other, and both would speculate as though the other did not exist. Serial consumption is also what makes the enrolment check safe.

The request claim carries two guards and both are load-bearing. It re-reads the request immediately before the compare-and-swap, so the version it writes against is fresh — which means a version guard alone would compare equal and write `batched` straight over a request cancel had already finished cancelling. The state check rejects a cancellation that completed before the read; the version guard rejects one landing in the window that remains between the read and the write.

Three supporting changes:

`speculate`'s run admits any batch it finds still in `created`. `read()` already lists `ActiveBatchStates()`, which includes `Created`, so every run already holds every stalled batch at no extra cost. Admission now depends on the batch's state rather than on one message surviving; repairs are best-effort and never fail the run.

The request-scoped DLQ reconciler no longer fails a request whose outcome a batch already owns. Since the batch stage mints one batch per delivery, a lost ack leaves a redundant attempt; if that attempt dead-letters, `TerminateRequest` would happily overwrite `batched` with `error` while the batch that actually enrolled the request is still speculating. Ownership is not the batch state alone: the association is written before the claim, so it outlives an abandoned attempt too, and a batch left in `creating` by a lost claim is inert — nothing promotes it and no conclude ever runs for it, so treating it as an owner would suppress the reconcile forever and strand the request short of terminal. A `creating` batch therefore owns the request only when the request itself reads `batched`, which only the claim writes.

`RequestStatusBatching` reports that a batch is being built, published once the batch row exists. `batched` now means the dependency set is resolved. This also fills a gap in the vocabulary: every other stage already pairs `-ing` with `-ed`.

`corebatch.FindByRequestID` is extracted from `cancel.findBatches` and shared, since three call sites now need the same association walk.

AGENTS.md gains a naming rule — identifiers state what they act on — and its comment rules are rewritten from qualitative advice into a closed category list with size budgets, after measurement showed the guidance was not moving the numbers.

## Test Plan
✅ `bazel test //submitqueue/... //platform/... //service/...` — 76 pass
✅ `bazel test //test/integration/...` — 8 pass
✅ `bazel test //test/e2e/submitqueue:all --test_output=streamed` — passes

All runs used `--nocache_test_results`.

Unit coverage spans the new stage's analysis, conflict dedupe, redelivery after transition, the mid-index-write retry, halted requests and batches, an already-enrolled duplicate, a claim lost to cancel, a cancellation that completed mid-analysis, and both halves of the DLQ ownership check — a request owned by a live batch, and a `creating` batch whose claim never landed. The two race cases were each confirmed to fail against the unguarded code before the guard was restored.

Two e2e cases were added and verified against orchestrator logs rather than assertions alone: a redelivered batch message enrols the request exactly once, and a batch stranded in `created` is admitted by a later run — the batch was observed being admitted three times, once on creation and again after being stranded.

## Issue

Closes CODEM-444


submitqueue/orchestrator/controller/dependencyanalysis/dependencyanalysis.go
```diff
@@ -0,0 +1,468 @@
...
+		if err != nil {
+			return err
+		}
+		if err := c.writeDependentIndexes(ctx, store, batch, dependencies); err != nil {
+			return err
+		}
+		if err := c.associateRequestsWithBatch(ctx, store, batch); err != nil {
+			return err
+		}
+		claimed, err := c.claimRequestsForBatch(ctx, store, batch)
```
mnoah1, line 191: In between this line (where we mark it batched), and line 205 where the batch gets promoted to `Created`, Cancellation could miss cancelling the batch and lead to it entering speculation:
```
Dependency analysis: request Validated → Batched
Cancel:              request Batched → Cancelling
Cancel:              sees only a Creating batch (not cancellable)
Cancel:              request Cancelling → Cancelled
Dependency analysis: batch Creating → Created → speculate/merge
```

May want to close this gap as a follow-up.

submitqueue/orchestrator/pipeline.go
```diff
@@ -139,12 +140,23 @@ var Stages = []pipeline.Stage[Deps]{
...
 	},
+	{
+		Key:           topickey.TopicKeyDependencyAnalysis,
+		Name:          "dependency-analysis",
+		ConsumerGroup: "orchestrator",
+		New: func(d Deps, sc pipeline.StageContext) (consumer.Controller, error) {
+			return dependencyanalysis.NewController(d.Logger, d.Scope, d.Storage, d.Analyzer, sc.Registry, sc.TopicKey, sc.ConsumerGroup), nil
+		},
+		DLQ: func(d Deps, sc pipeline.StageContext) (consumer.Controller, error) {
+			return dlq.NewDLQBatchController(d.Logger, d.Scope, d.Storage, sc.Registry, sc.TopicKey, sc.ConsumerGroup), nil
```
mnoah1, line 157: If a duplicate batch gets created here (https://github.com/uber/submitqueue/blob/c34c87fbac6da2b3d9f19241e4f4b035a5f1f132/submitqueue/orchestrator/controller/batch/batch.go#L184), and the extra Created row makes it to the DLQ controller, current logic in the DLQ controller would fail all requests even if they are in another successful batch.  May want to add some extra guardrails in the DLQ controller around this case.

mnoah1 (approved):

submitqueue/orchestrator/controller/dependencyanalysis/dependencyanalysis.go
behinddwalls: Agreed — tracked as [CODEM-452](https://linear.app/uber/issue/CODEM-452/cancel-misses-a-batch-mid-promotion-letting-a-cancelled-request-reach).

Confirmed the interleaving you sketched: `BatchStateCreating` is in `nonCancellableBatchStates`, so none of the three cases in cancel's batch loop match, `foundApplicableBatch` stays false, and cancel falls through to `cancelRequest`. The claim's guards only reject a cancel landing before or during the claim, and `firstHaltedRequestID` runs earlier still, so neither covers the post-claim window.

submitqueue/orchestrator/pipeline.go
behinddwalls: Good catch — tracked as [CODEM-453](https://linear.app/uber/issue/CODEM-453/batch-scoped-dlq-reconciler-fails-requests-that-another-live-batch).

`failBatch` fans out over `batch.Contains` with no ownership check, and `TerminateRequest` only declines when the request is already terminal — `batched` is not — so the redundant attempt's dead letter does overwrite a live batch's claim. The fix is to mirror `owningBatch` from the request-scoped reconciler in the inverse direction, excluding the dead-lettered batch itself so the guard can't suppress its own fan-out forever.

Issue #209, opened 2026-06-05
Consider a generic change-URI metadata/augmentation language (Change API)


behinddwalls:
## Context

In #197, the extension fakes inject failures via an ad-hoc `sq-fake=<token>`
marker embedded in a change URI, parsed by the **test-only** `core/fakemarker`
package (`Token([]string)` / `TokenInChanges([]entity.Change)`).

Reviewing that PR, @sbalabanov raised two related design questions:

- Should we build a **generic augmentation / metadata language** into change
  URIs, usable by these fakes *and similar* tooling — rather than a
  fake-specific marker convention?
- The marker lookup (`TokenInChanges`) arguably **belongs on the Change API**
  (`entity.Change`) rather than living in a test-only helper.

This issue captures that idea as a follow-up so #197 can land as-is (the marker
stays test-only there).

## Design considerations

A generic change-URI metadata facility would need to reconcile with the
**strict** provider change-ID parsers, which today reject anything that isn't a
canonical resource identifier:

- `submitqueue/entity/github/change_id.go` — requires a full 40-char lowercase
  hex head SHA; a `?key=value` query string would not parse.
- `submitqueue/entity/phabricator/change_id.go` — fixed `phab://` form.
- `submitqueue/extension/changeprovider/github/validate.go` — rejects
  unexpected schemes.

Open questions:

1. **Scope** — is URI-attached metadata a *production* capability, or strictly a
   test/example affordance? Putting parsing on `entity.Change` couples the
   production type to the convention; keeping it in `core/fakemarker` keeps the
   separation the fakes deliberately maintain ("never production").
2. **Shape** — RFC-3986 query params (`?k=v`) vs. fragment vs. a dedicated
   sidecar field on `Change`. Query params collide with the strict parsers
   above; a typed field avoids string-encoding entirely.
3. **Generality** — a neutral `key → value` accessor (e.g.
   `Change.URIParam(key)`) that `fakemarker` layers `sq-fake` semantics on top
   of, vs. a richer metadata model.

## References

- PR #197 — fake implementations with error injection
- Threads: the `core/fakemarker.go` discussion in #197.

Filed as a follow-up per the #197 review discussion. Not blocking.

PR #596, opened 2026-08-14
doc(simulator): add simulator RFC

roychying:
add simulator RFC

### DIFF
--- a/doc/rfc/submitqueue/simulator.md
+++ b/doc/rfc/submitqueue/simulator.md
@@ -1,0 +1,258 @@
+# Simulator
+
+Design notes for a simulator that verifies and benchmarks changes to orchestrator logic before they ship: per-extension record and replay, controller replay, and whole-pipeline simulation. This document captures **design decisions and rationale only**; the code lands after this RFC is reviewed.
+
+## Problem
+
+Changing orchestrator logic today leaves two options: trust a unit test against synthetic fixtures, or ship the change and watch production. Neither answers the question that matters. Does this behave differently from what is running now, and is the difference an improvement?
+
+A simulator closes that gap. It replays real or synthetic traffic through a candidate implementation, compares the result against the incumbent, and reports where and by how much the two diverge.
+
+The requirement is narrower than "simulate the whole queue." Every stage — validation, conflict detection, scoring, speculation, building, merging — must be independently swappable and comparable, and so must the seams inside a stage. Speculation is the case that makes this concrete: [speculation.md](speculation.md) describes two pluggable components, a Generator that enumerates candidate paths and an Allocator that funds them, and a change to one is not a change to the other. A harness that could substitute only the whole `Speculator` could evaluate neither. That constraint shapes the design more than the end-to-end case does.
+
+Two uses pull on it. They need different machinery and carry very different risk, so they are worth naming separately:
+
+- **Comparison** asks whether a candidate produces different output from the incumbent on the same input. It needs a corpus and a diff, and almost no modeling assumptions.
+- **Benchmarking** asks whether that difference improves the system. It needs a running pipeline, a model of build outcomes, and a defensible objective function. Every one of those introduces error.
+
+## Scope
+
+The simulator models SubmitQueue and Runway, which together form a closed pre-merge loop. Runway is stateless — no request store, no database, idempotency derived from the VCS — so the harness substitutes storage for one domain only.
+
+Stovepipe is out of scope by design rather than by convenience. It is a post-merge trunk-health poller with no coupling to SubmitQueue in either direction, and its only outputs are hooks, whose contract states that hook outcomes never write pipeline state. Nothing Stovepipe learns can reach the queue. Escaped conflicts, where changes pass individually and break together, surface inside SubmitQueue's own build stage, so measuring them needs no post-merge modeling.
+
+One hazard does live in the gap between the two. Runway performs transforming merges: rebase and squash-rebase rewrite commits, so the tree that lands is not always the tree that was validated. A break introduced by the transform itself is invisible to pre-merge validation, and therefore to any simulation of it. That is a limit on what the safety metric can claim, not a component to model.
+
+Concurrency-correctness work is also out of scope — invariant assertions, fault injection, and controlled interleaving exercise the delivery contract rather than orchestrator policy, need no corpus and no build oracle, and belong with end-to-end testing.
+
+## The unit of substitution
+
+Swappability is defined at the finest seam a queue can wire independently, not at the granularity of a pipeline stage.
+
+| Stage | Substitutable seams |
+|---|---|
+| validate | `Validator`; `MergeChecker`, designed but not yet wired |
+| batch | `conflict.Analyzer`, and the path projection it is constructed with |
+| speculate | `Generator` and `Allocator`, each on its own and jointly as the `Speculator` composing them; the `Scorer` the Generator is built with |
+| build | `BuildRunner` |
+| merge | `Merger`, which is Runway's rather than SubmitQueue's, and the default strategy it is constructed with |
+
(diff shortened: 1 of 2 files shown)


doc/rfc/submitqueue/simulator.md
```diff
@@ -0,0 +1,258 @@
...
+
+- **Comparison** asks whether a candidate produces different output from the incumbent on the same input. It needs a corpus and a diff, and almost no modeling assumptions.
+- **Benchmarking** asks whether that difference improves the system. It needs a running pipeline, a model of build outcomes, and a defensible objective function. Every one of those introduces error.
+
+## Scope
+
+The simulator models SubmitQueue and Runway, which together form a closed pre-merge loop. Runway is stateless — no request store, no database, idempotency derived from the VCS — so the harness substitutes storage for one domain only.
+
+Stovepipe is out of scope by design rather than by convenience. It is a post-merge trunk-health poller with no coupling to SubmitQueue in either direction, and its only outputs are hooks, whose contract states that hook outcomes never write pipeline state. Nothing Stovepipe learns can reach the queue. Escaped conflicts, where changes pass individually and break together, surface inside SubmitQueue's own build stage, so measuring them needs no post-merge modeling.
```
behinddwalls, line 22: > Stovepipe is out of scope by design rather than by convenience.

i think we also have need eventually at some point to say if some relaxation algorithm make main broken more than the others? would we want stovepipe to be in scope

doc/rfc/submitqueue/simulator.md
```diff
@@ -0,0 +1,258 @@
...
+
+One hazard does live in the gap between the two. Runway performs transforming merges: rebase and squash-rebase rewrite commits, so the tree that lands is not always the tree that was validated. A break introduced by the transform itself is invisible to pre-merge validation, and therefore to any simulation of it. That is a limit on what the safety metric can claim, not a component to model.
```
behinddwalls, line 24: > Runway performs transforming merges: rebase and squash-rebase rewrite commits, so the tree that lands is not always the tree that was validated. A break introduced by the transform itself is invisible to pre-merge validation, and therefore to any simulation of it. That is a limit on what the safety metric can claim, not a component to model.

can we simply the wording here? i think i get the gist for needs to be direct

doc/rfc/submitqueue/simulator.md
```diff
@@ -0,0 +1,258 @@
...
+
+Concurrency-correctness work is also out of scope — invariant assertions, fault injection, and controlled interleaving exercise the delivery contract rather than orchestrator policy, need no corpus and no build oracle, and belong with end-to-end testing.
```
behinddwalls, line 26: > invariant assertions, fault injection, and controlled interleaving exercise the delivery contract rather than orchestrator policy, need no corpus and no build oracle, and belong with end-to-end testing.

explain? we may want to be able to inject faults and test things, maybe a separate thing to think about

doc/rfc/submitqueue/simulator.md
```diff
@@ -0,0 +1,258 @@
...
+|---|---|
+| validate | `Validator`; `MergeChecker`, designed but not yet wired |
+| batch | `conflict.Analyzer`, and the path projection it is constructed with |
+| speculate | `Generator` and `Allocator`, each on its own and jointly as the `Speculator` composing them; the `Scorer` the Generator is built with |
+| build | `BuildRunner` |
+| merge | `Merger`, which is Runway's rather than SubmitQueue's, and the default strategy it is constructed with |
+
+A variant may differ in as many seams as the question needs. Pinning all but one is what makes a difference *attributable*: comparing two Generators under the same Allocator and Scorer is the only way to read the ranking-quality and starvation [diagnostics](#diagnostics) apart. The seam list belongs to the extension contract rather than to this document, and grows with it.
+
+A seam must be substitutable on its own. Nothing can supply a `Generator`'s `Iterator` without supplying the Generator, so the Generator is the unit; construction parameters such as the path projection are knobs rather than seams.
```
behinddwalls, line 42: > Nothing can supply a Generator’s Iterator without supplying the Generator, so the Generator is the unit; construction parameters such as the path projection are knobs rather than seams.

what if i want to try between 2 scorers or regression models under the hood? should that be a full generator wire-up or just scorer/model

doc/rfc/submitqueue/simulator.md
```diff
@@ -0,0 +1,258 @@
...
+
+**Layer 2 — controller replay.** Finalization, refutation, output validation, and the merge gate are controller-owned, so Layer 1 cannot reach them and Layer 3 is too heavy to run on every change. Layer 2 feeds a recorded queue snapshot to a controller and diffs the actions and writes it emits. A speculate run is a pure function of a single read, so a snapshot is a complete input; other snapshot-determined controllers can use the same mechanism.
+
+**Layer 3 — system simulation.** Some questions are about consequences rather than verdicts: does a looser `conflict.Analyzer` reduce land time once batching and speculation react to the changed dependency graph? That needs the real pipeline running forward — gateway, orchestrator, and Runway in one process, in-memory storage and message queues in place of MySQL, and fakes only at the boundaries reaching source control, CI, and git.
```
behinddwalls, line 52: > fakes only at the boundaries reaching source control, CI, and git.

do we need to consider this as well? https://github.com/uber/submitqueue/issues/209

behinddwalls, line 52: > in-memory storage and message queues in place of MySQL

would this emulate fakes? so say if we wanted to test a failure of Ack or Nack? can we simulate with it

doc/rfc/submitqueue/simulator.md
```diff
@@ -0,0 +1,258 @@
...
+Isolation is a property of the wiring rather than of configuration: a call reaching a boundary the harness has not faked fails the run. Intercepting only the boundaries known when the harness was written fails quietly, because the first new dependency anyone adds reaches production unreported.
+
+## How it works
+
+| Component | Responsibility | Layer |
+|---|---|---|
+| Corpus store | identities, recorded inputs and outputs, provenance | all |
+| Recorder | mines history, records at the seam, or shadows in production | all |
+| Replayer | drives one extension from corpus identities | 1 |
+| Snapshot replayer | drives one controller from a recorded queue snapshot | 2 |
```
behinddwalls, line 63: > Snapshot replayer

unsure of this vs just replaer? how they are different

doc/rfc/submitqueue/simulator.md
```diff
@@ -0,0 +1,258 @@
...
+```
+
+Layer 2 sits between the two: it loads one recorded queue snapshot, runs a single controller against it, and diffs the actions and writes that controller emits.
+
+## What a difference means
+
+A replay produces one of three outcomes, and collapsing them into pass and fail destroys the instrument.
+
+- **Harness fault.** The run did not complete: a fake failed, a resolver timed out, a candidate panicked. Says nothing about either implementation.
+- **Corpus fault.** The run completed, but the baseline no longer reproduces its own recorded output. Every candidate number derived from that entry is void.
```
behinddwalls, line 104: > Harness fault. The run did not complete: a fake failed, a resolver timed out, a candidate panicked. Says nothing about either implementation.
> Corpus fault. The run completed, but the baseline no longer reproduces its own recorded output. Every candidate number derived from that entry is void.

we need simplified terminologies? i am unsure of what harness means here?

doc/rfc/submitqueue/simulator.md
```diff
@@ -0,0 +1,258 @@
...
+- **Behavioral difference.** Both implementations ran and disagreed. This is the measurement, not an error.
+
+The third is why output equality cannot be the pass condition: a candidate `conflict.Analyzer` reporting different conflicts is doing the thing it was written to do, and a harness built around equality would report every interesting change as broken. What must hold instead is narrower — a candidate is free to change what an extension decides, and not free to change how the pipeline responds to a decision.
+
+## Reproducibility per extension
+
+An extension reproducible from a pinned identity can be replayed against any candidate, including on inputs history never saw. One that is not can only be compared against the single answer history recorded. That decides how its corpus is collected (see [Corpus collection](#corpus-collection)).
+
+| Extension | Reproducible? | Why |
```
behinddwalls, line 113: > Reproducible?

this is what currently implemented but not what might be eventually, for example conflicts would be coming from Tango, scorer will be ML based?

PR #600, opened 2026-08-14
fix(client): keep a watch's frame inside the window

behinddwalls:
### Why?

A watch of more requests than the window can hold repaints the whole screen on every redraw, flickering continuously until it settles.

The renderer never knew how tall the terminal was. `terminalSize` asked for the size and threw the height away, so a frame was allowed to be any number of lines. Each redraw then moves the cursor up by the number of lines it emitted, on the assumption that those lines are still on screen — and once a frame is taller than the window, they are not. Drawing it scrolled the window, and the cursor cannot move above its first row, so the terminal clamped the jump. Every frame after that started from the wrong origin, overwriting whatever happened to be at the top and leaving the rest of the previous frame stranded below.

The threshold is the window height, which is why this looks like a size problem rather than a bug: a short list is fine, and one row more than fits is not.

### What?

The renderer reads the height along with the width, re-reading both before every draw so a window resized mid-watch is picked up, and trims a frame that will not fit.

Settled requests are dropped first. They have stopped changing, so leaving them out costs a reader nothing `land-list` will not tell them, while dropping the ones still moving would hide the only part of the table that is doing anything. Within each group the original order is kept, so rows do not jump between redraws, and a line reports how many are not shown — a short table is never silently a partial one.

Two cases deliberately keep every row, because nothing is ever drawn over them and the terminal's own scrollback is the right answer:

- redirected output, which is a log rather than a window;
- a one-shot `list`, which is drawn once and scrolled back through rather than redrawn.

## Test Plan

- ✅ 30 requests in a 20-row window: the frame is 19 lines, the cursor moves back 19, and the footer reads `… 23 settled request(s) not shown`. Before this, the frame was over 60 lines in the same window
- ✅ `make land-list LIMIT=30` in that same 20-row window still prints all 30 rows
- ✅ new tests: a frame never exceeds the window and `lastLines` matches what was emitted; an unsettled row survives the trim while settled ones are dropped; piped output and one-shot listings keep every row
- ✅ mutation-checked the height test by making the trim a no-op, confirming it fails without the fix rather than passing vacuously
- ✅ `make test` (105 targets), `make lint`, `make gazelle`

### DIFF
--- a/doc/howto/QUICKSTART.md
+++ b/doc/howto/QUICKSTART.md
@@ -105,7 +105,16 @@ make land-list SINCE=24h LIMIT=200   # a wider window
 make land-watch                      # follow them until they settle
 ```
 
-Both draw the same table `make demo-requests` does — the demo tool and the CLI share it — but against whatever the queue already holds, so watching a queue no longer means adding to it. They do not carry the same information, though: `list` is a one-shot read of the queue's receipts and does not fetch histories, so its `STAGE` column is always `…`, while `watch` follows the history API and fills the trail in as each request moves.
+Both draw the same table `make demo-requests` does — the demo tool and the CLI share it — but against whatever the queue already holds, so watching a queue no longer means adding to it. They do not carry the same information, though: `list` is a one-shot read of the queue's receipts and does not fetch a history per request, so its `STAGE` column shows where each request is and not how it got there, while `watch` follows the history API and fills the whole trail in as each one moves.
+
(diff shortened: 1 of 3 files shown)


behinddwalls:
Folded into #598 — both changes are to what the watch shows, and the viewport clamp reviews better as part of that story than on its own. No content was lost: the folded branch is byte-identical to this one's tip.

PR #607, opened 2026-08-17
feat(hook): Deliver events to integrations

prathshenoy:
## Summary
**What**:
- Consume lifecycle events from a durable queue and hand each one to the integrations it resolves to, running them concurrently and retrying then dead-lettering the ones that keep failing.
- Record every dead-lettered event with its full payload and failure reason, and count it on a metric meant to page.

**Why**:
- Keep a slow or broken integration from stalling or failing the work that triggered it.
- Make a lost notification recoverable and impossible to miss, since nothing else in the system notices when one goes missing.

## Test Plan
- [x] Add unit tests.

## Revert Plan
- Revert this PR. No service constructs the controller or subscribes to the hook topic.

## Issues
- [CODEM-416](https://linear.app/uber/issue/CODEM-416/hooks-integration-downstream-notificaiton)


behinddwalls:

The rebase of `prath.shenoy/hook-dispatch` completed without conflicts, but the resulting code diff does not match the original. This means the rebase silently altered code content. The branch was **not** force-pushed.

platform/extension/hook/composite/hook.go
```diff
@@ -0,0 +1,80 @@
...
+// failures, each attributed to the child that raised it, or nil when all
+// succeeded.
+func (h Hook) Handle(ctx context.Context, event *basehook.HookEvent) error {
+	var failures []error
+	for _, child := range h.children {
+		if err := child.Handle(ctx, event); err != nil {
+			failures = append(failures, fmt.Errorf("hook %s: %w", child.Name(), err))
+		}
+	}
+	return errors.Join(failures...)
```
mnoah1: `errors.Join` creates an error tree (`Unwrap() []error`), but `platform/errs.NewClassifierProcessor` only walks linear `Unwrap() error` chains. Plain transient errors returned by child hooks can therefore bypass configured classifiers and default to non-retryable, sending the event directly to the DLQ. Let's think through how to make the error classifier properly unwrap joined error chains and classify them for retryability.

platform/extension/hook/composite/hook.go
```diff
@@ -0,0 +1,80 @@
...
+func New(children ...hook.Hook) Hook {
+	return Hook{children: children}
+}
+
+// Handle implements hook.Hook. It runs every child and returns the joined
+// failures, each attributed to the child that raised it, or nil when all
+// succeeded.
+func (h Hook) Handle(ctx context.Context, event *basehook.HookEvent) error {
+	var failures []error
+	for _, child := range h.children {
```
mnoah1: Should these independent hooks run concurrently rather than serially?  This would avoid one slow hook from blocking others from starting.  Just need to be careful to avoid any leaked goroutines here.

platform/extension/hook/hook.go
```diff
@@ -0,0 +1,65 @@
...
+// limitations under the License.
+
+// Package hook defines the contract for a hook: a pluggable side effect run in
+// response to a pipeline lifecycle event. Warehouse exports, code-host comments,
+// notifications, and audit trails are all hooks.
+//
+// A hook is wired once per host rather than resolved per queue, because what an
+// integration does — post a comment, write a row — is a property of the
+// deployment, not of the queue the event came from. There is therefore no Config
+// and no Factory here: the host constructs its hook directly and hands it to the
```
behinddwalls: not really, different queues can point to dfferent providers so each queue should be able define and pick what hooks it wants to wire up

platform/extension/hook/hook.go
```diff
@@ -0,0 +1,65 @@
...
+//go:generate mockgen -source=hook.go -destination=mock/hook_mock.go -package=mock
+
+import (
+	"context"
+
+	basehook "github.com/uber/submitqueue/api/base/hook"
+)
+
+// Hook performs a side effect in response to a lifecycle event.
+type Hook interface {
```
behinddwalls, line 37: my general question here is why is it an extension vs a plain consumer who gets hooks events from wire and implements consumer controller? why it needs to be an interface?

behinddwalls (dismissed):

platform/extension/hook/composite/hook.go
```diff
@@ -0,0 +1,80 @@
...
+// New returns a Hook that hands each event to every child in the order given.
+// With no children it accepts every event and does nothing.
+func New(children ...hook.Hook) Hook {
+	return Hook{children: children}
+}
+
+// Handle implements hook.Hook. It runs every child and returns the joined
+// failures, each attributed to the child that raised it, or nil when all
+// succeeded.
+func (h Hook) Handle(ctx context.Context, event *basehook.HookEvent) error {
```
behinddwalls: why do we need composite all all? if i understand correctly, dispatcher can dispatch different hooks, is there a need to use composite which dispatching or even using that from callsite for publishing...

platform/hook/dlq.go
```diff
@@ -0,0 +1,130 @@
...
+// recover. So this controller makes the loss impossible to miss and hands it
+// over: it records the complete event and why it failed, counts it on a metric
+// meant to page, and acks so the event does not sit in the DLQ unnoticed.
+// Republishing the logged event recovers it.
+//
+// That is a deliberate step up from the log topic's DLQ, which warns and moves
+// on. Dropping an observability row costs a gap in a read model; dropping a
+// merge-failure comment costs a support ticket, and nothing else in the system
+// will notice it is missing.
+type DLQController struct {
```
behinddwalls, line 46: do we need to provide DLQ Hook extension as well? so we do think in most cases, there is nothing to really reconcile, but say for example, I want to close PR when merge is done and for whatever reason that fails..should we have ability to handle those from retries and DLQ? just a thought, we don't need to support as such

prathshenoy: I haven't thought it much, but that doesn't feel like reconciliation. In that example, the hook has no pipeline state that we can reconcile. Retrying the external effect should be handled by republishing the event (which we already do), but maybe I'm missing something.

platform/extension/hook/hook.go
```diff
@@ -0,0 +1,84 @@
...
+	Handle(ctx context.Context, event *basehook.HookEvent) error
+
+	// Name identifies the hook in logs, metrics, and the failure attribution
+	// the controller reports. Stable and unique among the hooks a host wires.
+	Name() string
+}
+
+// Hooks resolves the hooks that run for an event.
+type Hooks interface {
+	// For returns the hooks to run for event, in the order they should run.
```
mnoah1: these run concurrently now - so can note that execution ordering is nondeterministic

mnoah1 (approved):

PR #608, opened 2026-08-18
feat(orch): Consume hook events

prathshenoy:
## Summary
**What**:
- Register the orchestrator's hook stage, so lifecycle events are consumed from the queue and handed to the integrations each event resolves to. No integration is wired yet, so events are accepted and discarded.
- Register the matching dead-letter stage, so an event that fails every retry is recorded and alerted on instead of accumulating unread.

**Why**:
- Prove the delivery path end to end before anything produces an event, so a later change adds producers to a route that already works.
- Keep a deployment with no integrations configured distinguishable from one that is silently losing events.

## Test Plan
- [x] Add unit tests.

## Revert Plan
- Revert this PR. Since nothing publishes hook events yet and the resolver returns only a no-op hook, no side effect is lost.

## Issues
- [CODEM-416](https://linear.app/uber/issue/CODEM-416/hooks-integration-downstream-notificaiton)

## Stack
1. #607
1. @ #608
1. #632

### DIFF
--- a/service/submitqueue/orchestrator/server/BUILD.bazel
+++ b/service/submitqueue/orchestrator/server/BUILD.bazel
@@ -16,6 +16,7 @@ go_library(
     importpath = "github.com/uber/submitqueue/service/submitqueue/orchestrator/server",
     visibility = ["//visibility:private"],
     deps = [
+        "//api/base/hook:go_default_library",
         "//api/submitqueue/orchestrator/protopb:go_default_library",
         "//platform/buildkite:go_default_library",
         "//platform/errs/generic:go_default_library",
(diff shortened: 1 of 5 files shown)


mnoah1 (approved):
As a follow-up - you can wire this into stovepipe both here, and within our internal service.

behinddwalls:

**To fix manually:**
```bash
git fetch origin
git checkout prath.shenoy/hook-wiring
git rebase --onto origin/main e08c06bf4aad707f895d58dffb99131c423ea644 prath.shenoy/hook-wiring
# resolve conflicts, then:
git push --force-with-lease
```

```bash
gh pr edit 608 --base main
```

PR #613, opened 2026-08-19
refactor(runway): compose the checkout git env from gitexec

behinddwalls:
## Summary

runGit in the Runway server's checkout provisioning kept its own copy of the scrubbed-plus-transport environment. Build it through gitexec.Env instead — the same source the merger uses — so provisioning and merging assemble an identical environment from one definition.

This also closes a gap the two hand-written copies had already opened: provisioning omitted GIT_ATTR_NOSYSTEM and the no-pager/no-editor settings the merge path sets, so a system gitattributes file could influence the initial checkout but not later merge operations. Both now run under the same scrub set.

Provisioning also passes the runtime's PassthroughEnv through gitexec, as the merger already does, so a deployment whose remote needs an extra variable reaches it during the provisioning fetch and not merges alone. The field is unset today, so this only aligns the two paths rather than changing behavior.

## Test Plan

✅ `//service/runway/server` green. Provisioning and merge paths now assemble an identical git environment from one source.

## Stack
1. #609
1. #610
1. #611
1. #612
1. @ #613

### DIFF
--- a/service/runway/server/checkout.go
+++ b/service/runway/server/checkout.go
@@ -26,6 +26,7 @@ import (
 
 	"go.uber.org/zap"
 
+	gitexec "github.com/uber/submitqueue/platform/git/exec"
 	gitmerger "github.com/uber/submitqueue/runway/extension/merger/git"
 )
 
@@ -190,8 +191,9 @@ func setLocalConfig(checkoutPath, key, value string) error {
 }
 
 // runGit invokes the pinned git in dir with an environment scrubbed of ambient
-// configuration but retaining what is needed to reach a remote — the same split
-// the merger draws, so provisioning and merging authenticate identically.
+// configuration but retaining what is needed to reach a remote. It composes that
+// environment through gitexec.Env, the same source the merger uses, so
+// provisioning and merging authenticate — and behave — identically.
 func runGit(ctx context.Context, runtime gitmerger.GitRuntime, dir string, args ...string) ([]byte, error) {
 	full := append([]string{
 		"--exec-path=" + runtime.ExecPath,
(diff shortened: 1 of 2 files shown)


service/runway/server/checkout.go
```diff
@@ -200,27 +202,16 @@ func runGit(ctx context.Context, runtime gitmerger.GitRuntime, dir string, args
...
-		"GIT_SSH", "GIT_SSH_COMMAND", "GIT_SSH_VARIANT",
-		"GIT_SSL_CAINFO", "GIT_SSL_CAPATH", "SSL_CERT_DIR", "SSL_CERT_FILE",
-		"HTTP_PROXY", "HTTPS_PROXY", "NO_PROXY",
-		"http_proxy", "https_proxy", "no_proxy",
-	} {
-		if v, ok := os.LookupEnv(name); ok {
-			cmd.Env = append(cmd.Env, name+"="+v)
-		}
-	}
+	cmd.Env = gitexec.Env(gitexec.EnvOptions{
```
mnoah1, line 205: Should this include Passthrough: runtime.PassthroughEnv, matching the merger’s environment construction? Otherwise a custom transport variable works during merging but not during the provisioning fetch. The current server does not populate this field, but it leaves the two paths short of the stated identical environment.

mnoah1 (approved):

service/runway/server/checkout.go
behinddwalls: Good catch — done. `runGit` now passes `Passthrough: runtime.PassthroughEnv`, matching `newGitCommand` in the merger, so the two paths compose from the same shape. As you noted the field is unset today (`resolveGitRuntime` leaves it nil), so this changes no behavior now; it just closes the gap so that if a deployment ever sets `PassthroughEnv` for an unusual remote, the provisioning fetch honors it too and not merges alone. Wiring an actual config source for the field is a separate change — the merger's runtime resolution does not populate it either.

PR #621, opened 2026-08-19
fix(errs): Retry transient grouped failures

prathshenoy:
## Summary
**What**:
- Retry work when any one of several failures reported together is transient instead of giving up on the first attempt.
- Limit a classification attached to one grouped failure to that failure alone, instead of letting it decide the outcome for the others reported alongside it.

**Why**:
- Stop discarding failures that would have cleared on their own by keeping a step that fans work out to several handlers from being abandoned because one handler hit a momentary error.

## Test Plan
- [x] Add unit tests.

## Revert Plan
- Revert this PR. The changes in this PR are confined to the error classification framework.

## Issues
- [CODEM-462](https://linear.app/uber/issue/CODEM-462/err-classification-framework-should-handle-bundled-errors)

### DIFF
--- a/platform/errs/README.md
+++ b/platform/errs/README.md
@@ -44,12 +44,43 @@ An `ErrorProcessor` runs the per-chain pass that turns a raw chain into a wrappe
 Two implementations ship in this package:
 
 - **`NewClassifierProcessor(classifiers...)`** — the standard pass for primary pipeline consumers. Walks the chain twice:
-  1. **Pass 1 — framework-wrap check.** A cheap type switch looks for an existing `*userError` / `*infraError` anywhere in the chain. If found, the chain is already interpretable and the processor returns `err` unchanged. **No classifier is invoked.**
+  1. **Pass 1 — framework-wrap check.** Looks for an existing `*userError` / `*infraError` on the error's single-cause spine. If found, the chain is already interpretable and the processor returns `err` unchanged. **No classifier is invoked.**
   2. **Pass 2 — classifier walk.** From outermost to innermost node, each registered classifier is asked for a verdict. The first non-`Unknown` verdict wins and `err` is wrapped with the matching framework constructor.
 
   If no classifier recognises anything, `err` is returned unchanged — and behaves as non-retryable infra at the helper layer.
 
 - **`AlwaysRetryableProcessor`** — unconditionally wraps every non-nil error with `NewRetryableError`, overriding any inner framework wrap. Use it for narrowly-scoped consumers — typically DLQ reconciliation — that must redeliver on any failure because there is no further dead-letter destination. Side-effect: an inner `*infraError(dependency=true)` is masked by the outer `retryable=true` wrap, since `errors.As` matches the outermost `*infraError` first. This is acceptable for the intended DLQ use case where only `IsRetryable` drives transport behaviour; do not pair this processor with a primary pipeline consumer or genuine user errors will retry forever instead of reaching their DLQ.
+
+### Grouped errors
+
+`Group(errs...)` reports several failures that happened together as one error. It drops nils and returns nil when every member is nil, so a step that fans work out to independent handlers can accumulate failures in a loop and return the result directly:
+
+```go
+var failures []error
+for _, h := range handlers {
+    if err := h.Handle(ctx, event); err != nil {
+        failures = append(failures, fmt.Errorf("%s: %w", h.Name(), err))
+    }
+}
+return errs.Group(failures...)
+```
+
+Pass 2 descends into a group's members as well as into ordinary single-cause wraps; Pass 1 walks only the single-cause spine. Without that descent, `errors.Unwrap` returns nil for a group, so a walk built on it alone sees the group node and nothing beneath it, and every member goes unclassified.
+
+**Grouping is opt-in — only `Group` is weighed.** `Unwrap() []error` is also what `errors.Join` and `fmt.Errorf` with several `%w` produce, and there the extra causes are incidental: a cleanup failure hung off the real one, or context that happens to be an error. Weighing those would let an unrelated sibling decide retryability for the whole chain. `Group` is the one spelling that means "these failures are independent, rank them against each other", so the walk keys on that type and every other multi-cause error stays opaque — classified by its outermost recognisable node like any single-cause chain. A caller that wants its members ranked says so by returning `Group`, as `submitqueue/extension/validator/composite` does for its children.
+
+The two shapes combine differently, because they mean different things:
+
+- **Down a wrap chain**, the outermost verdict wins. A wrapper saw the error it wrapped and classified anyway, so it speaks with more knowledge than its cause.
+- **Across the members of a group**, nothing shadows anything. The members are independent failures reported together, and their order is the order the caller ran them in, not a precedence. They combine by rank, so the result cannot depend on which member failed first.
+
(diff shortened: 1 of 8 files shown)


platform/errs/processor.go
```diff
@@ -50,26 +50,30 @@ type ErrorProcessor interface {
...
 //     behavior for an unwrapped chain, so no wrap is added.
-//   - If no classifier recognises anything, err is returned unchanged.
+//   - If no classifier recognizes anything, err is returned unchanged.
 //
-// Implementation: two passes over the chain. Pass 1 is a cheap type check
-// looking for an existing framework wrap and short-circuits if one is found —
-// no classifier is invoked. Pass 2 runs the configured classifiers per node.
-// Walking the chain is cheap relative to a classifier call, so this avoids
-// running classifiers whenever the chain is already classified deeper down.
+// Implementation: two passes over the chain. Pass 1 looks for an existing
```
mnoah1, line 65: What if instead, we just create errs.Group(errors...) and let controllers return that to group errors:
- Internal groupedError type with members[]
- when it's a grouped error, each member gets passed through Process, which gets its classification
- the whole thing gets wrapped as retryable if any of them are retryable

Would be a bit more explicit than trying to unwrap from the results returned by errors.Join

platform/errs/processor.go
```diff
@@ -127,6 +122,113 @@ func (p classifierProcessor) Process(err error) error {
...
+// outcome.
+func verdictRank(v Verdict) int {
+	switch v {
+	case InfraRetryable:
+		return 5
+	case InfraDependencyRetryable:
+		return 4
+	case Infra:
+		return 3
+	case User:
```
behinddwalls, line 212: why user before infraDep?

platform/errs/processor.go
```diff
@@ -127,6 +122,113 @@ func (p classifierProcessor) Process(err error) error {
...
+		case e.dependency:
+			return InfraDependency
+		default:
+			return Infra
+		}
+	}
+	return Unknown
+}
+
+// verdictRank orders verdicts for combining the independent branches of a
```
behinddwalls: too verbose? please ask Claude to make it crisp... it looses the clarity

platform/errs/processor.go
```diff
@@ -127,6 +122,113 @@ func (p classifierProcessor) Process(err error) error {
...
+			}
+			return verdict
+		}
+
+		cur = errors.Unwrap(cur)
+	}
+	return Unknown
+}
+
+// wrapVerdict returns the verdict the framework wrap err carries, or Unknown if
```
behinddwalls: do we need the doc..i think ideally code is clear enough to not explain, unless you think there is need for one, may shorten it

behinddwalls (dismissed):

platform/errs/processor.go
```diff
@@ -50,26 +50,30 @@ type ErrorProcessor interface {
...
-// Walking the chain is cheap relative to a classifier call, so this avoids
-// running classifiers whenever the chain is already classified deeper down.
+// Implementation: two passes over the chain. Pass 1 looks for an existing
+// framework wrap and short-circuits if one is found — no classifier is invoked.
+// Pass 2 runs the configured classifiers per node. Walking the chain is cheap
+// relative to a classifier call, so this avoids running classifiers whenever
+// the chain is already classified deeper down.
 //
-// Passing no classifiers is valid — the processor will still honour any
+// Both passes traverse joined errors (errors.Join, or anything else exposing
```
behinddwalls: This says both passes traverse joins, but Pass 1 doesn't — it walks only via `errors.Unwrap() error`, which yields nothing for a join. The inline comment 26 lines below (L97-99) says exactly that: *"along the single-cause spine only… The walk stops at a join."*

Worth correcting precisely because the design depends on it: Pass 1 **not** descending is what lets a wrapped branch be ranked against its siblings instead of short-circuiting the whole join. `TestNewClassifierProcessor_WrappedBranchesAreWeighed` only passes because of it — if Pass 1 did descend, `Process` would return the branch's `*userError` verbatim rather than wrapping it in `NewRetryableError`.

Suggested scoping:

> Pass 2 traverses joined errors (errors.Join, or anything else exposing `Unwrap() []error`) as well as ordinary single-cause wraps; Pass 1 walks only the single-cause spine.

platform/errs/README.md
```diff
@@ -44,13 +44,30 @@ An `ErrorProcessor` runs the per-chain pass that turns a raw chain into a wrappe
...
 
+### Joined errors
+
+Both passes descend into joined errors — `errors.Join`, `fmt.Errorf` with more than one `%w`, or any other error exposing `Unwrap() []error` — as well as ordinary single-cause wraps. This matters for anything that fans work out to several children and reports their failures together, `submitqueue/extension/validator/composite` being the current example: `errors.Unwrap` returns nil for a join, so a walk built on it alone sees the join node and nothing beneath it, and every branch goes unclassified.
```
behinddwalls: Same issue as `processor.go` L71 — Pass 1 doesn't descend into joins, only Pass 2 does. This one also contradicts L47 nine lines up, which correctly says Pass 1 looks *"on the error's single-cause spine."*

`s/Both passes descend/Pass 2 descends/` and the rest of the paragraph reads fine as-is.

platform/errs/README.md
```diff
@@ -44,13 +44,30 @@ An `ErrorProcessor` runs the per-chain pass that turns a raw chain into a wrappe
...
+
+A framework wrap classifies the subtree beneath it and no further. Above a join it covers the whole join and Pass 1 returns the error verbatim, so no branch is consulted. *Inside* a branch it is one branch's account of one failure, with no standing to classify the failures beside it — so it contributes its own verdict to the rank like any other branch. That is what keeps a sibling's transient failure from being discarded by a branch that happened to arrive pre-classified, and it also removes an ordering artifact: two wrapped branches of differing retryability used to resolve by whichever one `errors.As` reached first.
+
+The losing branch keeps its wrap in the chain, so `IsUserError` and `IsRetryable` can both report true for the same joined error — one from a branch, one from the outer wrap. Only the outer wrap drives the retry decision, the same way it does under `AlwaysRetryableProcessor`.
```
behinddwalls: Design question rather than a defect — this is clearly deliberate, and documented.

I checked every current caller and they all test `IsRetryable` first, so nothing is broken today:

- `platform/consumer/consumer.go:442` — `if !errs.IsRetryable(err)`, the actual retry decision
- `platform/consumer/consumer.go:644` — `if IsRetryable {} else if IsUserError {}`, metrics tagging only
- `stovepipe/controller/process/process.go:144` — `if !errs.IsRetryable(err)`

The concern is reachability. Dual-true went from "a controller had to deliberately double-wrap" to "join any two children," but nothing in the API enforces the precedence. A future caller writing the equally natural `if IsUserError(err) { ack } else if IsRetryable(err) { retry }` would silently drop a transient failure — this PR's own bug arriving through a different door.

Might be worth either a contract note on `IsUserError` saying it must not be consulted before `IsRetryable`, or a single `RetryDecision(err)` helper that makes the precedence unskippable. Not blocking.

platform/errs/processor.go
prathshenoy: It ranks the nearest actor first: our own defect, then the requester's request, then a third party perhaps nobody here can act on.

platform/errs/README.md
prathshenoy: I added a note in the README.

mnoah1 (approved):

PR #622, opened 2026-08-19
fix(messagequeue): run queue garbage collection on busy partitions

Jal-Bafana:
## Why?

Garbage collection of acknowledged `queue_messages` rows currently runs only on idle poll ticks.

A continuously busy partition resets the GC counter whenever it delivers a message, so it can never reach the GC threshold. Under sustained traffic, acknowledged message rows can therefore accumulate indefinitely.

This also means the message deduplication horizon can remain unnecessarily large on busy partitions.

## What?

Run the existing garbage collection cadence on every poll tick rather than only idle ticks.

The existing 100-tick throttle is preserved, so the per-partition GC frequency is not increased. The GC safety mechanism is also unchanged: garbage collection still uses the minimum acknowledged offset across consumer groups.

Added:
- unit regression coverage proving GC runs on busy ticks
- MySQL integration coverage proving acknowledged rows are reclaimed under continuous traffic

## Testing

- `go test ./platform/extension/messagequeue/...` — PASS
- `go test ./platform/base/...` — PASS
- `go vet` on the messagequeue and integration packages — PASS
- `gofmt` and `git diff --check` — clean

Not run locally due to environment limitations:
- MySQL integration test `TestGCReclaimsAckedRowsUnderContinuousTraffic` — Docker daemon unavailable
- `make gazelle`, `make fmt`, `make lint`, `make check-tidy`, `make check-gazelle` and Bazel tests — `make` unavailable and Bazel's Windows C++ toolchain is not configured

### DIFF
--- a/platform/extension/messagequeue/mysql/subscriber.go
+++ b/platform/extension/messagequeue/mysql/subscriber.go
@@ -49,12 +49,6 @@ const (
 	// so it converges over multiple calls even with large backlogs.
 	watermarkAdvancementLimit = 1000
 
-	// gcIdleTickInterval controls how often GC runs during idle poll ticks.
-	// GC runs every Nth idle tick instead of every tick to avoid excessive
-	// queries when many partitions are idle (e.g., 50 idle partitions at 100ms
-	// poll interval = 500 GC queries/sec without throttling).
-	gcIdleTickInterval = 100
-
 	// heartbeatPurgeAfterLeaseDurations sets the age threshold for purging
 	// abandoned heartbeat rows, as a multiple of LeaseDurationMs (10x = 5min
 	// at defaults). Well past every transient window in the protocol — a row
@@ -80,6 +74,13 @@ const (
 	// refresh or remove) a stale lease on a partition with no messages.
 	leasePurgeAfterLeaseDurations = 10
 )
+
+// gcTickInterval is the number of poll ticks between garbage collection runs.
+// GC runs every Nth tick regardless of delivery activity; without throttling,
+// many partitions polling in lockstep would flood the store with queries
+// (e.g., 50 partitions at 100ms poll interval = 500 GC queries/sec). A var so
+// tests can shorten it; production always uses the default.
+var gcTickInterval = 100
 
 // HookSignal identifies the type of subscriber lifecycle event.
 // Named after behavioral concerns (what happened) rather than implementation
@@ -167,8 +168,7 @@ type partitionWorker struct {
 	// partition. Set once on the first successful poll, avoiding repeated
 	// initialization calls on every tick.
 	offsetInitialized bool
-	// gcCounter counts idle poll ticks. GC only runs every gcIdleTickInterval
-	// ticks to avoid excessive queries when many partitions are idle.
+	// gcCounter counts poll ticks since the last garbage collection run.
 	gcCounter int
 }
 
@@ -1200,17 +1200,14 @@ func (w *partitionWorker) pollAndDeliver(ctx context.Context) (retErr error) {
 		)
 	}
 
-	// Run GC periodically (throttled to every Nth idle tick)
-	if messageCount == 0 {
-		w.gcCounter++
-		if w.gcCounter >= gcIdleTickInterval {
-			w.gcCounter = 0
-			if err := w.garbageCollect(ctx); err != nil {
-				return fmt.Errorf("garbage collect: %w", err)
-			}
-		}
-	} else {
+	// GC runs every Nth tick regardless of delivery activity; an idle-only
+	// gate starved continuously busy partitions of garbage collection.
+	w.gcCounter++
+	if w.gcCounter >= gcTickInterval {
 		w.gcCounter = 0
+		if err := w.garbageCollect(ctx); err != nil {
+			return fmt.Errorf("garbage collect: %w", err)
+		}
 	}
 
 	// Record poll metrics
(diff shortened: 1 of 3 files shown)


behinddwalls (commented):
Two notes on the diff — the core change looks right to me. Removing the `messageCount == 0` gate genuinely fixes the starvation (the old `else` branch reset the counter to `0` forever on a continuously busy partition), GC load stays bounded the same way it was, and `run()` only logs `pollAndDeliver` errors so a GC failure can't kill a worker or cause re-delivery.

platform/extension/messagequeue/mysql/subscriber.go
```diff
@@ -1200,17 +1200,14 @@ func (w *partitionWorker) pollAndDeliver(ctx context.Context) (retErr error) {
...
-			}
-		}
-	} else {
+	// GC runs every Nth tick regardless of delivery activity; an idle-only
+	// gate starved continuously busy partitions of garbage collection.
+	w.gcCounter++
+	if w.gcCounter >= gcTickInterval {
 		w.gcCounter = 0
+		if err := w.garbageCollect(ctx); err != nil {
+			return fmt.Errorf("garbage collect: %w", err)
```
behinddwalls, line 1217: This early return now suppresses the `messages_delivered` counter on a busy tick.

Before this change GC could only run when `messageCount == 0`, so the `garbage collect` error path could never skip the `if messageCount > 0` metrics block a few lines below. Now it can: the messages were already pushed to `deliveryCh`, but the counter is dropped — so throughput reads low exactly when the store is unhealthy and you most want the number.

Moving the metrics block above the GC block (or folding it into the existing `defer`) fixes it.

test/integration/extension/messagequeue/mysql/queue_test.go
```diff
@@ -2268,6 +2268,67 @@ func (s *SQLQueueIntegrationSuite) TestIdleLeaseRelease() {
...
+	// because only GC deletes message rows.
+	const initialBatch = 200
+	for i := 0; i < initialBatch; i++ {
+		require.NoError(t, q.Publisher().Publish(s.ctx, topic,
+			entityqueue.NewMessage(fmt.Sprintf("gc-%d", i), []byte("x"), partition, nil)))
+	}
+	receiveN(t, deliveryChan, initialBatch, func(d extqueue.Delivery, _ int) {
+		require.NoError(t, d.Ack(s.ctx))
+	})
+	require.Equal(t, initialBatch, countMessages())
```
behinddwalls, line 2319: This assertion assumes GC hasn't fired yet, but the change under test is what can invalidate that — the drain above *is* busy ticks, which now increment `gcCounter`.

The subscription is live before the publishes, so the counter accrues across both the publish phase and the drain. At `PollIntervalMs = 50` that lands somewhere around 30–80 ticks on a fast box and passes; if the 200 inserts take more than ~3.5s on a loaded CI runner with Dockerized MySQL, the counter crosses the 100-tick threshold, GC reclaims acked rows mid-drain, and this `require.Equal` fails. Roughly 2x margin against a wall-clock threshold.

Asserting `countMessages() > 0` here, or seeding the acked backlog before subscribing, removes the timing dependency without weakening what the test actually proves (the `waitForCondition` below is the real assertion).

behinddwalls (changes requested):

test/integration/extension/messagequeue/mysql/queue_test.go
```diff
@@ -2268,6 +2268,67 @@ func (s *SQLQueueIntegrationSuite) TestIdleLeaseRelease() {
...
+// on idle poll ticks, so a continuously busy partition grew its message log
+// without bound; it must now run on its own tick cadence regardless of traffic.
+func (s *SQLQueueIntegrationSuite) TestGCReclaimsAckedRowsUnderContinuousTraffic() {
+	t := s.T()
+
+	topic := "gc_busy_topic"
+	partition := "gc-busy-part"
+	consumerGroup := "gc-busy-cg"
+
+	signalCh := make(chan queueMySQL.HookSignal, 100)
```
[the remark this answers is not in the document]
Jal-Bafana: Fixed in d3b5968: `drainSignals(signalCh)` now runs after the initial backlog drain and inside every traffic-loop iteration, so buffered signals are cleared before they can approach the 100-slot capacity while the worker emits a blocking signal on every poll tick.

platform/extension/messagequeue/mysql/subscriber.go
Jal-Bafana: Fixed in d3b5968: the `messages_delivered` counter is now recorded before the GC block, so a GC failure's early return can no longer suppress the metric for messages already pushed to the delivery channel on that tick. The busy-tick cadence and the 100-tick throttle are unchanged.

test/integration/extension/messagequeue/mysql/queue_test.go
Jal-Bafana: Fixed in d3b5968: the 200-message backlog is now seeded before Subscribe, so the GC counter starts accruing only when polling begins — roughly 20-30 ticks through the drain, well below the 100-tick threshold — and the stronger require.Equal(200, countMessages()) assertion stays instead of weakening it to > 0. The traffic loop also drains signals per iteration so the blocking signal send cannot stall the worker mid-loop.

Jal-Bafana:
@behinddwalls I’ve addressed all the review comments and pushed the changes. The review threads have also been replied to and resolved. Could you please re-check the latest changes and, if everything looks good, dismiss the previous changes-requested status / approve the PR? Thanks again for the feedback!

behinddwalls (approved):

Jal-Bafana:
Thanks for the review and feedback, @behinddwalls. Glad we got it merged!

PR #626, opened 2026-08-20
feat(speculation): predict a batch's outcome from its scorer price and its builds

behinddwalls:
## Summary
A batch's price reflects only its diff size; everything the queue learns during the run (a build passed, the batch is merging) is thrown away. Add a `predictor.Predictor` contract that takes the batch's build progress and revises the `Scorer`'s price with it. Implementation `predictor/evidence` multiplies the price's odds by one factor per piece of evidence — logistic regression in logs, so hand-set factors can later be fitted. Path sets come from the Speculator's single-read snapshot, not a store re-read. Config `predictor:` factors default to 1, so behavior is unchanged until set; shipped values are placeholders.

Design: `doc/rfc/submitqueue/outcome-predictor.md`.

## Test Plan
✅ `bazel test //submitqueue/... //service/... //platform/...` — passes
✅ New: factor application, compounding failures, all-succeed caveat, extreme prices, error cases, `New` validation, per-dependency path sets, config defaults/validation
✅ `make fmt`, `make gazelle`, `make mocks` — idempotent


sbalabanov:
plz simplify Summary...

behinddwalls:
Restacking: RFC is #682; impl will reopen with base preetam/outcome-predictor-rfc.

PR #630, opened 2026-08-24
feat(stovepipe): tag controller metrics by queue

mnoah1:
## Summary

Tag Stovepipe message-processing metrics for the process, build, buildsignal, record, and DLQ controllers, together with consumer-framework metrics, using the Stovepipe queue that owns each message. `publishProcess` places the queue name in context immediately before the initial process publish; the publish framework persists it under the message metadata key `queue_name` and propagates it across later publishes. Before invoking the gate or controller, the consumer restores the messagequeue context and explicitly attaches the metric tag `queue` (for example, `queue=monorepo/main`) through `metrics.WithContextTags`. `metrics.TagsFromContext` reads only those generic metrics-owned tags.

## Intent

- Make process, build, buildsignal, record, gate, acknowledgement, retry, and DLQ activity attributable to a queue.
- Make the queue available even for failures that happen before the payload can be decoded.
- Keep metric attribution request-local and safe when shared controllers process messages concurrently.
- Prevent individual pipeline stages from accidentally dropping queue attribution when publishing the next message.
- Keep the central metrics package independent of messagequeue-specific context.

## Approach and reasoning

- **Use message metadata instead of the partition value.** A partition key controls transport ordering and is not guaranteed to represent the Stovepipe queue for every controller. `queue_name` states the intended meaning directly.
- **Do not derive the tag from the payload.** Payload queue fields remain authoritative for storage lookup and validation, but deserialization can fail before they are available. Metadata lets early consumer, gate, decode-error, and DLQ metrics still carry the queue.
- **Translate at the delivery boundary.** The consumer reads the allowlisted `queue_name` before checking the gate or calling `controller.Process`. It stores the queue separately in messagequeue context for republishing and in metrics context as `queue` for attribution.
- **Keep metrics generic.** `platform/metrics` owns only a private collection of `metrics.Tag` values through `WithContextTags` and `TagsFromContext`; it does not import or understand messagequeue. Future platform code can select bounded metric dimensions at its own boundary without extending the metrics package.
- **Propagate in the publish framework.** `publish.Message` and `publish.MessageWithMetadata` clone caller metadata and add the messagequeue context value when the caller did not supply `queue_name`. Stages no longer need to re-add queue metadata manually, while explicit metadata and intentional overrides remain supported.
- **Keep context request-local.** Per-delivery context avoids mutating shared controllers or Tally scopes, so concurrent messages cannot overwrite one another’s queue attribution.
- **Allowlist context-derived data.** Other message metadata has separate semantics, such as failure details. The consumer explicitly maps only `queue_name` to an automatic metric tag.
- **Remain rollout-safe.** Messages published before `queue_name` was introduced simply emit metrics without the `queue` tag.

## Changes

- Seed queue context immediately before the initial Stovepipe process publish.
- Automatically persist and propagate it as `queue_name` metadata through both platform publish helpers.
- Restore only `queue_name` to messagequeue context before gate and controller processing.
- Add generic `metrics.WithContextTags` and `metrics.TagsFromContext` APIs with no messagequeue dependency.
- Map `queue_name` to the `queue` metric tag once in the consumer.
- Apply context-derived tags to process, build, buildsignal, record, and DLQ controller metrics, plus gate and message-outcome metrics.
- Cover propagation, metadata merging and immutability, explicit overrides, missing metadata, generic context tags, and emitted queue-tagged series.

## Test Plan

- `./tool/bazel test //platform/base/messagequeue:go_default_test //platform/publish:go_default_test //platform/consumer:go_default_test //platform/metrics:go_default_test //stovepipe/controller:go_default_test //stovepipe/controller/build:go_default_test //stovepipe/controller/buildsignal:go_default_test //stovepipe/controller/dlq:go_default_test //stovepipe/controller/process:go_default_test //stovepipe/controller/record:go_default_test`
- Deploy and verify affected metric series contain `queue=<queue-name>` while older messages without metadata continue processing normally.

## Revert Plan

- Revert the commit to remove queue metadata propagation and restore the prior untagged metrics.

## Issues

- None.

### DIFF
--- a/platform/metrics/metrics.go
+++ b/platform/metrics/metrics.go
@@ -30,9 +30,34 @@ type Tag struct {
 	Value string
 }
 
+type contextTagsKey struct{}
+
 // NewTag creates a Tag with the given key and value.
 func NewTag(key, value string) Tag {
 	return Tag{Key: key, Value: value}
+}
+
+// WithContextTags returns a child context carrying the supplied metric tags.
+// Tags already carried by ctx are preserved, and both inputs are copied.
+func WithContextTags(ctx context.Context, tags ...Tag) context.Context {
+	existing, _ := ctx.Value(contextTagsKey{}).([]Tag)
+	contextTags := make([]Tag, 0, len(existing)+len(tags))
+	contextTags = append(contextTags, existing...)
+	contextTags = append(contextTags, tags...)
+	return context.WithValue(ctx, contextTagsKey{}, contextTags)
+}
+
+// TagsFromContext appends metric tags explicitly carried by ctx. Additional
+// tags are preserved, and context-derived tags win when the same key is present.
+func TagsFromContext(ctx context.Context, tags ...Tag) []Tag {
+	stored, _ := ctx.Value(contextTagsKey{}).([]Tag)
+	if len(stored) == 0 {
+		return tags
+	}
+
+	contextTags := make([]Tag, 0, len(tags)+len(stored))
+	contextTags = append(contextTags, tags...)
+	return append(contextTags, stored...)
 }
 
 // Common duration bucket sets for latency histograms. Operations differ widely
--- a/platform/base/messagequeue/BUILD.bazel
+++ b/platform/base/messagequeue/BUILD.bazel
@@ -2,14 +2,20 @@ load("@rules_go//go:def.bzl", "go_library", "go_test")
 
 go_library(
     name = "go_default_library",
-    srcs = ["message.go"],
+    srcs = [
+        "context.go",
+        "message.go",
+    ],
     importpath = "github.com/uber/submitqueue/platform/base/messagequeue",
     visibility = ["//visibility:public"],
 )
 
 go_test(
     name = "go_default_test",
-    srcs = ["message_test.go"],
+    srcs = [
+        "context_test.go",
+        "message_test.go",
+    ],
     embed = [":go_default_library"],
     deps = ["@com_github_stretchr_testify//assert:go_default_library"],
 )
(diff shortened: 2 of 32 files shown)


platform/metrics/metrics.go
```diff
@@ -35,6 +36,21 @@ func NewTag(key, value string) Tag {
 	return Tag{Key: key, Value: value}
 }
 
+// TagsFromContext appends the allowlisted metric tags carried by ctx. Additional
+// tags are preserved, and context-derived tags win when the same key is present.
+// Messages published before a context value was introduced return only the
+// additional tags.
+func TagsFromContext(ctx context.Context, tags ...Tag) []Tag {
+	queueName, ok := entityqueue.QueueName(ctx)
```
behinddwalls: i am wondering if this should depend the base messagequeue context or not? this is bit central in nature so if we add this and say we need something else in some other part of the platform would we also put it here? can we pass the keys which we need to extract in the call itself so we don't need to depend on `messagequeue` entity itself

behinddwalls (changes requested):

platform/metrics/metrics.go
mnoah1: Addressed in `e5a3500`. `platform/metrics` no longer depends on messagequeue. `consumer.processDelivery` now explicitly maps the allowlisted `queue_name` metadata to `metrics.WithContextTags(..., metrics.NewTag("queue", ...))`, and `TagsFromContext` reads only metrics-owned context tags. I kept selection at the delivery boundary instead of passing messagequeue accessors at every metric call.

behinddwalls (approved):

PR #633, opened 2026-08-26
docs(stovepipe): propose request log

mnoah1:
## Summary

Intent:
- Define an append-only internal request log that explains how Stovepipe reached a request outcome.
- Match SubmitQueue terminology: retained domain records are a Log, while the user-facing projection is History.

Changes:
- Specify the `RequestLog` entity, state and event vocabularies, stable identities, and queue-scoped storage contract.
- Keep core lifecycle data in columns and move occurrence-specific context such as superseding request ID, build ID, and fact degree into metadata.
- Treat metadata as opaque JSON at the entity and storage boundary; key-specific enforcement is deferred to later writer and projection work.
- Exclude the raw metadata map from the public history API.
- Define source-write, log-write, and dependent-publish ordering together with idempotent retry and rollout requirements.

## Test Plan

- `./tool/bazel test //api/stovepipe/... //service/stovepipe/... //stovepipe/...`
- `make lint check-tidy check-gazelle`

## Revert Plan

Revert this documentation-only PR.

## Jira Issues

None.

### DIFF
--- a/doc/rfc/stovepipe/request-log.md
+++ b/doc/rfc/stovepipe/request-log.md
@@ -1,0 +1,271 @@
+# Stovepipe Request Log
+
+## Summary
+
+Stovepipe retains an append-only request log for each validation request. Its internal `RequestLog` is the counterpart of SubmitQueue's `RequestLog`: both retain request status changes and explanatory lifecycle events, while Stovepipe persists records directly instead of sending them through a cross-service log topic and materializer. The public API presents these records as request history. The log records every durable `Request.State` transition plus three asynchronous milestones needed to explain those transitions and the public verdict:
+
+- `build_triggered`;
+- `build_finished`;
+- `validation_fact_recorded`.
+
+The model deliberately follows SubmitQueue's distinction between statuses describing where a request is and events describing important activity that does not move it. It remains a bounded request-lifecycle log rather than a generic event bus or an audit of every correlated operation.
+
+`Request`, `Build`, and `ValidationFact` remain the operational sources of truth. The request log is a required audit/read model derived from successful durable operations; it does not replace their optimistic-locking state machines.
+
+The separate [Stovepipe Request History API](request-history-api.md) defines public lookup, projection, materialization, ordering, and retention semantics.
+
+Related documents are [SubmitQueue Gateway Request History APIs](../submitqueue/history-api.md), [Stovepipe Workflow](workflow.md), [Process](steps/process.md), [Build](steps/build.md), [Buildsignal](steps/buildsignal.md), and [Record](steps/record.md).
+
+## Problem and Scope
+
+Stovepipe currently retains only the latest mutable state of a validation request and build. A `ValidationFact` retains the immutable verdict for a commit, but none of these records explains how a request reached that verdict.
+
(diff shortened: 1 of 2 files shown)


doc/rfc/stovepipe/request-event-history.md
```diff
@@ -0,0 +1,290 @@
+# Stovepipe Request Event History
```
behinddwalls: is this going to be synonymous to SQ request log?

doc/rfc/stovepipe/request-log.md
```diff
@@ -0,0 +1,290 @@
...
+| Build finished | Request ID, event kind, and build ID |
+| Validation fact recorded | Request ID, event kind, and whole-repository fact identity |
+
+The recorder calls `RequestHistoryStore.Create`. If the ID already exists, it loads the stored entry and compares every semantic field. Identical content is idempotent success; conflicting content is an internal consistency error, and the stored entry is never overwritten.
+
+## Storage Contract
+
+The queue-scoped store is append-only and key/value-shaped:
+
+```go
```
behinddwalls, line 163: > type RequestHistoryCursor struct {

can we use page tokens? similar to SQ

doc/rfc/stovepipe/request-event-history.md
mnoah1: Yes - after taking another look, I see that SQ splits the History and Log terminology between API level vs. internal/storage layer.  Just updated this RFC to use `Log` here, and we'll keep `History` for the API (next PR).

doc/rfc/stovepipe/request-log.md
mnoah1: Ended up removing pagination here - SQ does not paginate for History and we should only be logging a limited number of key events per request and avoid unbounded growth here.

doc/rfc/stovepipe/request-log.md
```diff
@@ -0,0 +1,272 @@
...
+    // RequestVersion is the successful request version recorded by a state entry. It is zero on an event entry.
+    RequestVersion int32
+    // SupersededByRequestID identifies the newer request that caused a superseded outcome.
+    SupersededByRequestID string
+    // BuildID identifies the build associated with the occurrence, or is empty when no build applies.
+    BuildID string
+    // OutcomeReason is the domain reason for a terminal request state and never names a transport mechanism.
+    OutcomeReason RequestOutcomeReason
+    // FactDegree is the recorded health degree on validation_fact_recorded. The event kind distinguishes degree zero from absence.
+    FactDegree float64
```
mnoah1: note: we'll add metadata later if needed but it could fit here.

doc/rfc/stovepipe/request-log.md
```diff
@@ -0,0 +1,272 @@
...
+    Queue string
+    // RequestID identifies the request whose log contains this record.
+    RequestID string
+    // TimestampMs is the durable occurrence time in Unix milliseconds.
+    TimestampMs int64
+    // State is the durable request state recorded by a state entry. It is unset on an event entry.
+    State RequestState
+    // Event identifies the occurrence recorded by an event entry. It is unset on a state entry.
+    Event RequestEvent
+    // RequestVersion is the successful request version recorded by a state entry. It is zero on an event entry.
```
mnoah1: Let's move these into metadata.

behinddwalls (approved):

PR #637, opened 2026-08-26
feat(stovepipe): define GetProjectStatusByURI API

mnoah1:
## Summary
Define and publish the Stovepipe `GetProjectStatusByURI` current-state API for queue- and commit-scoped validation lookup.

## Intent

- Provide CD clients with the current request lifecycle and durable validation result for an exact commit URI.
- Keep the RFC and reviewed protobuf contract together in one core design change.
- Reserve a paginated project-result shape for the later project-level rollout.

## Changes

- Adds the `GetProjectStatusByURI` request, response, RPC, and generated gRPC/YARPC bindings.
- Uses optional scalar breakage degrees so an absent result remains distinct from fully green degree zero.
- Defines authoritative request selection, public lifecycle states, project-list completion, pagination, and phased rollout behavior.
- Links the RFC to the published protobuf types instead of duplicating the wire contract.


api/stovepipe/proto/stovepipe.proto
```diff
@@ -54,11 +54,65 @@ message IngestResponse {
...
+    optional string project = 3;
+    // Maximum projects to return. Zero selects the server default; negative values and values
+    // above the server maximum are invalid.
+    int32 page_size = 4;
+    // Opaque continuation token. Empty selects the first page.
+    string page_token = 5;
+}
+
+// ProjectValidation contains one planned project and any recorded result.
+message ProjectValidation {
```
behinddwalls: should this be just ProjectStatus

api/stovepipe/proto/stovepipe.proto
```diff
@@ -54,11 +54,65 @@ message IngestResponse {
...
+    string base_uri = 4;
+    // Stable public lifecycle state and terminal outcome of the request.
+    string request_state = 5;
+    // Breakage degree measures how broken the whole repository is on [0.0, 1.0]. Unset until its
+    // fact is recorded; zero is fully green and one is fully broken. Values are directly
+    // comparable: lower values are greener, and any value above zero is not green.
+    optional double repository_breakage_degree = 6;
+    // Whether every planned project has one durable result and completion is recorded. False
+    // until the project list is available.
+    bool project_results_complete = 7;
```
behinddwalls, line 119: what is this for? I am unable to relate what does it support?

api/stovepipe/proto/stovepipe.proto
```diff
@@ -54,11 +54,65 @@ message IngestResponse {
...
+    // VCS-agnostic commit URI validated by the request.
+    string change_uri = 3;
+    // Baseline URI for incremental validation. Empty for a full build.
+    string base_uri = 4;
+    // Stable public lifecycle state and terminal outcome of the request.
+    string request_state = 5;
+    // Breakage degree measures how broken the whole repository is on [0.0, 1.0]. Unset until its
+    // fact is recorded; zero is fully green and one is fully broken. Values are directly
+    // comparable: lower values are greener, and any value above zero is not green.
+    optional double repository_breakage_degree = 6;
```
behinddwalls, line 116: why is this optional? I assume if project is not provided we are retruning overall status plus paged projects

mnoah1: The API may be called before that result is recorded, so we need to distinguish between “no value” from “green" - so it is blank if the request is still in a state where there is no result yet.

api/stovepipe/proto/stovepipe.proto
mnoah1: I am thinking that we want to avoid the assumption that `projects` will always be totally fixed - and have the API explicitly indicate when it reaches that point.  For example we may want to gradually populate individual project results as they become available from the build and let those projects become green while the build or analysis is still finishing.  Or pre-populate the list of pending projects much earlier in the request, so callers could know what projects to expect as an outcome of this request.  I'm thinking that adding this indicator now gives us more flexibility to consider those cases in the future.

api/stovepipe/proto/stovepipe.proto
mnoah1: Renamed

api/stovepipe/proto/stovepipe.proto
```diff
@@ -54,11 +54,83 @@ message IngestResponse {
     string id = 1;
 }
 
+// GetProjectStatusByURIRequest selects the authoritative validation for an exact commit URI.
+message GetProjectStatusByURIRequest {
```
mnoah1: Split this into a `List` and a `Get`

api/stovepipe/proto/stovepipe.proto
```diff
@@ -54,11 +54,83 @@ message IngestResponse {
...
+    // Globally unique identifier of the authoritative request.
+    string request_id = 1;
+    // Logical queue containing the request.
+    string queue = 2;
+    // VCS-agnostic commit URI validated by the request.
+    string change_uri = 3;
+    // Baseline URI for incremental validation. Empty for a full build.
+    string base_uri = 4;
+    // Stable public lifecycle state and terminal outcome of the request.
+    RequestState request_state = 5;
```
mnoah1: Could take this out and use breakage as a sentinel

mnoah1:
Duplicate

PR #641, opened 2026-08-27
feat(errs): classify YARPC status errors

mnoah1:
## Summary
Intent:
- Retry transient YARPC failures instead of dead-lettering them as unknown errors.
- Preserve dependency attribution while distinguishing caller cancellation.

Changes:
- Classify typed YARPC statuses by retryability and origin.
- Register the classifier with Stovepipe and document the mapping.

### DIFF
--- a/platform/errs/yarpc/yarpc.go
+++ b/platform/errs/yarpc/yarpc.go
@@ -1,0 +1,76 @@
+// Copyright (c) 2026 Uber Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


platform/errs/yarpc/yarpc.go
```diff
@@ -0,0 +1,76 @@
...
+		yarpcerrors.CodeAlreadyExists,
+		yarpcerrors.CodePermissionDenied,
+		yarpcerrors.CodeFailedPrecondition,
+		yarpcerrors.CodeOutOfRange,
+		yarpcerrors.CodeUnimplemented,
+		yarpcerrors.CodeDataLoss,
+		yarpcerrors.CodeUnauthenticated:
+		return errs.InfraDependency
+
+	default:
```
behinddwalls, line 73: do we want to run the same generic classifiers plus other when we hit this? like any error should probably run through it?

behinddwalls: ohh i see this will get plugged into the classifier itself

behinddwalls: maybe worth adding how it will adding in RPC APIs?

mnoah1: Updated readme with a bit of additional detail here.

behinddwalls (approved):

PR #646, opened 2026-08-27
feat(stovepipe): Announce validation results

prathshenoy:
## Summary
**What**:
- Publish an event whenever a commit's validation result becomes durable, carrying whether the commit is healthy or broken and what it was measured against.
- Publish a distinct event when validation ends without a verdict, and fail the work when either event cannot be sent.

**Why**:
- Let subscriber systems act on trunk validation results as they land, instead of polling for them.
- Keep the event vendor-neutral so each deployment wires its own subscriber to forward results onward, instead of building any single environment's delivery path into the pipeline.

## Test Plan
- [x] Add unit tests.

## Revert Plan
- Revert this PR. The events are additive and nothing consumes them yet.

## Issues
- [CODEM-459](https://linear.app/uber/issue/CODEM-459/record-step-emit-events-internally-on-new-green)


stovepipe/controller/record/record.go
```diff
@@ -126,34 +156,29 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) er
...
 		}
-		if !holdsBookmark {
-			// A later green commit already holds the bookmark, so it also owns
-			// the promotion ref: promoting this older commit would move the ref
-			// backwards.
-			return nil
-		}
-		return c.promote(ctx, request)
+		return c.publishHookEvent(ctx, request, HookEventTypeValidationRecorded, map[string]any{
+			"queue":       request.Queue,
```
prathshenoy: Should this be a struct?

stovepipe/controller/record/record.go
```diff
@@ -63,12 +71,33 @@ const _opName = "record"
 // attribution that this stage does not do, so every fact it writes is whole-repository.
 const wholeRepositoryProject = ""
 
+// HookEventSource is the source reported on every hook event this stage publishes.
+// A consumer serving several domains matches on it to recognize stovepipe's events.
+const HookEventSource = "stovepipe"
```
mnoah1: Should we centralize the hook events in their own package?  We could have `stovepipe/core/hookevent` that defines stovepipe's different hook event types, and provides constructures that can be used to get the event:
```
package hookevent

const Source = "stovepipe"

type Type string

const (
	TypeValidationRepositoryRecorded  Type = "validation.repository.recorded"
	TypeValidationRepositoryCancelled Type = "validation.repository.cancelled"
)

NewValidationRepositoryRecorded(...) basehook.HookEvent
```

stovepipe/controller/record/record.go
```diff
@@ -452,6 +512,59 @@ func (c *Controller) promote(ctx context.Context, request entity.Request) error
...
+// Called last, after the fact write and after both caches derived from it have
+// moved: the payload carries no entity snapshot, so a hook reacting to a green
+// commit by reading the bookmark or resolving the promotion ref must not find
+// either still pointing at the previous commit.
+//
+// The message id is the event id, which is derived from the transition rather
+// than the clock, so a redelivery republishing the same event dedups into the
+// original message instead of enqueuing a second one. Partitioning by request id
+// matches the record topic's own, carrying per-request ordering across the seam.
+func (c *Controller) publishHookEvent(ctx context.Context, request entity.Request, eventType string, payload map[string]any) error {
```
mnoah1: As this is more like shared hook mechanics, could we put it in platform/hook/publisher.go as its own helper?  Something like:

```
func Publish(
    ctx context.Context,
    registry consumer.TopicRegistry,
    event *basehook.HookEvent,
    partitionKey string,
) error
```

Then between this and the other changes above, controllers could do something like:
```
event, err := hookevent.NewValidationRepositoryRecorded(request, fact, msg.ID)
if err != nil {
    return err
}

if err := platformhook.Publish(ctx, c.registry, event, request.ID); err != nil {
    // record-stage metrics and error context
}
```

stovepipe/controller/record/record.go
```diff
@@ -126,34 +156,29 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) er
...
-		if !holdsBookmark {
-			// A later green commit already holds the bookmark, so it also owns
-			// the promotion ref: promoting this older commit would move the ref
-			// backwards.
-			return nil
-		}
-		return c.promote(ctx, request)
+		return c.publishHookEvent(ctx, request, HookEventTypeValidationRecorded, map[string]any{
+			"queue":       request.Queue,
+			"request_id":  request.ID,
```
mnoah1: For Hooks we primarily just want to include "identity" fields (similar to the jumps between controller stages).  Then, the hooks themselves can look up info as needed.

mnoah1 (approved):

PR #651, opened 2026-08-31
fix(messagequeue): apply retry backoff to nacked messages

mnoah1:
## Summary
Message queue framework already provided for a retry backoff config, but reading/respecting it was not yet implemented.

Intent:
- Ensure explicit Nacks honor the existing retry backoff configuration instead of retrying on every poll cycle.
- Prevent transient downstream failures from immediately consuming the retry budget and dead-lettering messages.

Changes:
- Calculate each retry delay from the delivery attempt and configured initial delay, multiplier, and maximum.
- Persist the delayed visibility timestamp so backoff survives restarts and subscriber handoffs.
- Cover delay calculation, overflow handling, persistence, and redelivery behavior with unit and integration tests.

## Test Plan
- `./tool/bazel test //platform/extension/messagequeue/mysql:go_default_test //platform/consumer:go_default_test //platform/extension/messagequeue:go_default_test`
- `./tool/bazel build //test/integration/extension/messagequeue/mysql:go_default_test`

## Revert Plan
- Revert this PR to restore immediate redelivery after an explicit Nack.


platform/extension/messagequeue/mysql/delivery_state_store.go
```diff
@@ -146,16 +147,21 @@ func (s *sqldeliveryStateStore) MarkAcked(ctx context.Context, consumerGroup, to
...
 	op := metrics.Begin(s.scope, "mark_nacked", metrics.StorageLatencyBuckets,
 		metrics.NewTag("topic", topic),
 		metrics.NewTag("consumer_group", consumerGroup))
 	defer func() { op.Complete(retErr) }()
 
-	invisibleUntil := time.Now().UnixMilli()
+	nowMs := time.Now().UnixMilli()
+	invisibleUntil := nowMs
+	if delayMs > math.MaxInt64-nowMs {
+		invisibleUntil = math.MaxInt64
```
behinddwalls: should this ever happen or even if it happens should we honor it?

behinddwalls (commented):
**`delivery_state_store.go:160-161`** — The overflow guard `delayMs > math.MaxInt64 - nowMs` saturates `invisibleUntil` to `MaxInt64`, which means "never redeliver." The message isn't acked and isn't in the DLQ — it's silently stuck, and the partition watermark can never advance past it. For this to trigger, someone would need `InitialBackoffMs` near `MaxInt64` with no `MaxBackoffMs` cap — a misconfiguration, not a scenario to gracefully handle. Returning an error here instead would let the nack fail, the visibility timeout lapse, and the message redeliver normally. The same applies to the `retryBackoffMs` saturation in `subscriber.go:1531-1533` — clamping to `MaxInt64` creates the same invisible-forever problem. Consider returning an error (or clamping to `MaxBackoffMs` / a hard ceiling like 24h) so a config bug surfaces as a retryable failure rather than silent data loss.

platform/extension/messagequeue/mysql/delivery_state_store.go
behinddwalls: CLAUDE Generated:

It should never happen in practice — `delayMs` would need to be within ~1.7T of `MaxInt64` (roughly 292 million years), which means someone configured `InitialBackoffMs` near `MaxInt64` with no `MaxBackoffMs` cap. That's a misconfiguration, not a scenario to gracefully handle.

But no, we shouldn't honor it. Saturating `invisibleUntil` to `MaxInt64` means "never redeliver" — the message isn't acked and isn't in the DLQ, it's silently stuck, and the partition watermark can never advance past it. A hot loop would at least hit `MaxAttempts` and dead-letter; this just makes the message disappear with no signal.

Consider returning an error here instead — the nack fails, the visibility timeout lapses (~60s default), and the message redelivers normally. Same issue exists in `retryBackoffMs` (`subscriber.go:1531-1533`) where the float64 overflow clamps to `MaxInt64`. Both spots should either error or clamp to a survivable ceiling (e.g. `MaxBackoffMs` or 24h) so a config bug surfaces as a retryable failure rather than silent data loss.

mnoah1: Since this error mainly indicates a config issue, we can fix it there.  Updated this to validate the config when the subscriber starts as it should fail at the beginning, and also added the error message as fallback.

behinddwalls (approved):

PR #653, opened 2026-08-31
docs: align READMEs and RFCs with current code and AGENTS.md conventions

behinddwalls:
### Why?

A batch's merge used to wait for every dependency to resolve even when its passed builds already covered every way those dependencies could turn out. That made a small change sit behind a slow one for no reason: both the with- and without-dependency trees were validated, so whichever future arrives is already green.

### What?

Adds bypassablePath to the speculate controller: decide() merges a head once passed, unbroken, well-formed paths cover every 2^n combination of its unsettled dependencies. Settled dependencies still pin each path to reality through assumptionBroken. Finalize prefers a strict mergeable path and falls back to bypass coverage, superseding siblings either way.

Tightened isWellFormed to require dependencies in the head's canonical order, and read the coverage signature positionally off the slice (dropping the per-path map). Order is load-bearing: Base() projects the path positionally, so a permuted path is a different stack, and a reordered path can no longer masquerade as coverage for the canonical combination.

## Test Plan

- ./tool/bazel test //submitqueue/orchestrator/controller/speculate:go_default_test
- ./tool/bazel test //submitqueue/extension/speculation/... //submitqueue/entity:go_default_test //submitqueue/orchestrator/controller/build:go_default_test
- New TestBypassablePath "does not count a reordered path" and TestIsWellFormed "dependencies out of queue order" fail before, pass after.
- Four bypass E2E tests pass (BypassedHeadLandsFirst, BypassesMergingDependency, NoBypassWhenCoverageIsIncomplete, BypassesAnUnresolvedDependency).


behinddwalls:
Closing: this PR bundled two independent changes (the docs-alignment commit and the speculate bypass feature) under a misleading auto-taken title. Splitting into a stacked pair for clean review: docs commit first, bypass feature stacked on top.

PR #663, opened 2026-09-02
fix(gateway): align land strategy mapping with contract

sbalabanov:
## Summary
Intent:
- Preserve server-configured defaults instead of forcing rebase at the gateway boundary.
- Support every merge strategy exposed by the shared API contract.

Changes:
- Map DEFAULT to the domain sentinel so downstream Runway configuration selects the strategy.
- Map PROMOTE to its domain strategy while continuing to reject unknown numeric enum values.
- Extend mapper coverage for default and promote behavior.

---

<sub>Generated by the 🪄 pr-create skill.</sub>

### DIFF
--- a/service/submitqueue/gateway/server/mapper/land.go
+++ b/service/submitqueue/gateway/server/mapper/land.go
@@ -59,6 +59,8 @@ func resolveMergeStrategy(s mergestrategypb.Strategy) (mergestrategy.MergeStrate
 		return mergestrategy.MergeStrategySquashRebase, nil
 	case mergestrategypb.Strategy_MERGE:
 		return mergestrategy.MergeStrategyMerge, nil
+	case mergestrategypb.Strategy_PROMOTE:
+		return mergestrategy.MergeStrategyPromote, nil
 	default:
 		return mergestrategy.MergeStrategyUnknown, fmt.Errorf("%w: %v", errUnknownStrategy, s)
 	}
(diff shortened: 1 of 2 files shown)


service/submitqueue/gateway/server/mapper/land.go
```diff
@@ -51,14 +51,15 @@ func ProtoToLandRequest(req *pb.LandRequest) (entity.LandRequest, error) {
 func resolveMergeStrategy(s mergestrategypb.Strategy) (mergestrategy.MergeStrategy, error) {
 	switch s {
 	case mergestrategypb.Strategy_DEFAULT:
-		// TODO: resolve default strategy based on queue configuration
-		return mergestrategy.MergeStrategyRebase, nil
+		return mergestrategy.MergeStrategyUnknown, nil
```
sbalabanov: do not change that, restore Rebase and TODO

sbalabanov: Restored the existing DEFAULT-to-Rebase behavior and TODO in 7ec7230faf3d8f0bd52041bf1f66fa82387fb605; PROMOTE remains the only mapping change.

This preserves the current queue default behavior while adding support for the explicit PROMOTE contract value. The mapper test now covers PROMOTE without changing DEFAULT semantics.

[addressed by agent]

behinddwalls (approved):

PR #666, opened 2026-09-02
feat(stovepipe): add request logs for outcomes and lifecycle events

mnoah1:
## Summary
This PR builds on #665, which records Process-owned request states.

Intent:
- Retain terminal request outcomes and the durable build/fact milestones that explain them.
- Let queue redelivery repair any missing occurrence before downstream work continues.

Changes:
- Persist succeeded, failed, and cancelled request states with their build outcome reasons.
- Record build_triggered, build_finished, and validation_fact_recorded events with stable identities and bounded metadata.
- Keep each source write ahead of event materialization and materialization ahead of downstream publication or derived work.
- Wire the shared materializer into Build, BuildSignal, and Record.

## Test Plan
- Run the focused Build, BuildSignal, Record, request-log materializer, and server wiring Bazel tests.

## Revert Plan
- Revert this change to stop recording terminal outcomes and lifecycle events.

---


stovepipe/controller/build/build.go
```diff
@@ -158,6 +165,37 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) er
 	return nil
 }
 
+func (c *Controller) persistBuild(ctx context.Context, store storage.Storage, build entity.Build) error {
+	buildStore := store.GetBuildStore()
+	if err := buildStore.Create(ctx, build); err == nil {
+		return nil
+	} else if !errors.Is(err, storage.ErrAlreadyExists) {
```
sbalabanov: no need for else after return

stovepipe/controller/build/build.go
```diff
@@ -158,6 +165,37 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) er
...
+		return fmt.Errorf("failed to persist build %s: %w", build.ID, err)
+	}
+
```
sbalabanov: The whole workflow after this line deserves a quick explanation why we are doing it and what is a potential use case it is trying to validate

stovepipe/controller/build/build.go
```diff
@@ -158,6 +165,37 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) er
...
+	if err != nil {
+		return fmt.Errorf("failed to load existing build %s: %w", build.ID, err)
+	}
+	if stored.RequestID != build.RequestID {
+		return fmt.Errorf("build %s belongs to request %s, not %s", build.ID, stored.RequestID, build.RequestID)
+	}
+	return nil
+}
+
+func (c *Controller) persistBuildTriggered(ctx context.Context, store storage.Storage, request entity.Request, buildID string) error {
```
sbalabanov: the name of the function suggests it is somewhat similar to `persistBuild` but in fact it only writes to event log

sbalabanov (changes requested):

stovepipe/controller/build/build.go
mnoah1: Renamed

stovepipe/controller/build/build.go
mnoah1: Added comment with explanation.

stovepipe/controller/build/build.go
```diff
@@ -158,6 +165,43 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) er
...
+
+	// ErrAlreadyExists can mean an earlier delivery already stored this Request's Build,
+	// or another Request received the same runner-assigned build ID first. Only the first
+	// case is a safe retry; continuing after a collision would make BuildSignal update the
+	// wrong Request.
+	stored, err := buildStore.Get(ctx, build.ID)
+	if err != nil {
+		return fmt.Errorf("failed to load existing build %s: %w", build.ID, err)
+	}
+	if stored.RequestID != build.RequestID {
```
sbalabanov: In particular, to the previous question, I do not quite understand how the above is possible

sbalabanov (approved):

stovepipe/controller/build/build.go
mnoah1: I think I got a bit mixed up here - my thinking was that a retry that fails while triggering the build, could later re-run and find the expected build number now assigned to a different build.  However, given the ordering here, we actually don't store any number unless it's been assigned/returned by the build runner, so it can't get assigned to another request.

I think what this actually surfaces is that I may want to put an additional check before `Trigger` above to see if there is any existing build to reload, before triggering.  I'll remove this logic here for now, and look adding some more handling for that separately.

PR #669, opened 2026-09-02
feat(stovepipe): request history api - entities/mappers

mnoah1:
## Summary
Intent:
- Make retained request logs the authority for history availability.
- Establish the public and domain projections used by history reads.

Changes:
- Add history entity and mapper coverage and align the RFC with log-only lookup semantics.
- Return not found when no request log records are retained.
- Remove unused, undeployed history fields without reserving their names or tags.


api/stovepipe/proto/stovepipe.proto
```diff
@@ -83,10 +83,6 @@ message HistoryEvent {
         // Event that occurred without changing the request state.
         string event = 4;
     }
-    // Newer request that caused a superseded state. Empty when not applicable.
-    string superseded_by_request_id = 5;
-    // Build associated with the occurrence. Empty when no build applies.
-    string build_id = 6;
```
mnoah1, line 89: These haven't been used yet so deleting them here in backwards incompatible way.  They're already unused elsewhere in the code/db schema.

api/stovepipe/proto/stovepipe.proto
```diff
@@ -83,10 +83,6 @@ message HistoryEvent {
...
     // Stable domain reason for a terminal request outcome. Empty otherwise.
     string outcome_reason = 7;
```
sbalabanov: either change this ordinal to 5 (which will make it a breaking change) or reserve 5 and 6

sbalabanov (approved):

api/stovepipe/proto/stovepipe.proto
mnoah1: Updated outcome_reason to use field 5 in ab8eed39d7d809ee8658c8e382206454f2d0017f.

This contract has not been deployed, so reusing the first available ordinal keeps the schema compact without preserving unused field numbers 5 and 6.

[addressed by agent]

PR #670, opened 2026-09-02
feat(stovepipe): request history api - by request ID

mnoah1:
## Summary
Intent:
- Implement endpoint to return request history by request ID

Changes:
- Add the request-history controller and bounded selector validation.
- Read only the queue-scoped request log and preserve context metric tags.
- Cover successful, invalid, missing, and infrastructure outcomes.

PR #671, opened 2026-09-02
feat(stovepipe): request history api - by request URI

mnoah1:
## Summary
Intent:
- Support history lookup from an exact commit URI while preserving log-only authority.
- Keep the plural response model ready for a future multi-attempt URI index.

Changes:
- Resolve the current URI mapping and load its retained request log.
- Return one grouped history under the insert-once mapping contract.
- Cover selector validation, missing history, infrastructure errors, and metric tags.

## Stack
1. #669
1. #670
1. @ #671
1. #672

### DIFF
--- a/stovepipe/controller/read_errors.go
+++ b/stovepipe/controller/read_errors.go
@@ -44,10 +44,14 @@ func validateHistoryIdentifier(name, value string) error {
 // RequestHistoryNotFoundError indicates that no retained history exists for a selector.
 type RequestHistoryNotFoundError struct {
 	RequestID string
+	URI       string
 }
 
 // Error implements error.
 func (e *RequestHistoryNotFoundError) Error() string {
+	if e.RequestID == "" {
+		return fmt.Sprintf("request history not found for URI %q", e.URI)
+	}
 	return fmt.Sprintf("request history not found for request ID %q", e.RequestID)
 }
 
--- a/stovepipe/controller/read_errors.go
+++ b/stovepipe/controller/read_errors.go
@@ -41,22 +41,34 @@ func validateHistoryIdentifier(name, value string) error {
 	return nil
 }
 
-// RequestHistoryNotFoundError indicates that no retained history exists for a selector.
-type RequestHistoryNotFoundError struct {
+// RequestHistoryByIDNotFoundError indicates that no retained history exists for a request ID.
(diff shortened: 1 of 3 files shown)


stovepipe/controller/request_history.go
```diff
@@ -78,13 +79,65 @@ func (c *requestHistoryController) readHistoryByID(ctx context.Context, req enti
...
+		return entity.RequestHistory{}, fmt.Errorf("GetRequestHistoryByURI failed to list request logs uri=%s request_id=%s: %w", req.URI, requestID, err)
+	}
+
+	return entity.RequestHistory{RequestID: requestID, Events: logs}, nil
+}
+
+func loadRequestLogs(ctx context.Context, store storage.RequestLogStore, requestID string, notFound *RequestHistoryNotFoundError) ([]entity.RequestLog, error) {
+	logs, err := store.List(ctx, requestID)
+	if storage.IsNotFound(err) {
+		return nil, errs.NewUserError(notFound)
```
sbalabanov: why is it a user error? If it finds requestID by URI, should the data down the pipe be consistent and if it is not, that's not a user's fault?

stovepipe/controller/read_errors.go
```diff
@@ -44,10 +44,14 @@ func validateHistoryIdentifier(name, value string) error {
 // RequestHistoryNotFoundError indicates that no retained history exists for a selector.
 type RequestHistoryNotFoundError struct {
 	RequestID string
+	URI       string
```
sbalabanov: Having expectations on how fields are set in the same struct could be error-prone. You may consider other designs:
1. Specific error type per each operation (one will have request id and the other uri as a single field)
2. - id (string)
    - type (enum: requestID, url)

sbalabanov (changes requested):

stovepipe/controller/request_history.go
mnoah1: Updated this - data should be consistent but the first event gets created just after URI creation, so a failure in between inserting those is possible but not the user's fault.

stovepipe/controller/read_errors.go
mnoah1: Updated to separate types

sbalabanov (approved):

PR #675, opened 2026-09-04
fix(runway): ISS-001 reconcile stale merge retries

sbalabanov:
## Summary
Intent:
- Let at-least-once merge deliveries converge when a prior attempt already updated Git but failed to publish its result.
- Preserve staleness protection for superseded requests that would still modify the target.

Changes:
- Apply transforming merge strategies locally before staleness validation and return success when the target already satisfies every step.
- Check PROMOTE containment before freshness and account for head refs moved by a prior contention attempt.
- Add real-Git fail-once end-to-end coverage, strategy-level regression tests, and adjacent reproduction documentation.

---

<sub>Generated by the 🪄 pr-create skill in devexp-agent-marketplace</sub>

## Test Plan

| Validator | Status | Issues |
|----------|--------|--------|
| android-lint | not_applicable | 0 |
| java-coverage | not_applicable | 0 |
| ios-lint | not_applicable | 0 |
| web-repocheck | not_applicable | 0 |
| java-lint | not_applicable | 0 |
| go-coverage | not_applicable | 0 |
| go-gazelle | not_applicable | 0 |
| arc-lint | not_applicable | 0 |
| visual-ios | not_applicable | 0 |
| visual-android | not_applicable | 0 |
| go-proto-lint | not_applicable | 0 |
| android-coverage | not_applicable | 0 |
| arc-unit | not_applicable | 0 |
| go-thrift-lint | not_applicable | 0 |
| ios-test | not_applicable | 0 |
| uber-one | not_applicable | 0 |
| web-lint | not_applicable | 0 |
| web-coverage | not_applicable | 0 |
| diff-template | not_applicable | 0 |
| fix-disclosure | not_applicable | 0 |
| go-lint | not_applicable | 0 |
| web-typecheck | not_applicable | 0 |
| web-unit | not_applicable | 0 |
| merge-conflict | not_applicable | 0 |
| visual-web | not_applicable | 0 |
| ureview | completed | 0 |
| custom | not_applicable | 0 |

**0** issues detected

<sub>Skipped validators: claude · [EngWiki](http://t.uber.com/ai-verification)</sub>

<details>
<summary>Prior runs</summary>

**Run at `c9ee7a8` on Sep 4 22:04 UTC** · 5 files · 2s · 0 issues detected

</details>

### DIFF
--- a/runway/extension/merger/git/git_merger.go
+++ b/runway/extension/merger/git/git_merger.go
@@ -420,16 +420,19 @@ func (o *providerCheck) check(ref changeRef, stepID string) error {
 // strategies (REBASE, SQUASH_REBASE, MERGE), retrying on remote contention when
 // committing. For a dry run it applies the steps locally then discards them.
 func (m *gitMerger) applyTransforming(ctx context.Context, req *runwaymq.MergeRequest, steps []resolvedStep, commit bool) (*runwaymq.MergeResult, error) {
-	// Fetch and vet every change the request names before the first attempt, so
-	// an unusable request fails without having touched the checkout. This sits
-	// outside the retry loop deliberately: the refs are fixed for the whole
-	// request, and re-checking them per attempt would re-query the remote to
-	// learn what it already told us.
+	// Fetch every change the request names before the first attempt, so a
+	// request naming an unreachable commit fails without touching the checkout.
+	//
+	// Freshness is not checked here. It cannot be: a redelivery of work that
+	// already landed looks stale — its head branch has moved to the commit the
+	// merger made of it — and rejecting it would report an error for a change
+	// that is sitting in the target. The check moves to after local application,
+	// where "did this actually change anything" is answerable. The cost is that
+	// a genuinely superseded request now runs a full fetch, reset --hard,
+	// clean -fdx and every cherry-pick before it is rejected; only the local
+	// checkout is touched, and no ref is written.
 	refs := stepChangeRefs(steps)
 	if err := m.ensureObjects(ctx, refs); err != nil {
-		return nil, err
-	}
-	if err := m.checkStale(ctx, refs); err != nil {
 		return nil, err
 	}
 
@@ -439,34 +442,20 @@ func (m *gitMerger) applyTransforming(ctx context.Context, req *runwaymq.MergeRe
 	// the branch is moved on rather than stranded on a commit that never landed.
 	tracked := make(headBranchTracker)
 	for attempt := 1; attempt <= m.maxPushAttempts; attempt++ {
-		baseSHA, stepResults, err := m.tryApply(ctx, steps, commit, tracked)
+		result, baseSHA, err := m.attemptTransforming(ctx, req, steps, refs, commit, tracked)
 		if err == nil {
-			if !commit {
-				// Discard the local commits the dry run created so the checkout
-				// is clean for the next operation, and report empty Outputs.
-				if derr := m.resetToRemote(ctx); derr != nil {
-					return nil, fmt.Errorf("discard after dry-run: %w", derr)
-				}
-				stripOutputs(stepResults)
-			}
 			m.logger.Debugw("merge complete", "id", req.GetId(), "target", m.target, "commit", commit)
-			return successResult(req, stepResults), nil
+			return result, nil
 		}
-
-		// A conflict is terminal — no retry. Discard any partial dry-run state.
-		if errors.Is(err, merger.ErrConflict) {
-			if !commit {
-				_ = m.resetToRemote(ctx)
-			}
+		// A conflict or a superseded request is the answer, not something to
+		// retry. baseSHA is empty when the attempt failed before reset captured
+		// a base, which leaves nothing to compare the tip against.
+		if merger.IsTerminal(err) || !commit || baseSHA == "" {
 			return nil, err
 		}
 
-		// Only a push failure caused by the remote tip moving under us (between
-		// reset and push) is worth retrying; everything else is fatal. baseSHA
-		// is empty when the failure happened before reset captured a base.
-		if !commit || baseSHA == "" {
-			return nil, err
-		}
+		// Only a failure caused by the remote tip moving under us (between reset
+		// and push) is worth retrying; everything else is fatal.
 		currentSHA, refetchErr := m.refetchTipSHA(ctx)
 		if refetchErr != nil {
 			return nil, fmt.Errorf("refetch after push failure failed: %v (original push error: %w)", refetchErr, err)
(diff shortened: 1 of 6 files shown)


runway/extension/merger/git/git_merger.go
```diff
@@ -420,53 +420,70 @@ func (o *providerCheck) check(ref changeRef, stepID string) error {
...
 
 		// Only a push failure caused by the remote tip moving under us (between
 		// reset and push) is worth retrying; everything else is fatal. baseSHA
 		// is empty when the failure happened before reset captured a base.
-		if !commit || baseSHA == "" {
-			return nil, err
-		}
 		currentSHA, refetchErr := m.refetchTipSHA(ctx)
 		if refetchErr != nil {
 			return nil, fmt.Errorf("refetch after push failure failed: %v (original push error: %w)", refetchErr, err)
```
sbalabanov, line 461: **A cleaner shape is possible.** Three things make this hard to follow:

1. The `if / else if / else` chain has three arms, two of which fall through to the retry code at the bottom and one of which always returns. You have to trace each arm to know which.
2. `err` is reused. Inside the last arm it stops meaning "apply failed" and starts meaning "head-branch or push failed" (`err = m.updateHeadBranches(...)`, `err = pushErr`). Both then land on the same `refetchTipSHA` block.
3. `checkStale` is called twice, with the same two lines.

Pulling one attempt into its own function fixes all three. The loop keeps only "try, succeed, or decide whether to retry":

```go
for attempt := 1; attempt <= m.maxPushAttempts; attempt++ {
	result, baseSHA, err := m.attemptTransforming(ctx, req, steps, refs, commit, tracked)
	if err == nil {
		m.logger.Debugw("merge complete", ...)
		return result, nil
	}
	// A conflict or a stale request is the answer, not a failure to retry.
	if merger.IsTerminal(err) || !commit || baseSHA == "" {
		return nil, err
	}
	... existing refetchTipSHA contention check, unchanged ...
}
```

and the attempt reads top to bottom with one early return per outcome:

```go
func (m *gitMerger) attemptTransforming(...) (*runwaymq.MergeResult, string, error) {
	baseSHA, stepResults, heads, err := m.tryApply(ctx, steps)

	// Nothing to apply: the target already satisfies every step.
	if err == nil && !stepResultsHaveOutputs(stepResults) {
		return successResult(req, stepResults), baseSHA, nil
	}
	// Everything past here would change the target, so it must be current.
	if staleErr := m.checkStale(ctx, refs, tracked); staleErr != nil {
		return nil, "", staleErr
	}
	if err != nil {
		return nil, baseSHA, err
	}
	if !commit {
		... discard + stripOutputs ...
		return successResult(req, stepResults), baseSHA, nil
	}
	if m.updateHeadBranch {
		if err := m.updateHeadBranches(ctx, heads, tracked); err != nil {
			return nil, baseSHA, err
		}
	}
	if err := m.push(ctx); err != nil {
		coremetrics.NamedCounter(m.metricsScope, "merge", "git_push_errors", 1)
		return nil, baseSHA, err
	}
	return successResult(req, stepResults), baseSHA, nil
}
```

`merger.IsTerminal` covers the stale case and the conflict case together, so the explicit `errors.Is(err, merger.ErrConflict)` branch goes away too.

I built this on your head commit and ran the package: `//runway/extension/merger/git:go_default_test` **passes unchanged**, including the new `TestMerge_StaleAlreadySatisfiedSucceeds` and `TestMerge_StalePendingChangeRejected`. Measured after:

| | cyclomatic | max nesting |
|---|---|---|
| `applyTransforming` | 9 | 2 |
| `attemptTransforming` | 10 | 2 |

Down from 18/5, and both below the file's previous worst.

<sub>🤖 [posted by agent] — automated review by Claude Code, requested by @sbalabanov.</sub>

runway/extension/merger/git/git_merger.go
```diff
@@ -420,53 +420,70 @@ func (o *providerCheck) check(ref changeRef, stepID string) error {
...
+				return nil, staleErr
+			}
+
+			// A conflict is terminal — no retry. The failing apply function
+			// already discarded its in-progress Git operation.
+			if errors.Is(err, merger.ErrConflict) {
+				return nil, err
+			}
+			if !commit || baseSHA == "" {
+				return nil, err
```
sbalabanov: The dry-run reset was dropped here, and the new comment is not quite right.

The apply function aborts **its own** step (`merge --abort` / `cherry-pick --abort`), but in a multi-step request the earlier steps have already committed to the local checkout. Those commits stay.

I think it is harmless — every entry point starts with `resetToRemote`, so the next caller cleans up — but the comment reads as "nothing is left behind", which is not the case. Suggest saying the checkout is left dirty on purpose and the next attempt resets it.

<sub>🤖 [posted by agent] — automated review by Claude Code, requested by @sbalabanov.</sub>

runway/extension/merger/git/git_merger.go
```diff
@@ -420,53 +420,70 @@ func (o *providerCheck) check(ref changeRef, stepID string) error {
...
 // committing. For a dry run it applies the steps locally then discards them.
 func (m *gitMerger) applyTransforming(ctx context.Context, req *runwaymq.MergeRequest, steps []resolvedStep, commit bool) (*runwaymq.MergeResult, error) {
-	// Fetch and vet every change the request names before the first attempt, so
-	// an unusable request fails without having touched the checkout. This sits
-	// outside the retry loop deliberately: the refs are fixed for the whole
-	// request, and re-checking them per attempt would re-query the remote to
-	// learn what it already told us.
+	// Fetch every change the request names before the first attempt. Freshness
+	// is checked after local application, when the merger knows whether the
+	// target already satisfies the request.
```
sbalabanov: Worth saying out loud in this comment: a stale request now does a full `fetch` + `reset --hard` + `clean -fdx` + all the cherry-picks *before* it is rejected.

The old comment made a point of the opposite ("so an unusable request fails without having touched the checkout"). Trading that away is what makes the fix work, so it is the right call — but the sentence explaining the old ordering was deleted rather than replaced, and the new cost is invisible to the next reader.

<sub>🤖 [posted by agent] — automated review by Claude Code, requested by @sbalabanov.</sub>

runway/extension/merger/git/git_merger.go
```diff
@@ -748,6 +743,10 @@ func (m *gitMerger) promote(ctx context.Context, req *runwaymq.MergeRequest, rs
 			return promoteResult(req, rs, sha, commit), nil
 		}
 
+		if err := m.checkStale(ctx, []changeRef{ref}, nil); err != nil {
```
sbalabanov: `checkStale` now runs inside the retry loop, so it re-queries the remote (`ls-remote`) on every contention attempt. The comment you removed said it sat outside the loop deliberately, because the commit being promoted is fixed for the whole request and only the tip moves.

That reasoning still holds — the check has to move *after* containment, but it does not have to move *inside* the loop. Running it once, the first time containment misses, keeps both properties.

Minor; only matters when `maxPushAttempts` is being used.

<sub>🤖 [posted by agent] — automated review by Claude Code, requested by @sbalabanov.</sub>

runway/extension/merger/git/objects.go
```diff
@@ -110,6 +110,10 @@ func (m *gitMerger) checkStale(ctx context.Context, refs []changeRef) error {
 			continue
 		}
 		if current := fields[0]; current != ref.SHA {
+			branch, movedTo, moved := tracked.lookup(ref.SHA)
+			if moved && branch == ref.Ref && movedTo == current {
+				continue
+			}
```
sbalabanov: This only forgives a head branch that **this invocation** moved, since `tracked` starts empty on every delivery.

So the case where the process dies after `updateHeadBranches` pushed but before the target push is still rejected as stale on redelivery: the head branch is at `A'`, the request still pins `A`, the target never moved, so the apply produces outputs and this check fires.

Not a regression — it failed the same way before. But it is the same shape as ISS-001 ("a prior attempt already updated Git"), so is it deliberately out of scope, or just not hit by the e2e test because the failure there is injected after both pushes?

<sub>🤖 [posted by agent] — automated review by Claude Code, requested by @sbalabanov.</sub>

sbalabanov (commented):
Reviewed with the same lens as #678. The fix itself looks right: applying first and checking freshness only when the apply would actually change the target is the correct way to make a redelivery converge, and the terminal-vs-retryable split still holds (`checkStale` returns `ErrInvalidRequest`, the controller turns it into FAILED and acks). The e2e fail-once trigger is a genuinely good regression test.

My main comment is the one you flagged: the loop got much harder to read. Measured on `git_merger.go`:

| | cyclomatic | max nesting |
|---|---|---|
| `applyTransforming` before | 13 | 4 |
| `applyTransforming` after | **18** | **5** |

It is now the most complex function in the file (next is `resolveAndValidate` at 14). I tried a smaller shape, and it works — details inline.

Four smaller notes below.

<sub>🤖 [posted by agent] — automated review by Claude Code, requested by @sbalabanov.</sub>

runway/extension/merger/git/git_merger.go
sbalabanov: address

runway/extension/merger/git/git_merger.go
sbalabanov: ignore

runway/extension/merger/git/git_merger.go
sbalabanov: address

runway/extension/merger/git/git_merger.go
sbalabanov: address

runway/extension/merger/git/objects.go
sbalabanov: address, and update e2e test

test/e2e/submitqueue/git_suite_test.go
```diff
@@ -286,6 +320,87 @@ func (s *GitMergeSuite) requireStatus(sqid string, want entity.RequestStatus) {
...
+		for _, delivery := range parked {
+			if delivery.Topic == runwaymq.TopicKeyMerge.String() && delivery.MessageID == batchID {
+				return true
+			}
+		}
+		return false
+	})
+}
+
+func (s *GitMergeSuite) installMergeSignalFailOnce(messageID string) {
```
sbalabanov, line 398: Will it work only on mysql databases with trigger support? Add a TODO comment to enhance e2e test infra to allow inside-the-controller breakpoints, if that's what this new function is trying to achieve.

runway/extension/merger/git/git_merger.go
sbalabanov: Done in `8dae6f14`. The cycle moved into `attemptTransforming`, so the loop holds only "try, succeed, or decide whether to retry". The three-arm chain, the reused `err`, and the duplicate `checkStale` are all gone, and `merger.IsTerminal` now covers conflict and stale together.

`applyTransforming` 18 → **9** cyclomatic, 5 → **2** nesting. `attemptTransforming` is 10/2. Package tests pass.

<sub>🤖 [addressed by agent] — Claude Code, at @sbalabanov's request.</sub>

runway/extension/merger/git/git_merger.go
sbalabanov: Ignored as instructed — the comment and the missing dry-run reset are unchanged.

<sub>🤖 [addressed by agent] — Claude Code, at @sbalabanov's request.</sub>

runway/extension/merger/git/git_merger.go
sbalabanov: Done. The comment now says what the new ordering costs: a superseded request runs a full `fetch`, `reset --hard`, `clean -fdx` and every cherry-pick before it is rejected, and that only the local checkout is touched — no ref is written. It also says why the check cannot run earlier.

<sub>🤖 [addressed by agent] — Claude Code, at @sbalabanov's request.</sub>

runway/extension/merger/git/git_merger.go
sbalabanov: Done. `promote` keeps a `freshnessChecked` flag and runs `checkStale` at most once, the first time containment misses. No repeated `ls-remote` per contention attempt.

Costs one branch (`promote` 13 → 14 cyclomatic). I kept the explicit flag rather than a trick like emptying the ref slice, since the flag says what it means.

<sub>🤖 [addressed by agent] — Claude Code, at @sbalabanov's request.</sub>

runway/extension/merger/git/objects.go
sbalabanov: Done, and it needed more than the tracker.

The branch is now recognised by **who committed the commit it sits on**. The merger injects its committer identity into every git command it runs, so a head branch on a commit with that identity was moved by an earlier attempt — including an earlier *delivery*, which is what the tracker could never cover, because it lives only as long as one request. A branch the author advanced carries their identity and is still rejected.

That let the `tracked` parameter come back off `checkStale` entirely, so the signature change in this PR is reverted: one mechanism instead of two, and the durable one.

**The e2e turned up a second half to the bug.** With only the staleness fix, the change lands but the head branch stays stranded on the commit the lost delivery made: `updateHeadBranchFor` looks for a branch sitting at the pinned SHA, finds none, and skips. That leaves exactly the contradictory state ISS-001 is about — landed in Git, not merged as far as a provider can tell. Fixed with the same signal in `headBranchMovedByMerger`, so the branch is moved on instead of skipped. Shout if you would rather that were a separate PR.

Tests:
- `TestMerge_HeadBranchMovedByEarlierDelivery_IsNotStale` — verified it **fails** without the fix and passes with it.
- `TestMerge_HeadBranchMovedByAuthor_IsStale` — guards the other direction; passes either way.
- `TestLand_HeadBranchMovedByLostDelivery_StillLands` in the git e2e, plus a section in `ISS-001.md`. All 7 e2e tests pass.

The e2e sets the ref state up directly rather than killing Runway mid-push. Worth knowing why: on this branch a rejected push is not retryable, so driving it through a real failed push dead-letters instead of redelivering. What the redelivery actually sees is the refs, so that is what the test arranges.

<sub>🤖 [addressed by agent] — Claude Code, at @sbalabanov's request.</sub>

test/e2e/submitqueue/git_suite_test.go
sbalabanov: Yes — MySQL only. It needs trigger support, a non-transactional table so the counters survive the rolled-back insert, and `SIGNAL` to raise an error the publisher retries. A queue on any other backend could not run this test.

Added a doc comment saying so, and the TODO:

```go
// TODO(e2e): give the harness a backend-independent way to fail a controller
// partway — a fault-injection seam the consumer honours — so a test like this
// stops depending on the queue's storage engine.
```

That is what the helper is reaching for: there is currently no way to stop a controller between two of its own steps from outside the process.

<sub>🤖 [addressed by agent] — Claude Code, at @sbalabanov's request.</sub>

test/e2e/submitqueue/ISS-001.md
```diff
@@ -0,0 +1,73 @@
+# ISS-001: stale retry after a successful Git push
```
behinddwalls, line 1: is this needed in source?

behinddwalls (approved):

test/e2e/submitqueue/ISS-001.md
sbalabanov: I've decided to complement an E2E test with a test case study, for future references. I am not positive yet on the best way to store these studies and relate them from these regression tests.

sbalabanov:
Cancelled in favor of https://github.com/uber/submitqueue/pull/723

PR #676, opened 2026-09-04
fix(runway): ISS-010 stop logging queue credentials

sbalabanov:
### Intent

- Prevent `QUEUE_MYSQL_DSN` credentials and query parameters from appearing in Runway startup logs.
- Preserve a useful queue-initialization signal without rendering connection configuration.

### Changes

- Log the queue backend instead of the raw DSN.
- Add observer-based regression coverage with sentinel username, password, and query values.

### Reproduction

Before the fix, starting Runway with `QUEUE_MYSQL_DSN=user:super-secret@tcp(mysql:3306)/submitqueue?parseTime=true` produced `initialized queue {"dsn":"user:super-secret@tcp(mysql:3306)/submitqueue?parseTime=true"}`.

After the fix, the same configuration produces `initialized queue {"backend":"mysql"}` with no DSN components.

---

### DIFF
--- a/service/runway/server/main.go
+++ b/service/runway/server/main.go
@@ -146,7 +146,7 @@ func run() error {
 	}
 	defer mysqlQueue.Close()
 
-	logger.Info("initialized queue", zap.String("dsn", queueDSN))
+	logQueueInitialized(logger)
 
 	subscriberName := os.Getenv("HOSTNAME")
 	if subscriberName == "" {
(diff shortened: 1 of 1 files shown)


service/runway/server/main_test.go
```diff
@@ -0,0 +1,80 @@
+// Copyright (c) 2026 Uber Technologies, Inc.
```
sbalabanov: no need for tests for this change, remove test file from the change

sbalabanov: Removed the test file and its BUILD entries in cd36fb21e0d8ba4c00c9aaa6144499c102c74641, keeping the change focused on removing the credential-bearing log field.

[addressed by agent]

behinddwalls (approved):

PR #681, opened 2026-09-07
feat(messagequeue): shard MySQL queue tables by tenant

behinddwalls:
### Why?

The MySQL message queue keyed every table on topic and partition key. That cannot serve as a Vitess vindex because SubmitQueue uses partition keys for ordering units such as request and build IDs; one business queue would therefore scatter across shards and share shard capacity with unrelated queues.

### What?

- Add a tenant shard identity to every MySQL MQ table, primary key, store query, message, and admin operation while preserving partition_key as the ordering unit.
- Use readable ASCII VARCHAR fields for tenant, topic, and consumer identifiers and explicit utf8mb4 VARCHAR fields for partition keys and message IDs, keeping composite indexes under InnoDB's 3072-byte limit.
- Publish through one Message function with named MessageParams fields; tenant is required and mirrored into queue_name delivery metadata.
- Discover partitions from an authoritative MQ_TENANTS list parsed by shared service wiring; isolate discovery, lease, heartbeat, and shutdown failures per tenant.
- Map SubmitQueue queue names to tenants at publish/wiring boundaries and reject Stovepipe ingest for unconfigured tenants before side effects.
- Add schema checks, shard-linter coverage, and cross-tenant ACK/GC and DLQ isolation tests.

## Test Plan

✅ Platform publish and hook tests
✅ 28 focused controller and wiring test suites
✅ `./tool/bazel test //platform/extension/messagequeue/mysql/... //tool/linter/queueshard/... --test_output=errors`
✅ `./tool/bazel test //test/integration/extension/messagequeue/mysql:go_default_test --test_output=errors`
✅ `./tool/bazel build //...`
✅ `make fmt && make gazelle`


platform/extension/messagequeue/mysql/tenants.go
```diff
@@ -0,0 +1,46 @@
...
+	if len(tenants) == 0 {
+		return nil, fmt.Errorf("MQ_TENANTS must contain at least one tenant")
+	}
+	for _, tenant := range tenants {
+		if err := validateASCIIIdentifier("tenant", tenant); err != nil {
+			return nil, fmt.Errorf("invalid MQ_TENANTS: %w", err)
+		}
+	}
+	return tenants, nil
+}
```
behinddwalls: it only seems to be used in service/examples? why does it need to be in main impl folder? why not just sit in service/submitqueue... path?

platform/publish/publish.go
```diff
@@ -68,6 +68,11 @@ func MessageWithMetadata(ctx context.Context, registry consumer.TopicRegistry, k
 	}
 
 	msg := entityqueue.NewMessage(msgID, payload, partitionKey, metadataFromContext(ctx, metadata))
+	if tenant := msg.Metadata[entityqueue.MetadataKeyQueueName]; tenant != "" {
+		msg.Tenant = tenant
+	} else if queueName, ok := entityqueue.QueueName(ctx); ok {
+		msg.Tenant = queueName
+	}
```
behinddwalls: is this the right way? it hides the first level concept?

platform/extension/messagequeue/mysql/tenants.go
behinddwalls: Done in 6a6bf07d. Moved MQ_TENANTS parsing out of the MySQL implementation into service/messagequeue. It is shared service wiring rather than submitqueue-specific because Runway and Stovepipe use the same configuration.

platform/publish/publish.go
behinddwalls: Done in 6a6bf07d. Tenant is now a required first-class argument to publish.Message, MessageWithMetadata, and hook.Publish. Every producer passes its authoritative queue separately from the partition key; publish mirrors it into queue_name metadata and rejects conflicting metadata.

platform/publish/publish.go
```diff
@@ -48,16 +48,18 @@ import (
 // this particular message exists for. A retry of the same cause then dedups,
 // which is what makes redelivery safe, while a new cause about the same entity
 // can never be swallowed by an older row.
-func Message(ctx context.Context, registry consumer.TopicRegistry, key consumer.TopicKey, msgID string, payload []byte, partitionKey string) error {
-	return MessageWithMetadata(ctx, registry, key, msgID, payload, partitionKey, nil)
+func Message(ctx context.Context, registry consumer.TopicRegistry, key consumer.TopicKey, tenant, msgID string, payload []byte, partitionKey string) error {
```
behinddwalls: should we use some sort of Object/struct for args? so that we don't have to change again in future and also we can probably just do one method instead of 2?

behinddwalls: Done in cfce51e8. Replaced Message and MessageWithMetadata with one Message function taking a documented MessageParams struct. All producers now use named fields, including optional Metadata, so future attributes can be added without another positional API migration.

PR #682, opened 2026-09-07
docs(submitqueue): combine predictor into Scorer

behinddwalls:
### Why?

Ranking only needs one number per unresolved batch. A sibling Predictor factory duplicated the per-queue seam and made evidence look like a second public extension. The review asked for one ranking interface, framed as a logit-linear model.

### What?

Scorer is the ranking extension: `Score(ctx, batch, paths)`. Evidence is the default impl wrapping a nested base (heuristic or composite). YAML is `type: evidence` plus `factors` plus `base`. The revision is `logit(p') = logit(p_base) + Σ w_i x_i` with configured weights (`w_i = log(factor_i)`), not a fitted GLM.

The RFC file is `doc/rfc/submitqueue/outcome-scorer.md`. The evidence factor for the land-stage batch is `landing`, matching `BatchStateLanding`.

## Test Plan

Docs-only change.

## Stack

- https://github.com/uber/submitqueue/pull/682 ⬅️
- https://github.com/uber/submitqueue/pull/683
- https://github.com/uber/submitqueue/pull/684
- https://github.com/uber/submitqueue/pull/686


sbalabanov (changes requested):
1. Is predictor a first class abstraction (i.e. an interface) or implementation detail of the generator?
2. Let's discuss as an alternative to have all the decisions inside scorer. Scorer is called for each batch anyways? Why can't be made evidence-aware?
3. The proposed implementation of the predictor is somewhat hard to read. Can we frame it as a linear regression model with defined features and configured weights? What in the proposal does not fit into a definition of LGM?

behinddwalls:
Agreed on all three — the stack now does (1)+(2) as one Scorer, and (3) is documented on that impl.

1. **Scorer is the first-class ranking extension.** Evidence is a scorer implementation wrapping a nested `base`, not a Generator private type and not a sibling Factory. `bestfirst` does not import a Predictor.

2. **One `Score` call.** The contract is `Score(ctx, batch, paths)`. Heuristic/composite ignore `paths`; evidence revises `p_base` from the path set and batch state. YAML is `type: evidence` + `factors` + `base` (heuristic/composite only under `base`).

3. **Logit-linear / GLM with configured weights:** `logit(p') = logit(p_base) + Σ w_i x_i`, `w_i = log(factor_i)`. Neutral `1` is `w = 0`. Weights are YAML, not trained; the epsilon clamp is numeric, not part of the linear predictor.

The rewrite is in this PR; impl / wiring / demo are #683, #684, #686.

doc/rfc/submitqueue/outcome-predictor.md
```diff
@@ -0,0 +1,126 @@
+# Outcome Scorer
```
sbalabanov, line 1: Predictor is gone, file name still specifies it

sbalabanov (approved):

doc/rfc/submitqueue/outcome-scorer.md
```diff
@@ -0,0 +1,126 @@
+# Outcome Scorer
```
behinddwalls: Renamed to `doc/rfc/submitqueue/outcome-scorer.md` and updated the RFC index / speculation RFC links. Rebased the stack onto current `main` as well, so the evidence factor is `landing` to match `BatchStateLanding`.

PR #698, opened 2026-09-08
docs: add Web/UI Design RFC

behinddwalls:
### Why?

Operators have no browser UI, and the frontend should support both the initial read-only console and richer interactions without replacing its stack.

### What?

Defines per-domain Next.js/React BFFs with shared Node hooks, OSS example hosts, and production overlays. Every data RPC forwards the end-user identity; the gateway authorizes before returning any data. Node is pinned to 24.x. Rejected alternatives are templ/HTMX, a React SPA with a Go API host, Fusion-only rewrite, and direct grpc-web. Ping is not part of the first UI.

### DIFF
--- a/doc/rfc/web.md
+++ b/doc/rfc/web.md
@@ -1,0 +1,175 @@
+# Web/UI Design
+
+## Problem
+
+Operators have no browser UI. The CLI polls gateway history. Gateway already has summary, history, and receipt-time `List` ([status-list-api.md](submitqueue/status-list-api.md), [history-api.md](submitqueue/history-api.md)); Stovepipe has a parallel history API ([stovepipe/request-history-api.md](stovepipe/request-history-api.md)). Those RPCs are gRPC. A browser cannot call them the way the CLI does.
+
+The first UI is a **read** operator console: pick a queue, list requests, open summary and history, search by change URI, poll while a request is in flight. That is the same surface as `submitqueue/client` watch, not a workflow editor or a Cadence-scale SPA.
+
+Two ways to serve that HTML:
+
+1. **Go templates** — an importable Go library plus a Go HTTP host (OSS example under `service/`, production in go-code). Same library/host split as gateway.
+2. **Next.js BFF** — a Node process in OSS that the browser talks HTTP/JSON to; that process calls gRPC. Production is a source overlay (Cadence Web v4).
+
+SubmitQueue picks (1). The rest of this RFC is that design. Next.js stays the documented alternative if the UX outgrows pages, forms, fragments, and polling.
+
+## Choice
+
+Lightweight operator reads fit server-rendered pages. The product and Uber wiring are already Go. A Node BFF would add npm, a second observability stack, and an overlay repo that cannot import `{domain}/web` the way go-code imports the gateway.
+
+Cadence Web v4 is still the closest **rich** analog (Next.js, React, Base Web). This UX is pages, forms, fragments, and polling: Go + [`templ`](https://templ.guide/) HTML, HTMX for updates, server as source of truth.
+
+## Picture
+
+```
+Browser
+   |  HTTP: pages, forms, HTMX fragments
+   v
+Go web host
+   |  imports submitqueue/web
+   |  owns HTTP, auth, zap, tally, tracing, gateway client
+   v
+SubmitQueue gateway RPC
+   |  authenticates and authorizes every operation
+   v
+Gateway controllers and stores
+
+
+This repository                         Uber go-code
+---------------                         ------------
+submitqueue/web       UI library <----  imported by web host
+service/submitqueue/web example host    HTTP, YARPC client, zap, tally, deploy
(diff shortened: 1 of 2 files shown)


doc/rfc/web.md
```diff
@@ -0,0 +1,175 @@
...
+Operators have no browser UI. The CLI polls gateway history. Gateway already has summary, history, and receipt-time `List` ([status-list-api.md](submitqueue/status-list-api.md), [history-api.md](submitqueue/history-api.md)); Stovepipe has a parallel history API ([stovepipe/request-history-api.md](stovepipe/request-history-api.md)). Those RPCs are gRPC. A browser cannot call them the way the CLI does.
+
+The first UI is a **read** operator console: pick a queue, list requests, open summary and history, search by change URI, poll while a request is in flight. That is the same surface as `submitqueue/client` watch, not a workflow editor or a Cadence-scale SPA.
+
+Two ways to serve that HTML:
+
+1. **Go templates** — an importable Go library plus a Go HTTP host (OSS example under `service/`, production in go-code). Same library/host split as gateway.
+2. **Next.js BFF** — a Node process in OSS that the browser talks HTTP/JSON to; that process calls gRPC. Production is a source overlay (Cadence Web v4).
+
+SubmitQueue picks (1). The rest of this RFC is that design. Next.js stays the documented alternative if the UX outgrows pages, forms, fragments, and polling.
```
sbalabanov: Rationale for the choice?
Go's HTML rendering abilities and web control richness are often considered inferior to the JS/TS ecosystem.

doc/rfc/web.md
```diff
@@ -0,0 +1,175 @@
...
+service/submitqueue/web example host    HTTP, YARPC client, zap, tally, deploy
+```
+
+The web host is a **separate process** from the gateway. UI reads go through the gateway’s public RPC and inbound auth. Mounting templates on the gateway and calling controllers in-process would couple deploys and skip that boundary.
+
+Wire is still protobuf RPC. The OSS example adapter uses `google.golang.org/grpc`. Uber `submitqueue-gateway` is YARPC (`yarpcfx`, grpc inbound on `UBER_PORT_GRPC`); the go-code web host uses the generated **YARPC client**, not `@grpc/grpc-js`.
+
+Go **library** vs **host**: [modular-queue-wiring.md](submitqueue/modular-queue-wiring.md). The UI library is a client of the hosted gateway API.
+
+## What we lose by choosing templ over Next
```
sbalabanov: here is an inreresting conversation on the subject: https://www.reddit.com/r/htmx/comments/1k5bz3y/htmx_a_great_framework_that_ill_never_use_again/

doc/rfc/web.md
```diff
@@ -0,0 +1,175 @@
...
+## Testing
+
+- Handlers: mock gateway, `httptest` (success, bad query, not found, RPC error).
+- Components: render view models; assert semantics, not golden HTML files.
+- Fragments: HTMX vs full page; terminal status stops poll.
+- Assets: content type and cache headers.
+- Example compose: health, list, summary (`svc-submitqueue-web`).
+- go-code: identity, outbound credentials, auth rejection, M3, wiring.
+
+## Rejected
```
sbalabanov, line 168: Why these two are the only options considered?

sbalabanov (changes requested):
back for offline discussions

doc/rfc/web.md
behinddwalls: my preference is also not use htmx/tmpl in general and use NodeJS/React instead, i was trying to see if we can get same semantic without doing overlays...

doc/rfc/web.md
behinddwalls: Updated the RFC to choose Node.js/React. The Decision now makes the rationale explicit: the JS/TS ecosystem provides stronger accessible controls, client-state tooling, virtualization, forms, charts, and browser testing as the operator surface grows.

doc/rfc/web.md
behinddwalls: Aligned. The proposal is now Node.js/React with the Cadence-style production overlay. I also documented the React SPA plus Go-host option for avoiding an overlay, but it leaves two build stacks and a custom JSON BFF contract without Next server integration.

doc/rfc/web.md
behinddwalls: Expanded this beyond two choices. The RFC now compares the chosen Next.js BFF, Go templ/HTMX SSR, and a React SPA with a Go static/API host; it also explains why direct grpc-web and a Fusion-only implementation were not selected.

doc/rfc/web.md
behinddwalls: Thanks. HTMX is no longer proposed. The revised alternatives section records the server-fragment/browser coupling as one reason templ/HTMX was not selected.

behinddwalls:
Closing; we are dropping this RFC and the web UI stack.

PR #704, opened 2026-09-11
feat(errs): export Git command classification policy

sbalabanov:
## Summary
Intent:
- Let local Git and remote Git execution backends share one retryability policy.
- Avoid requiring non-process backends to manufacture gitexec.CommandError values.

Changes:
- Export a backend-neutral CommandFailure contract accepted by the Git classifier.
- Export direct command classification and decouple the policy from platform/git/exec.
- Cover local and remote error implementations and document the extension point.

---


platform/errs/git/git.go
```diff
@@ -63,6 +62,16 @@ var Classifier errs.Classifier = classifier{}
 
 type classifier struct{}
 
+// CommandFailure is the Git failure information needed by Classifier.
+//
+// Execution backends may satisfy this contract with their native error type;
+// they do not need to reconstruct an error from another Git implementation.
+type CommandFailure interface {
```
sbalabanov, line 73: add comments to explain  what Operation() and Diagnostics() should be. Explain the limitations of git in a sense that it does not have binary protocol, always output exit code 1 and we have to rely on running command and stderr for classification

sbalabanov: Expanded CommandFailure documentation in 6c909ae to define Operation as the Git subcommand and Diagnostic as the complete Git output, and to explain why generic non-zero statuses require both signals.

The contract now makes remote backends' obligations explicit without tying them to local process errors.

[addressed by agent]

mnoah1 (approved):

PR #707, opened 2026-09-15
feat(stovepipe): time-based build-slot delay - storage updates

mnoah1:
## Summary
Intent:
Add a way to support time-based throttling of new builds on a queue.  This can be used to support two main scenarios where we might not want to immediately start the next build as soon as the build slot is free:
1) Throttling resource usage on a shared pipeline: Support spacing out builds in a time based manner, which can be used when there are CI resource constraints in running the job constantly.  Tradeoff between overall agent usage time vs. getting a green result sooner.
2) Cool off after failure - A job that is experiencing repeated fast failures can create significant resource thrash if it continues to start then fail, and could benefit from delay between retries.  We can consider some sort of backoff workflow depending on failure type, to avoid unnecessarily workload and potentially give downstream issues time to resolve rather than constantly retrying/failing.

Changes:
- Add the build admission deadline to the Queue entity and MySQL backend.


stovepipe/extension/storage/mysql/schema/queue.sql
```diff
@@ -1,11 +1,12 @@
...
-    latest_request_id      VARCHAR(255) NOT NULL DEFAULT '',
-    version                INT          NOT NULL,
-    last_green_request_id  VARCHAR(255) NOT NULL DEFAULT '',
+    name                          VARCHAR(255) NOT NULL,
+    last_green_uri                VARCHAR(255) NOT NULL DEFAULT '',
+    in_flight_count               INT          NOT NULL DEFAULT 0,
+    latest_request_id             VARCHAR(255) NOT NULL DEFAULT '',
+    version                       INT          NOT NULL,
+    last_green_request_id         VARCHAR(255) NOT NULL DEFAULT '',
+    build_admission_not_before_ms BIGINT NOT NULL DEFAULT 0,
```
mnoah1, line 10: Need a default during the initial migration (cannot have NOT NULL without a default value), but can remove later after existing tables are updated.  New field must be last in the table.

behinddwalls:
Discussed offline, 

Mainly to re-think we can combine overall scheduler/gating logic behind an extension say concurrent jobs, delay or any other logic we may end up adding later so that we can trigger a call to extension to evaluate the gate and then schedule it

PR #717, opened 2026-09-16
docs(stovepipe): Add status lookup RFC

prathshenoy:
## Summary
**What**:
- Define repository and project validation lookup semantics.

**Why**:
- Establish a stable public contract before implementation begins.

## Test Plan
- [x] Get dual model consensus, and review the RFC for API semantics, lifecycle/error behavior, pagination, and implementation flexibility.

## Revert Plan
- Revert this PR. The changes in this PR only affect documentation.

## Issues
- [CODEM-472](https://linear.app/uber/issue/CODEM-472/rfc-doc)

## Stack
1. @ #717
1. #718
1. #719


doc/rfc/stovepipe/get-project-status-by-uri-api.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Contract
+
+The request and response fields have the following API-level meaning. The published protobuf is the authoritative field-level contract.
+
+| Request field | Required | Meaning |
+| --- | --- | --- |
+| `queue` | Yes | Identifies the queue in which to find the validation request. |
+| `change_uri` | Yes | Identifies the exact commit URI under validation. |
+| `project` | No | Selects one reported project instead of a page of projects. |
```
mnoah1: Let's take a list of projects here, and make mandatory - to support batched requests/responses.

For now, this still doesn't provide protection from a malformed project name, because we don't have stovepipe call tango for an authoritative list of services, but that can be supported in this shape.  Also, the mandatory status could be relaxed in the future as we build out functionality to provide the full project list.

mnoah1 (approved):
See comment below

doc/rfc/stovepipe/get-project-status-by-uri-api.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Contract
+
+The request and response fields have the following API-level meaning. The published protobuf is the authoritative field-level contract.
+
+| Request field | Required | Meaning |
+| --- | --- | --- |
+| `queue` | Yes | Identifies the queue in which to find the validation request. |
+| `change_uri` | Yes | Identifies the exact commit URI under validation. |
+| `projects` | No | Limits project results to the supplied consumer-defined project IDs. |
```
mnoah1: This should still be mandatory, right?  It could loosen in the future if Stovepipe has an authoritative project list, but for now we always need at least one project to check against.

PR #726, opened 2026-09-18
fix(stovepipe): separate record dlq handler

mnoah1:
## Summary

During the record step, we were previously reusing the existing record handler as the dlq handler.  Since record has outbound side effects (e.g. Promote), this can lead to 1000x attempts to external APIs even if the request will never succeed.

This PR adds a separate DLQ handler for record, focused on repair/recording an event for the abandoned record attempt, rather than repeated attempts that will likely continue to fail if the original retry budget is exhausted.

Intent:
- Ensure record-stage work that exhausts the primary retry budget reaches an observable stopping point instead of retrying indefinitely on the DLQ.
- Preserve whatever durable effects already succeeded without trying to reconstruct or complete the original workflow.
- Keep fail-closed behavior when no validation fact was recorded. Persistent promotion permission failures are one example of this broader case.

Changes:
- Add a standalone record DLQ controller that reloads the durable Request and appends an idempotent `record_abandoned` history event.
- Keep the original failure in structured logs without adding an event-reason or arbitrary metadata contract.
- Acknowledge malformed or unresolvable poison after logging and counting it; retry only Request reads and history persistence.
- Do not create facts, advance bookmarks, call source control, or publish hooks from the record DLQ.
- Rely on the queue-scoped storage contract instead of rechecking the loaded Request queue.


stovepipe/controller/record/dlq.go
```diff
@@ -0,0 +1,160 @@
...
+				"request_id", rec.GetId(),
+				"queue", rec.GetQueueName(),
+			)
+			return nil
+		}
+		metrics.NamedCounter(c.metricsScope, _dlqOpName, "request_store_errors", 1, metrics.TagsFromContext(ctx)...)
+		return fmt.Errorf("failed to load request %s for record dlq: %w", rec.GetId(), err)
+	}
+	originalFailure, hasFailure := delivery.Failure()
+	log := requestlog.NewRequestEventLog(request, entity.RequestEventRecordAbandoned, "repository", nil)
```
behinddwalls, line 127: do we need to update the request itself?

behinddwalls (approved):

stovepipe/controller/record/dlq.go
mnoah1: The `record` step happens only after the request has already reached a terminal state via build/buildsignal, so there should not need to be further updates to the actual request entity by this point.

PR #735, opened 2026-09-21
feat(stovepipe): Record project validation facts

prathshenoy:
## Summary

**What**:
- Record named project outcomes alongside final repository validation results.
- Preserve repository-only validation for deployments without project attribution.

**Why**:
- Enable consumers to query durable project failures when an integration supplies attribution.

## Test Plan

- [x] Add unit tests.

## Revert Plan

- Revert this PR. This removes named project-result persistence while preserving repository validation.

## Issues
- [CODEM-529](https://linear.app/uber/issue/CODEM-529/persist-project-validation-results)


stovepipe/controller/record/record_test.go
```diff
@@ -125,6 +126,23 @@ func (f staticSourceControlFactory) For(sourcecontrol.Config) (sourcecontrol.Sou
 	return f.sourceControl, nil
 }
 
+type staticProjectResultFactory struct {
```
mnoah1: Can we use generated mocks instead of adding a static one here?

stovepipe/extension/projectresult/projectresult.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+// Factory constructs a Resolver for one queue.
+type Factory interface {
+	For(cfg Config) (Resolver, error)
+}
+
+// NoopFactory returns a resolver that records no project outcomes. It is the
+// default for deployments that do not configure project attribution.
+type NoopFactory struct{}
```
mnoah1: Can put this in its own package (similar to other pattern of putting implementations as subpackages below the interface's package)

stovepipe/controller/record/record.go
```diff
@@ -60,6 +62,7 @@ type Controller struct {
 	metricsScope  tally.Scope
 	stores        storage.Factory
 	materializer  requestlog.Materializer
+	projectResult projectresult.Factory
```
mnoah1: Let's think about the naming here, from the name alone it's a bit hard to tell if it's a provider of project results, or an actual result about a project.  Maybe something like `projectResultResolver`

stovepipe/controller/record/record.go
```diff
@@ -176,6 +183,46 @@ func (c *Controller) Process(ctx context.Context, delivery consumer.Delivery) er
...
+			URI:       request.URI,
+			Project:   result.Project,
+			Degree:    result.Degree,
+			RequestID: request.ID,
+			CreatedAt: time.Now().UnixMilli(),
+		})
+		if err != nil {
+			return err
+		}
+		if err := c.persistValidationFactRecordedLog(ctx, store, request, fact); err != nil {
```
mnoah1: We left the logging a bit open ended for this:  https://github.com/uber/submitqueue/blob/7b87743fb2d87d2f4bf1399d2f7b8a6cd30c3449/doc/rfc/stovepipe/request-log.md#vocabularies - it might result in excessive log volume to log every individual project.  Should we just create an event that indicates that recording of all projects is done?

mnoah1 (approved):

PR #736, opened 2026-09-22
feat(queue): add lease-owner metrics and processing logs

behinddwalls:
### Why?

Lease ownership was only visible in the DB and the admin CLI. Renewals were a silent batch update, steals had no event, and process logs carried message_id and partition_key but not which replica held the lease — so "who owns this key" and "which node processed this message" were hard to answer from metrics or logs.

### What?

Subscriber metrics tag leased_by, topic, tenant, and consumer_group, and deliberately never partition_key: acquire/steal/release counters, renew row counts, partitions_owned on every discovery, and active_subscribers plus fair_share_cap only after a successful lease-tick heartbeat query. A cap of 0 means unlimited internally; the gauge instead reports the size of the owned-plus-discovered partition set, so `partitions_owned > fair_share_cap` cannot fire trivially on a lone replica. Info logs cover steals (previous_owner from the pre-acquire snapshot), the release reason, and a per-tick summary even when renew fails; a first claim stays at Debug.

Release counters report effect rather than intent. `ReleaseLease` returns rows deleted alongside its error — it already computed the count and discarded it — so the rebalance and idle counters skip a release whose DELETE matched nothing, which happens whenever the lease was stolen or purged between the tick's read and the release. Shutdown releases one tenant at a time so the count carries a tenant tag and a failure names the tenant it belongs to. The idle path records the release as soon as the lease row is gone, before offset cleanup, so a DeleteOffset failure no longer suppresses a counter for a lease that genuinely was released.

Deliveries carry leased_by and consumer_group in metadata. The consumer snapshots those fields once before handing the delivery to the controller, because Metadata() exposes a map the controller can write to, and reuses that snapshot on the process, hold-ignored, reject, and nack logs so every line agrees on the owner.

## Test Plan

✅ `go test -race ./platform/extension/messagequeue/mysql/ ./platform/consumer/`
✅ `make fmt`, `make gazelle`, `make mocks`, `make check-gazelle`


platform/extension/messagequeue/mysql/subscriber.go
```diff
@@ -896,21 +910,95 @@ func (s *subscriber) acquireUnownedPartitions(ctx context.Context, sub *subscrip
...
+	}
+	if partitionKey != "" {
+		fields = append(fields, logPartitionKey, partitionKey)
+	}
+	return fields
+}
+
+func (s *subscriber) leaseTags(sub *subscription, tenant string) []metrics.Tag {
+	tags := []metrics.Tag{
+		metrics.NewTag(logLeasedBy, sub.config.SubscriberName),
```
mnoah1, line 949: Subscriber name is unique to each host + process combo, so it will result in unbounded cardinality for this metric tag.   The other fields should be ok and have a discrete set of possible values, but I would remove this here.

behinddwalls: i thought about it but it would be hard to debug, we could remove process from name and just keep host? thoughts

mnoah1: Ok that sounds good.  Alternatively, looks like there is an option for us to configure internally whether or not host gets included in the injected tally scope.

mnoah1 (approved):
Approve pending updates to remove pid from the tags

platform/extension/messagequeue/mysql/subscriber.go
behinddwalls: yes.. for now lets keep it, if it becomes and issues or we cannot drop then we can remove it...

